PR #20, opened 2018-06-07
Updated contributor docs

planetf1:
Added to developer contribution docs
(2nd merge request as first was cluttered with master pulls)

Signed-off-by: Nigel Jones <jonesn@uk.ibm.com>


planetf1:
I inserted the text at the wrong point. 

The current process refers to use of branches of the egeria repo which is suitable for a project where everyone has write access, but as a growing open source project I think we should use the github fork and full process.

I'll propose a more complete version of this in the next few commits

PR #47, opened 2018-07-04
gaian-connector initial commit

bogdan-sava:


mandy-chessell:
Code looks ok - there seems to be 2 levels of connector - we will need to move the generic connector to the data-store-connector maven module - but this can be done later - please raise a Jira to do this.

Issue #73, opened 2018-07-13
Add apache rat plugin to check if the license is present in the files


bogdan-sava:
If the license is not present, build fails.
In the maven fail message is the path of the rat.txt file with details about what files don't have the proper license.

PR #74, opened 2018-07-13
#73 add apache rat plugin

bogdan-sava:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -231,6 +231,8 @@
                         <exclude>**/*.json</exclude>
                         <exclude>**/*.log</exclude>
                         <exclude>**/*.iml</exclude>
+                        <exclude>**/*.registrystore</exclude>
+                        <exclude>**/*.results</exclude>
                         <exclude>**/target/**</exclude>
                     </excludes>
                     <licenses>


bogdan-sava:
Also there fixes for 2 files without license, detected by rat

bogdan-sava:
I have excluded .json from rat check because of this file:
open-metadata-implementation/governance-servers/admin-services/admin-services-configuration.postman_collection.json

This file has no license, and I'm not sure if should I add it or not.
Any suggestions?

pom.xml
```diff
@@ -217,6 +216,49 @@
...
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <version>0.12</version>
+                <configuration>
+                    <useDefaultExcludes>true</useDefaultExcludes>
+                    <useMavenDefaultExcludes>true</useMavenDefaultExcludes>
+                    <useIdeaDefaultExcludes>true</useIdeaDefaultExcludes>
+                    <useEclipseDefaultExcludes>true</useEclipseDefaultExcludes>
+                    <excludeSubProjects>true</excludeSubProjects>
+                    <excludes>
```
planetf1: I think we may need some additional excludes. The first that came to mind are the OMAG Server configuration files, which currently end up being put in the root directory.

There are also other files created during a build such as test results.

The top level .gitignore is a start - could we add that as an excludesFile (not tried it?) - even though some are already explicitly listed and/or covered by the IDEA exclusions etc

bogdan-sava: There are 62 implicit excludes. Among them .gitignore. 
The OMAGApplication configuration creates a .json file which is excluded.
Of course I have tested it, build the project with it and works.

mandy-chessell: The OMAG server config file is called *.config rather than *.json
The cohort registry file is called *.registrystore
The compliance test results file is called *.results

PR #95, opened 2018-07-20
[Egeria 22] Asset Catalog OMAS - initial commit

danielaotelea:


davidradl:
@planetf1 I see. The pom fix up in the server chassis in my PR, removes the dependency on the asset catalog omas so it allows it to build without asset catalog.

PR #101, opened 2018-07-23
EGERIA-29 fix merge issues in pom and archive builder

davidradl:
Fixing a pom corruption in the server chassis. 
Fixing a but in the archive builder - where an attribute is added twice. 

I am correcting the .gitignore to ignore the generated files


planetf1:
Unable to test this change as does not build..

however there’s still an issue with https://github.com/odpi/egeria/pull/95 which has introduced org.apache.atlas/omrs as a dependency @Daniela Otelea

One poss fix is to reject that PR and undo - if it doesn’t have other side effects.another option is to remove asset-catalog from the build, and then raise an issue/jira to fix it up and reintroduce.

planetf1:
Though this PR by itself 'looks' good, the previously committed breakage means I can't test. For that reason I propose the minimal extra fix to get clean build is done. That probably means removing asset catalog from the build & raising a jira/issue to track re-introduction. or we backup Subject area and asset catalog.. but that probably just delays further work. Given asset catalog isn't being consumed shall we go with that as a way forward? If you agree add to commit? Please also check my understanding. I am looking at maven central & cannot see the omrs component, which means a clean environment build failure

planetf1:
Merging this as it fixes some of the build issues. will open up new issue to fix remaining issues

PR #113, opened 2018-07-24
Add highlevel maven packages and OMAG Server Assembly

mandy-chessell:
Tracked under https://jira.odpi.org/browse/EGERIA-34

### DIFF
--- a/open-metadata-distribution/open-metadata-assemblies/pom.xml
+++ b/open-metadata-distribution/open-metadata-assemblies/pom.xml
@@ -1,0 +1,92 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <parent>
+        <artifactId>open-metadata-distribution</artifactId>
+        <groupId>org.odpi.egeria</groupId>
+        <version>0.1-SNAPSHOT</version>
+    </parent>
+
+    <name>Open Metadata Assemblies</name>
+    <description>
+        Packaged executables from Egeria.
+    </description>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>open-metadata-assemblies</artifactId>
+
+    <profiles>
+        <profile>
+            <id>Windows</id>
+            <activation>
+                <os>
+                    <family>windows</family>
+                </os>
+            </activation>
+            <properties>
+                <python.path.l>
+                    ${project.basedir}\src\bin;${project.basedir}\src\test\python\scripts;${project.basedir}\src\test\mock
+                </python.path.l>
+            </properties>
+        </profile>
+
+        <profile>
+            <id>Linux</id>
+            <activation>
+                <os>
+                    <family>!windows</family>
+                </os>
+            </activation>
+            <properties>
+                <python.path.l>
+                    ${project.basedir}/src/bin:${project.basedir}/src/test/mock:${project.basedir}/src/test/python/scripts
+                </python.path.l>
+            </properties>
+        </profile>
+
+        <profile>
+            <id>dist</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <artifactId>maven-assembly-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>single</goal>
+                                </goals>
+                                <phase>package</phase>
+                                <configuration>
+                                    <skipAssembly>false</skipAssembly>
+                                    <descriptors>
+                                        <descriptor>src/main/assemblies/omag-server.xml</descriptor>
+                                    </descriptors>
+                                    <finalName>egeria-${open-metadata.version}</finalName>
+                                    <tarLongFileMode>gnu</tarLongFileMode>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-deploy-plugin</artifactId>
+                        <version>2.7</version>
+                        <configuration>
+                            <skip>false</skip>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+</project>
--- a/open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
+++ b/open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
+<assembly xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2"
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
+    <formats>
+        <format>tar.gz</format>
+        <format>dir</format>
+    </formats>
+    <id>omag-server</id>
+    <baseDirectory>omag-server-${open-metadata.version}</baseDirectory>
+    <fileSets>
+        <fileSet>
+            <directory>../omag-server/target</directory>
+            <outputDirectory>.</outputDirectory>
+            <includes>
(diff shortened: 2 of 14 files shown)


open-metadata-distribution/open-metadata-assemblies/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+        Packaged executables from Egeria.
+    </description>
+
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>open-metadata-assemblies</artifactId>
+
+    <profiles>
+        <profile>
+            <id>Windows</id>
```
planetf1: For now I'm not sure we have any platform specifics. If/when we get to the point of adding .cmd / powershell scripts for windows and shell scripts for linux/mac, then this is a mechanism to get the right things included -- but I don't think we have any now, nor do we have python scripts, so I'd be inclined to remove them to avoid clutter - we can add once we identify a need?

open-metadata-distribution/open-metadata-assemblies/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+            </activation>
+            <properties>
+                <python.path.l>
+                    ${project.basedir}/src/bin:${project.basedir}/src/test/mock:${project.basedir}/src/test/python/scripts
+                </python.path.l>
+            </properties>
+        </profile>
+
+        <profile>
+            <id>dist</id>
```
planetf1: In Atlas one had to do a -Pdist to create the distribution packages. Maven has a 'package' step but in atlas this was used only for  building jars and also an 'install' step which is more about publishing the maven artifacts.

That being said if we have the right dependencies, we could remove the profile and bind to the package phase. Upside - simpler build, no need for -Pdist, always create arcives. Downloads - adds a little extra time. 

On balance I'd move to package lifecycle step & remove the profile

open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
```diff
@@ -0,0 +1,39 @@
...
+          xsi:schemaLocation="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2 http://maven.apache.org/xsd/assembly-1.1.2.xsd">
+    <formats>
+        <format>tar.gz</format>
+        <format>dir</format>
+    </formats>
+    <id>omag-server</id>
+    <baseDirectory>omag-server-${open-metadata.version}</baseDirectory>
+    <fileSets>
+        <fileSet>
+            <directory>../omag-server/target</directory>
```
planetf1: Is this the right relative path for omag server? It;s coming from the open-metadata-implementation/governance-servers/server-chassis area? (build failed: will comment more on this later)

open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
```diff
@@ -0,0 +1,39 @@
...
+            <outputDirectory>.</outputDirectory>
+            <includes>
+                <include>*.jar</include>
+                <include>*.sh</include>
```
planetf1: We don't have any shell scripts currently. harmless, but could be unintended so I would remove until we need it?

open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
```diff
@@ -0,0 +1,39 @@
...
+
+            </includes>
+            <fileMode>0755</fileMode>
+            <directoryMode>0755</directoryMode>
+        </fileSet>
+        <fileSet>
+            <directory>../omag-server</directory>
```
planetf1: I think this directory is wrong (see above)

open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/omag-server.xml
```diff
@@ -0,0 +1,39 @@
...
+                <include>*.md</include>
+
+            </includes>
+            <fileMode>0755</fileMode>
+            <directoryMode>0755</directoryMode>
+
+        </fileSet>
+
+    </fileSets>
+</assembly>
```
planetf1: I keep getting those nulls at end of line too. They show up as red on a review. I think there's a setting in intellij to fix. I've NOT tested but there is Preferences->Editor->General->Ensure line feed at file end on Save ? Maybe...

open-metadata-distribution/open-metadata-packages/open-metadata-caller-package/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+
+    <parent>
+        <artifactId>open-metadata-packages</artifactId>
+        <groupId>org.odpi.egeria</groupId>
+        <version>0.1-SNAPSHOT</version>
+    </parent>
+
+    <name>Open Metadata Caller Package</name>
+    <description>
+        Provides the basic package for calling egeria from a client.
```
planetf1: Since this is a package of our rest services (spring) it's what is needed in the server to support client access. This would be distinct from a client package which includes the java classes to make those rest calls easier, so I think I'd name it something like 'provides support for egeria clients in a server' or similar?

open-metadata-distribution/open-metadata-packages/open-metadata-client-package/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+
+    <parent>
+        <artifactId>open-metadata-packages</artifactId>
+        <groupId>org.odpi.egeria</groupId>
+        <version>0.1-SNAPSHOT</version>
+    </parent>
+
+    <name>Open Metadata Client Package</name>
+    <description>
+        Provides the basic package for calling egeria from a client.
```
planetf1: Maybe it was previously just cut/paste glitch. Name looks good here

planetf1:
Verified master is clean - was broken but now fixed.
Before pulling master, verifying just this PR

11:32 $ mvn clean install
[INFO] Scanning for projects...
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:subject-area-tools:jar:0.1-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.odpi.egeria:repository-services-implementation:jar -> duplicate declaration of version ${open-metadata.version} @ org.odpi.egeria:subject-area-tools:[unknown-version], /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/access-services/subject-area/subject-area-tools/pom.xml, line 24, column 21
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
[ERROR] [ERROR] Project 'org.odpi.egeria:admin-services-client:0.1-SNAPSHOT' is duplicated in the reactor @
[ERROR] Project 'org.odpi.egeria:admin-services-client:0.1-SNAPSHOT' is duplicated in the reactor -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DuplicateProjectException
✘-1 ~/src/egeria [mandy-chessell-distribution L|⚑ 4]

Pulled in latest master updates - no change, these two issues remain.

The warning is a simple duplication in the pom. That's in master too - but often not noticed until a build fails:

In the subject-area-tools pom:

 <dependencies>
        <dependency>
            <artifactId>repository-services-implementation</artifactId>
            <version>${open-metadata.version}</version>
        </dependency>
        <dependency>
            <artifactId>repository-services-implementation</artifactId>
            <version>${open-metadata.version}</version>
        </dependency>
    </dependencies>

We can delete one dependency entry

The second .. looking...

planetf1:
I think the build failure is caused by the same module being included in two places -- firstly as part of the main build tree (root->implementation->governance servers -> admin services -> client 
& secondly under distribution- packages - client

Maven doesn't like this as it sees the entries twice. It's possible to build the packages alone (presumably) but the full build will fail.

We need to think through what we need here.. I wonder if we just need a virtual pom that depends on the other poms - that's good enough, so they are not modules (and so don't affect build), but are dependencies so that if another app wants to pull in open-metadata-client-package from maven, they will automatically get all the pieces we list?

This is distinct from the assemblies where we are layng down an archive for someone to run a server and want to pull in other files, scripts etc.. ?

mandy-chessell:
Why is the build problem only picked up by subject area OMAS.  It is one of the last OMASs to build - is this because of the code generator?  It seems to suggest that the code generator is not properly integrated?  Did I understand that it is generating the code into src rather than target/generated-sources?  If so I think we may need to change that as I thought src should be the content of the SCM repository?

planetf1:
I was referring to a build failure of the overall project. Still fails with
[ERROR] [ERROR] Project 'org.odpi.egeria:admin-services-client:0.1-SNAPSHOT' is duplicated in the reactor @

This is due to the <module> statements in the new poms (they are already used elsewhere in the build>

You could use multiple aggregating poms and split from parent - though I wonder if this is overkill/confusing. Not an expert, but I would hope you could create a new module that simply prereqs the individual modules we need, and point consumers to those. So we will publish all the original poms, but make choosing the right ones slightly easier?

PR #120, opened 2018-07-25
EGERIA-31 Fix for null pointer for missing Kafka parameters

cmirash:
Signed-off-by: Chandana Mirashi <cmirash@us.ibm.com>

I added missing license information on 2 files CODE_OF_CONDUCT.md and IsTypeOf.java otherwise egeria build fails


planetf1:
Actually I think git merge is being sensible and allowing it just fine. testing and will commit if it looks good - as Graham is hitting a kafka issue too (may be unrelated)

planetf1:
looks good. not tested kafka function but as it's new & in process consider it appropriate to now merge. Doesn't seem to conflict with #122 where the same SPDX omissions were also fixed

PR #124, opened 2018-07-25
Add typedef guid and name to OMRSInstanceEvent in constructor

mlottesen:
Signed-off-by: Mark Ottesen <mottesen@us.ibm.com>

Need to set typeDefGUID and typeDefName on OMRSInstanceEvent for entities to be sent to the cohort.  This is a regression from when OMRSRepositoryEventManager.processInstanceEvent took the entity itself as its argument - was working for me last Friday, but no longer is.  

Also fixed a unit test that wasn't compiling while I was in the neighborhood.


planetf1:
I was able to build cleanly.... though am not using events yet myself so unable to verify that aspect

mandy-chessell:
I'll fix code layout later

Issue #134, opened 2018-08-03
Documentation cleanups


cmgrote:
Issue to track documentation cleanups:
- [x] maximise use of markdown for consume-ability (ie. language-specific syntax highlighting, intra-document linking, nested lists, etc)
- [x] consistently refer to project as "Egeria" rather than "egeria"
- [x] spelling / grammatical / copy/paste / typo corrections

PR #139, opened 2018-08-03
Doc cleanups

cmgrote:
Fixes (mostly minor) typos and formatting hiccups to address #134


cmgrote:
Hmmm... forgot to sign this one commit, but seems to still be problematic even after reverting it and committing back (with a signature).  Any ideas?

cmgrote:
Seems like this DCO signing is buggy -- that last commit (b31d179) is definitely signed, but still marked as not signed (?)

PR #160, opened 2018-08-20
Fix up Access Service POM file dependencies

mandy-chessell:
Many of the access services has incorrect dependencies between the
client, server and api modules.  Correct dependencies is that the client
and server modules should dependend on the api module.  There must be no
dependency between the client and the server module.


mandy-chessell:
This pull request is part of https://jira.odpi.org/browse/EGERIA-23

Issue #181, opened 2018-09-13
Add sample postman scripts to git for server config/launch


planetf1:
I have some postman scripts with variables (URL, user, server, cohort) which can be used to configure the server chassis from scratch for use with an in-memory repository. These would form a useful example for anyone trying to use egeria -- as I found out when I included kafka configuration body in the wrong rest call.

I propose to add these with a README

mandy-chessell:
Is tis complete?

planetf1:
Yes, as a self contained submission it is.

PR #182, opened 2018-09-13
Add sample postman scripts #181

planetf1:
Added some sample postman scripts to configure Egeria for in-memory repository


mandy-chessell:
There is already an example postman scripts within the admin services module

Issue #218, opened 2018-09-21
findEntitiesByClassification -> unexpected exception if type null


planetf1:
The following code fragment

  ```
         try {
                entities = metadataCollection.findEntitiesByClassification(userId, typeGuid,
                        classification,
                        null, null, 0,
                        null,
                        null, null, null, 0);//TODO Handle exceptions from findEntitiesByClassification
            } 
```
was used with
 userId = "me"
 typeGuid = null
 classification = "DataValue"
 - other parms 0/null

The exception returned was:
`org.odpi.openmetadata.repositoryservices.ffdc.exception.RepositoryErrorException: OMRS-REPOSITORY-400-055 An unexpected exception was received from a repository connector during the findEntitiesByClassification operation which had message: null`

The exception seems to be caught in EnterpriseOMRSMetadataCollection where 
 ```
                   List<EntityDetail> results = metadataCollection.findEntitiesByClassification(userId,
                                                                                                 entityTypeGUID,
                                                                                                 classificationName,
                                                                                                 matchClassificationProperties,
                                                                                                 matchCriteria,
                                                                                                 fromEntityElement,
                                                                                                 limitResultsByStatus,
                                                                                                 asOfTime,
                                                                                                 sequencingProperty,
                                                                                                 sequencingOrder,
                                                                                                 pageSize);
```
gets a null pointer exception

The method comments suggest a null type should be ok

The cause of this is in OMRSRepositoryContentValidator verifyEntityIsClassified where we iterate through the entity's classifications
```
    public boolean verifyEntityIsClassified(List<String>   requiredClassifications,
                                            EntitySummary  entity)
    {
        if (requiredClassifications != null)
        {
            List<Classification> entityClassifications = entity.getClassifications();

            for (String requiredClassification : requiredClassifications)
            {
                if (requiredClassification != null)
                {
                    for (Classification entityClassification : entityClassifications)
                    {
                        if (entityClassification != null)
```
At this point the method has been called with a sensible requireClassifications 
```
requiredClassifications = {ArrayList@7628}  size = 1
 0 = "AbstractConcept"
```
and the entity looks ok:
```
[...]

PR #222, opened 2018-09-24
#218 fix NPE for unclassified entity

planetf1:
Added additional null check as per #218 

Looks a simple change. Works in my code. No UT on this module, so would ask whoever merges to just check carefully - thanks :-)


planetf1:
Ok will make changes.
Of note is that { on same line is the default IntelliJ style for auto-code completion/formatting, and is the form I'm seeing most other's code use, so it may be a topic for discussion and documentation of any layout standard.

That being said completely agree that consistency at the innermost level is important so will update the change here.

planetf1:
reformatted and PR updated

mandy-chessell:
It is a common style true - but a proven less readable style.  The reasons for its original introduction no longer apply.  I have not insisted on the more readable layout throughout the project in the dev guide because I know that it is rowing uphill - just that the existing code layout is respected when making changes to existing code.

Issue #237, opened 2018-09-25
Add basic CRUD for Term and Category including samples


davidradl:

mandy-chessell:
Is this finshed?

davidradl:
yes - closing

PR #265, opened 2018-10-02
basic crud for term and category #237

davidradl:


open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/nodesummary/AssetSummary.java
```diff
@@ -0,0 +1,39 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
```
mandy-chessell: Can not use the ASF license in ODPi project

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/term/Term.java
```diff
@@ -40,7 +42,15 @@
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.PUBLIC_ONLY;
 
 /**
- * TODO
```
mandy-chessell: Having a TODO in class description Javadoc entry seems strange - would be better to start with a description of the class

open-metadata-resources/open-metadata-samples/access-services-samples/subject-area-client-samples/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/samples/CategorySample.java
```diff
@@ -0,0 +1,155 @@
+/* SPDX-License-Identifier: Apache-2.0 */
```
mandy-chessell: It would be nicer if the sample was based on Coco Pharmaceuticals

open-metadata-resources/open-metadata-samples/access-services-samples/subject-area-client-samples/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/samples/CategorySample.java
```diff
@@ -0,0 +1,155 @@
...
+                }
+            }
+        } catch (SubjectAreaCheckedExceptionBase e) {
+            System.out.println("ERROR: " + e.getErrorMessage() + " Suggested action: " + e.getReportedUserAction());
+        } catch (IOException e) {
+            System.out.println("Error getting user input");
+        }
+    }
+
+    private static Category createCategoryWithParentGlossaryName(String categoryName,Category parent, String glossaryName) throws SubjectAreaCheckedExceptionBase {
```
mandy-chessell: Why are statics used?

davidradl: @mandy-chessell  To keep the samples simple and easy to read and to allow re use of the static methods, I coded all the samples with static methods. Adding a new instance for a sample would tie the create glossary method to a particular sample unless I introduced class inheritance or static methods for re use. 

If you have strong views on this I can re-code with non-static methods.

mandy-chessell: I am not sure this does make them easy to follow.  I am finding it very hard to locate which pieces are scaffolding for the sample and where the API calls are that the sample is supposed to be teaching about.  It looks more like an functional verification test (which we do need and we need to think about where we put FV tests).  Partly my impression comes from the fact that the data values are of the form "term1" rather than a real business term such as "Employee".  But also, I don't think this is the style we would want someone to copy - for example, the userId is hardcoded.  There are no descriptions with the terms - is this because the sample is showing an initial brainstorm of terms?  It would be good to provide some context on where this fits in the overall process of building a glossary.   Also, it is worth thinking about whether the Subject Area API is simple enough if it is necessary to wrap the API calls with additional methods that provide common call patterns?  Maybe these static methods are closer to the API?

davidradl: I had written the samples more as FV tests. Maybe the best thing to do is remove the samples as they are not in the style of a scenario. I will do a scenario orientated subject area sample when I have exposed more capability in the Subject Area omas, I need to have relationships APIs.

mandy-chessell: Maybe leave them for now and we create a place for these types of tests - I really like the idea of having these types of test where we can fully exercise from the API to the backend.  And then we can also add the more CocoP type scenario.

In working through the Governance Program definition, I created a scenario where Erin and Peter are defining the core subject areas.  See https://github.com/odpi/data-governance/tree/master/docs/coco-pharmaceuticals/scenarios/defining-subject-areas.  I was thinking of creating a sample where Erin adds the subject areas defined in Figure 2 to the metadata repository.   Basically this would add a glossary, the subject areas with name and descriptions (there is a 2 level hierarchy) and then assign an owner to them.  I can add this capability to the Governance Program OMAS, but I was thinking that it would be better if this used the SubjectArea OMAS API.  What do you think?

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/term/Term.java
davidradl: It looks like I am deleting the TODO here and replacing with text.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/nodesummary/AssetSummary.java
davidradl: Good spot.  I have amended the java files to use the correct license.

Issue #279, opened 2018-10-05
kafka eventbus/cohort consumer group.id needs to be the same for same omag config/cohort instance starts


hjv333:
Currently, you get a different kafka consumer group.id whenever an omag server on a cohort gets restarted.  So, if the omag server is "down" for a period of a time and start again, you will not receive topic events from the last commit (since you've connected with a different consumer group.id), instead, will go to latest offset and consume from there.

Need to be able to specify the same kafka consumer group.id for the same omag server config on a cohort.

PR #295, opened 2018-10-09
Adding unique id to server config and passing to Kafka #279

mandy-chessell:
See https://github.com/odpi/egeria/issues/279


mandy-chessell:
Please validate this pull requests works with your Kafka setup.  One change that is important to note - the name of the topic in the connection endpoint was being ignored.  It should now use the proper value.  The connector is producing Audit log messages so you should see something like:

OCF-KAFKA-TOPIC-CONNECTOR-0001 Connecting to Apache Kafka Topic /open-metadata/repository-services/cohort/cohort1/OMRSTopic with a server identifier of fed729d6-c1d7-4270-8437-23c258d25017

The other change is to add the server Id to the key of each message and ignore any message with a key that matches the local server Id.  This is a performance enhancement.

Two questions:
Does the new configuration work in your environment?
Are all of the messages being transmitted correctly.

guptaneeru:
Server id is added in messages. Should this also be added in Member Registration class. Such that we can extract the server id of local server while persisting registry store?

mandy-chessell:
@guptaneeru 
> Server id is added in messages. Should this also be added in Member Registration class. Such that we can extract the server id of local server while persisting registry store?

The simple answer is no - the server Id is not currently needed to persist the registry store.   The registry store relates to the membership of the cohort, not the config of the server.

mandy-chessell:
Testing this fix revealed that the URL type structure for topic names that we have defined in OMRS and the OMASs do not work as Apache Kafka does not support "/" in a topic name.  So the topic names have been changed to this format: CP.open-metadata.access-services.AssetConsumer.inTopic

planetf1:
Worked through my GE omas scenario (configure event bus, enable access services, cohort, start instance, get classifications, add a few assets, get governed assets) - all worked fine.

Will review logs further and any additional notes, but so far behaviour seems the same with the PR as before. 

my //current// config is 
```
{
"producer":
{
"bootstrap.servers":"{{kafkaep}}",
"acks":"all",
"retries":"0",
"batch.size":"16384",
"linger.ms":"1",
"buffer.memory":"33554432",
"max.request.size":"10485760",
"key.serializer":"org.apache.kafka.common.serialization.StringSerializer",
"value.serializer":"org.apache.kafka.common.serialization.StringSerializer",
"kafka.omrs.topic.id":"{{cohort}}"
},
"consumer":
{
"bootstrap.servers":"{{kafkaep}}",
"zookeeper.session.timeout.ms":"400",
"zookeeper.sync.time.ms":"200",
"fetch.message.max.bytes":"10485760",
"max.partition.fetch.bytes":"10485760",
"key.deserializer":"org.apache.kafka.common.serialization.StringDeserializer",
"value.deserializer":"org.apache.kafka.common.serialization.StringDeserializer",
"kafka.omrs.topic.id":"{{cohort}}"
}
}
```

planetf1:
I was also able to run the same scenario with a simpler kafka config
```
{
"producer":
{
"bootstrap.servers":"{{kafkaep}}",
"kafka.omrs.topic.id":"{{cohort}}"
},
"consumer":
{
"bootstrap.servers":"{{kafkaep}}",
"kafka.omrs.topic.id":"{{cohort}}"
}
}
```
(Note those variables are defined by my postman environment - the endpoint URL, and a name - in my case the same as cohort name)

mandy-chessell:
Where is the value "localhost:9092/OMRSTopic" coming from?  I think it is coming from the event bus URLRoot parameter ?

bogdan-sava:
It is using  the endpoint.address as topic ConnectorConfigurationFactory.getDefaultEventBusConnection
 which is concatenated with topicURLRoot


if (topicURLRoot == null)
{
   endpoint.setAddress(topicName);
 }
else
{
   endpoint.setAddress(topicURLRoot + "." + topicName);
}

mandy-chessell:
If you look at the code above, the URLRoot and the topic name are connected with a ".".  This is the code I see in master.   The URL root is passed (or should be passed) as a request param on the event bus command - it is not passed in the properties from the request body.  In the original post, the topic name is coming out as "localhost:9092/OMRSTopic.open-metadata.repository-services.cohort.cohort1.OMRSTopic". This suggests to me that the URLRoot is set to "localhost:9092/OMRSTopic" which is an invalid value.  What is the URLRoot passed on the event bus command in your scenario?  I think this is a case of passing incorrect parameters on the configuration call but I want to be sure.

bogdan-sava:
Indeed, bad value for topicURLRoot query parameter in OMAGServerAdminResource.

However, we need validations for parameters, and return bad request if they are not compatible.
I have created a bug for it:
https://github.com/odpi/egeria/issues/302

bogdan-sava:
Maybe rename parameter name from topicURLRoot to topicRoot or topicPrefix

Issue #276, opened 2018-10-03
Add simple helm chart for deployment of Egeria to kubernetes


planetf1:
In #273 a simple docker image was added which is self-build from source

Docker images work well for a single component of an application environment, but we need much more for a real environment. there are multiple moving parts such as egeria itself, kafka, perhaps ranger, atlas, gaian & others.

This issue will target creation of a basic helm chart to deploy egeria.

Given a working kubernetes cluster, and helm installed (tiller on the backend) - such as on docker for mac, IBM cloud private or many other environments, the intent is that such a chart could be deployed with

helm add repo xxxx
helm install egeria-basic-cluster

or similar, with a 'values.yaml' optionally offering additional configuration

Also of note - #177 added instructions on creating a 2 node in-memory repository using docker compose. This works well in a local environment, but is more awkward for deployment to cloud environments, and not portable across all.

Once a simple proof point is in place we can consider additional charts (which can depend on other charts) , points of variability, and the publishing of required images to make for an easy end user experience be it for demos, development, education etc.

planetf1:
Added simple egeria chart & documentation

Note - currently whilst this deploys Egeria in-memory, and kafka - they are not immediately configured to work together. this still requires use of the Admin API

I have also checked in charts for
 * ranger -- we already used the prior docker version in previous demos
 * vdc-open -- this is work in progress, to incorporate other components needed for vdc demo. It will not contain non-open components. It may be split into further charts.

Neither of these are quite as clean, and need a few updates but are included

PR #328, opened 2018-10-26
#276 checkin initial helm charts for egeria deployment to kubernetes

planetf1:

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/category/Category.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/category/Category.java
@@ -15,6 +15,10 @@ public class Category extends GovernedNode{
         nodeType = NodeType.Category;
     }
 
+    /**
+     * The owning glossary for this category.
+     * @return <code>GlossarySummary</code>
+     */
     public GlossarySummary getGlossary() {
         return glossary;
     }
@@ -23,6 +27,10 @@ public class Category extends GovernedNode{
         this.glossary = glossary;
     }
 
+    /**
+     * "Identifies the parent category."
+     * @return <code>CategorySummary</code>
+     */
     public CategorySummary getParentCategory() {
         return parentCategory;
     }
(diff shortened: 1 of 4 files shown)


planetf1:
The paths that refer to planetf1 are not local files, they are pointers to a repository on docker.io, which is the default docker repository

Ideally I would like us to build docker images as part of egeria, or an associated project. Even better if we could do that under a 'odpi' id on docker hub, but that is going to take a while. We could also not publish the images anywhere except internally to the build environment/odpi infra, but that leaves a lot more work for anyone else to do.

The expedient approach was to post 'good enough' (and open) images to docker.io which anyone can use. This is what I've done so far for egeria, gaian, ranger. It will allow us to get started, learn by experience, invite broader collaboration, and quickly get to the point where we can deploy an egeria centric infrastructure with minimal effort - which is an acute challenge right now 

The references to myapp.name as to a variable. This is defined within each chart, and follows what appears to be a 'best practice'. Each kubernetes 'artifact' is tagged with the basename of the app uised to create it, and a component identifier. This makes it easier to search for, especially if the same k8s environment is hosting many many instances of egeria 

Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

The value of myapp is determined in _helpers.tpl . The default will be generated automatically by k8s. this is where we get the 'odd' names for a deployment as seen in the README.md. That is 'normal' behaviour. I think that makes it good for now - without duplicates (though am still working on testing).

Note I've not completed the ranger/vdc setups yet.

planetf1:
@bogdan-sava hi can you take a look and see if the response answers your questions? Thanks for reviewing

planetf1:
An addition on the choice of repository.

From https://kubernetes.io/docs/concepts/containers/images/
There are a number of solutions for configuring private registries. Here are some common use cases and suggested solutions.

Cluster running only non-proprietary (e.g. open-source) images. No need to hide images.
Use public images on the Docker hub.
No configuration required.

For getting started this made it the obvious choice. The same doc covers use of other container repos.

planetf1:
Here's the links to the 'experimental' images thus far
https://hub.docker.com/u/planetf1/

planetf1:
@bogdan-sava Hi - could you check if my comments satisfied your concerns?

bogdan-sava:
Yes. Fair enough. It would be great if you can show a demo of how to use.

PR #329, opened 2018-10-26
maven build for UI

bogdan-sava:


bogdan-sava:
The build runs OK on windows.
On linux VM fails at polymer build step.
Need to test it on linux.

PR #332, opened 2018-10-28
maven build UI

bogdan-sava:


bogdan-sava:
The build is working on linux ubuntu.
On jenkins VM still fails.
The error is:
/usr/bin/env: node: No such file or directory
[ERROR] Command execution failed.

Unfortunately the jenkins workspace is immediately deleted

planetf1:
I suggest contacting helpdesk@odpi.org - or raising a github issue against the odpi integration project? 
If you have a PR request I can also try a manual build on my linux env if it helps

bogdan-sava:
Thanks,
You can clone from my fork, the dev branch.
https://github.com/bogdan-sava/egeria

planetf1:
I get the exact same error
[INFO] --- exec-maven-plugin:1.6.0:exec (Polymer build) @ access-services-user-interface ---
/usr/bin/env: ‘node’: No such file or directory
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 127 (Exit value: 127)

planetf1:
recheck

planetf1:
I think I've figured out an example of a solution

* I noticed you were auto-activating the windows profile. This isn't necessary - maven should do this for us
* I added the 'node' command as the actual command for the build, with the polymer-cli .js script as the first parameter - for linux & mac only (additional os activated profiles)

With this it seems to work on linux. Not tested mac or windows yet.
Also we may want to rework the profile for 'unknown' OSs. Probably only windows should be different
I think some of the downloads still happen even if run a second time without clean, which seems unnecessary
Can discuss more next week.
Will post PR shortly (or patch)

bogdan-sava:
Thank you Nigel. I'll try it.

PR #358, opened 2018-11-07
ui build setup

bogdan-sava:
Signed-off-by: bogdan-sava <bogdan-mihail.sava@ing.com>

### DIFF
--- a/open-metadata-implementation/user-interfaces/access-services-user-interface/src/test/java/org/odpi/openmetadata/userinterface/accessservices/service/AssetCatalogOMASServiceTest.java
+++ b/open-metadata-implementation/user-interfaces/access-services-user-interface/src/test/java/org/odpi/openmetadata/userinterface/accessservices/service/AssetCatalogOMASServiceTest.java
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: Apache-2.0 */
 package org.odpi.openmetadata.userinterface.accessservices.service;
 
 import org.junit.jupiter.api.DisplayName;


open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
```diff
@@ -68,21 +68,161 @@
...
+                                    <goal>repackage</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.rat</groupId>
+                        <artifactId>apache-rat-plugin</artifactId>
+                        <configuration>
+                            <excludes>
```
planetf1: Two thoughts
a) Are any of these generic enough that we could move to the top level rat configuration -- having them in one place makes them easier to manage?
b) Some items in rat ignore are candidates for git ignore too - if the build is run 'in situ' - could you consider adding appropriate entries there too?
(subject to constraints of scoping within profiles)

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
```diff
@@ -68,21 +68,161 @@
...
-                        <goals>
-                            <goal>repackage</goal>
-                        </goals>
-                    </execution>
-                </executions>
+                <groupId>org.apache.rat</groupId>
+                <artifactId>apache-rat-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>**/node/**</exclude>
```
planetf1: as above ref: rat

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/README.md
```diff
@@ -0,0 +1,86 @@
+# Polymer App Toolbox - Starter Kit
```
planetf1: Do we need this file - or was it copied in as part of the install?
if to keep, it needs a license and copyright - but it isn't ours

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/index.html
```diff
@@ -0,0 +1,102 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+
```
planetf1: We need to start adding copyright to all files

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package-lock.json
```diff
@@ -0,0 +1,12822 @@
+{
```
planetf1: Is this a generated file?

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "egeria-ui",
+  "version": "0.0.1",
+  "description": "A starting point for Polymer apps",
+  "author": "The Polymer Authors",
+  "license": "BSD-3-Clause",
```
planetf1: Why BSD-3 ? Is this the license for polymer itself? How does it affect derived apps?
If supplied by them, do we not pull it down automatically and not checkin
I assume instead it's a sample, modified for our needs - in which case polymer aren't the authors? (subject to license/attribution)

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/service-worker.js
```diff
@@ -0,0 +1,13 @@
+/**
```
planetf1: This again doesn't look like our source? Does it need to be checked in - I'll skip further references to this, but would be good to clarify, as if we are reshipping we need to figure how to meet terms of license

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
```diff
@@ -68,21 +68,161 @@
...
+                <!--
+                IMPORTANT!!!
+                activation with jdk is used because setting activeByDefault true is ignored if other profile is activated,
+                and we have windows profile active in order to set polymer extension foe windows.
+                https://issues.apache.org/jira/browse/MNG-3229
+
+                To deactivate this profile, use:
+                mvn clean install -P !user-interface
+                -->
+                <jdk>[1.8,1.9)</jdk>
```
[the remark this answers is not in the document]
bogdan-sava: [1.8,*)

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
```diff
@@ -68,21 +68,161 @@
...
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>com.github.eirslett</groupId>
+                        <artifactId>frontend-maven-plugin</artifactId>
+                        <version>1.6</version>
```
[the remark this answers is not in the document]
bogdan-sava: fixed

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/service-worker.js
bogdan-sava: License changed.
It is generated by the polymer starter kit.

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package.json
bogdan-sava: Changed.
No! It is generated like this by the polymer starter kit.
The purpose of the PR is to have a build with polymer from maven.
That's why, I made it with default app polymer template app.

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/index.html
bogdan-sava: done

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package-lock.json
bogdan-sava: yes.
npm recommends to be committed
https://docs.npmjs.com/files/package-lock.json

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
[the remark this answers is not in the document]
planetf1: thanks. good!

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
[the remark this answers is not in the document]
planetf1: I opened up #361 as a related issue, but we are ok to add the code in.

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
planetf1: On b) in particular - this is still an issue.

Since it could lead to files being erroneously checked in with 'git add .' I think it's important to fix before merge.

As an example on my mac I ended up with:

15:20 $ git status
On branch bogdan-sava-ui
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/images/
	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/index.html
	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/manifest.json
	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/node_modules/
	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/service-worker.js
	open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/src/
	open-metadata-implementation/user-interfaces/access-services-user-interface/tools/


The .gitignore needs updating.

The RAT specs aren't critical - in fact it may be it's a good idea to keep them local - so happy to skip on that one.

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
planetf1: resolving the rat only issue for now

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/index.html
planetf1: There's also pom.xml -- though I see no precedence in the top level pom, on that basis no reason to block this PR for now. We probably need to scan/act on any report and fix up globally in the tree. will discuss/raise seperate issue

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package.json
planetf1: I was questioning the license -- if indeed it is 'their' code that we've modified I think we would have to stick with their license, and also add the necessary license text into the egeria top level docs.

If it's only in master temporarily, perhaps we should raise an issue to track the deviance .. acting as a reminder before any release? Or we go back to BSD-3. Any comments @mandy-chessell ?

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/service-worker.js
planetf1: as above (ok to close this issue as above one tracks it)

planetf1:
A quick summary from the review. I think we have 2 aspects to tidy up before merging
a) What do we do about the license. We 'should' leave it at BSD-3 , include license in the top level .. this should involve a broader team discussion. 
b) Can we fix up the fact there are additional files not excluded by git after a build, which could easily lead to contamination ie checking in those generated files in error - as many ppl will blindly do 'git add .' ...

planetf1:
are the license issues now resolved ? It would be good to get the UI into the master build so it can be used for demos etc (at least this PR forms the basis of that)

PR #359, opened 2018-11-07
Remove the security bypass in the code, which is redundant.

wyf-1116:
Now the security bypass setting should be configured in application.properties.


GuyAckermans:
Hi could you bring it up to date with latest master?

innerTide:
@GuyAckermans Done :)

PR #379, opened 2018-11-13
Port Sample Data scripts to postgresql

planetf1:
See #378 for details


planetf1:
thanks - fair comment. the script is more an example -- there's not enough error checking either

PR #410, opened 2018-11-23
Initial REST connector for IGC

cmgrote:
This is the library and related POJOs for a general connector to IGC via its REST API -- focused initially on read operations (ie. retrieve, search, paging, etc).


cmgrote:
Re-opening now that all files (except .iml) have Apache v2.0 license header.

cmgrote:
Would also really appreciate if someone can also review the `pom.xml` to ensure that this is standalone-able as an artifact?

PR #422, opened 2018-11-30
Update OMRS Event processing to handle batch events properly

mlottesen:
Signed-off-by: Mark Ottesen <mottesen@us.ibm.com>


mlottesen:
Went ahead and made the changes we discussed two Fridays ago such that a batch event is valid for processing if ANY entity or relationship in the batch is valid, and only send those that are.

PR #442, opened 2018-12-07
Tag propagation docs

planetf1:
Update documents for tag propagation as presented and discussed at the 4/5 dec meeting
Note: Further work is needed to complete the proposal of how tag propagation should work ultimately, and to clarify what compromises may be made in the near ter,m

pushing this PR so the docs represent what has been presented so far


planetf1:
recheck

PR #484, opened 2018-12-20
Ranger dependency updates

planetf1:
Pull forward dependency versions for ranger gaian plugin
Minor changes to support ranger 1.0 & greater (moved to 1.2 as latest public code level)

NOTE: Leave as PR until tested. PR needed to verify CLM build


planetf1:
@danielaotelea It would be really useful if, since you're currently using ranger, you could verify if the plugin still works with this more recent code. I'd like us to verify it before committing the PR ideally (a PR was needed to force the CLM build). Ideally we should also be using Ranger 1.2 as our server, rather than 0.7 as this is the latest shipped version

planetf1:
recheck

Issue #439, opened 2018-12-07
Fix licenses in ranger gaian plugin


planetf1:
As per email

There are 8 files within the /open-metadata-implementation/adapters/governance-engines-plugins/gaian-ranger-plugin/ directory, which have a different style of Apache-2.0 license header. This version of the license header is used by projects that are actually hosted by the Apache Software Foundation; it makes reference to the files being contributed to ASF via a CLA agreement. (Example: https://github.com/odpi/egeria/blob/master/open-metadata-implementation/adapters/governance-engines-plugins/gaian-ranger-plugin/src/main/resources/conf/ranger-gaian-security.xml)
Do you know whether these files did in fact originate from an ASF-hosted project? If so, then it likely makes sense to keep the existing header.
But if these files are instead original to Egeria, and were not from an ASF project, then this is likely the wrong license notice to use -- and should likely be replaced with just an SPDX identifier and/or the "standard" Apache-2.0 header text.

planetf1:
Licenses corrected
removed 2 unused files (for geo)

planetf1:
run-clm

PR #485, opened 2018-12-20
ranger gaian plugin : #439 fix license labels

planetf1:


planetf1:
@bogdan-sava Thanks for spotting. I had incorrectly added the license twice (and omitted copyright). unfortunately for now this still passes the RAT check which was tested. now fixed

PR #539, opened 2019-01-10
updates to site design

alexcontini:
Signed-off-by: Alex Contini <acontini@linuxfoundation.org>

@mandy-chessell


cmgrote:
Hi @alexcontini -- as best I can tell the check is failing because there is no license header in the `.scss` files...  Is it possible to add something like this to the top of the 2 `.scss` files (`assets/css/style-sub.scss` and `assets/css/style.scss`)?

```
/* SPDX-License-Identifier: Apache-2.0
 * Copyright Contributors to the ODPi Egeria project. */
```

@planetf1 I'm assuming that `.[s]css` should be treated as "code" and not ignored by the license check, right?

alexcontini:
@cmgrote I just updated the files, let me know if that doesn't work!

Issue #543, opened 2019-01-11
IGC Repo Connector - NEW_RELATIONSHIP event issue


ruxandraRosu:
InstanceProvenanceType of relationship is null thus atlas doesn't persist it.
The entity proxy1 type is the supertype Referenceable, should be the actual type of the entity proxy (RELATIONAL_COLUMN etc.)

cmgrote:
Thanks!  Just testing out these changes on my side before putting into a PR...

cmgrote:
Just added PR #544 to address these problems.

ruxandraRosu:
@cmgrote relationships are stored in atlas now so this issue can be closed once PR is merged

PR #544, opened 2019-01-11
Fixes event mapper issues in #543

cmgrote:
Adds provenance type to relationships and actual entity type for entity proxy (not just supertype as previously).

Also sets up AttributeTypeDefs properly as part of initialisation of the IGC connector now, too.


cmgrote:
^^ @ruxandraRosu @bogdan-sava

cmgrote:
Is there a way I can force a rebuild?  I'm building the same branch locally, with same commits, and it builds fine -- not clear where the compilation error is occurring...

ruxandraRosu:
I've tested branch igc-repo-conn and the relationships are added. There are some issues with UPDATED_ENTITY and UPDATED_RELATIONSHIPS events, these are investigated on atlas connector side.

cmgrote:
(Not sure why it requires significant resources -- in my compiled target for the referenced location it's 36MB ?)

cmgrote:
Will attempt to play with compiler plugin changes and re-test...

Issue #344, opened 2018-11-02
Add Sample Data to vdc-open chart


planetf1:
Add an additional pod to vdc-open helm chart with
 - postgress
 - OCO Pharmeceuticals sample data (unlike the charts thus far, this will require a persistant volume claim)

mstrelchuk:
In order to use Coco Pharmaceuticals sample datasets for BI use-cases we need to have Primary Keys and  Foreign keys defined on those datasets. Below you can see how Coco Pharmaceuticals tables are connected.

![pk and fk coco pharm](https://user-images.githubusercontent.com/40057799/49378881-0e84ff00-f70e-11e8-8975-ba84f709478f.png)

planetf1:
Also I will create a static gaian config for an interesting table to assist in testing (prior to virtualizer integration), but other views will be configured via the virtualizer (when ready)

planetf1:
Ok.. I have a postgres container. I am working on the egeria config currently, but will aim to get sample data loaded into that container. 

The sample data is at https://github.com/odpi/egeria/tree/master/open-metadata-resources/open-metadata-deployment/SampleData though in my latest PR I have moved that slightly (though not changed the files)

PR #556, opened 2019-01-14
Sample (meta)data

cmgrote:
Reworks the sample (meta)data to fully populate an IGC environment for demo purposes (including README on how to deploy it).  Should address significant portion of #344 but still need to move over alternative database backends for sample data (mariadb, etc).

These should simply be a matter of creating a new sub-directory under `database` with the scripts specific to that database software, and then running them is simply a matter of changing one variable in `defaults/main.yml` -- and it should automatically be picked up by the deployment.

In order for metadata ingestion to still setup all the correct relationships between terms and assets, we need to follow the same naming for database objects, tables, etc -- so when porting over these other databases I'd suggest we conform to what's been put in place for the `ibm-db2` sample (basically shortening the name of `CompanyDirectory` to `CompDir` and `EmpSalaryAnalysis` to `EmplSAnl`).

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/SubjectAreaErrorCode.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/SubjectAreaErrorCode.java
@@ -196,7 +196,7 @@ public enum SubjectAreaErrorCode {
             "Retry the request with a known relationship."),
 
 
-    // The following Exceptions are not descriptive enough - they need more informaiton from the OMRS Exception
+    // The following Exceptions are not descriptive enough - they need more information from the OMRS Exception
     INVALID_PARAMETER(400, "OMAS-SUBJECTAREA-400-046 ",
             "Invalid parameter.",
             "The system is unable to process the request.",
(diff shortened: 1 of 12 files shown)


cmgrote:
> I think ideally it would be good if we could split the sample data (which is independent from IGC) from IGC specifics. Whilst we are working a current demo using igc, we also want to have an all open-source configuration - yet will still need the sample data. ie IGC is very much one specific option that can be used as a metadata repository

Agreed -- this is why I've put the IGC-specific stuff into its own sub-directory (`ibm-igc-metadata`), the generic things like data files are under a non-specific directory (`data-files`) and the `databases` have space for additional sub-directories for each of the different databases we want to provide samples for (just `ibm-db2` for now as I haven't had a chance to port over and test the others, yet).

> Also note that there is a directory 'SampleData' which contains similar content -- and has recently added a subdir 'CocoPharma' since we may ultimately have a variety of sample data - I think it would be good to consolidate - or at least clarify via the name

I renamed directories to fit in with the convention used everywhere else: hyphenated-lowercase rather than CamelCase, and reorganised to fit with my thinking above on splitting generic and product-specific pieces.  The content of eg. the sample data files is identical, but I've renamed to provide consistency regardless of the backend product choices.  (And have implemented some primary keys and foreign keys in the DB2 sample tables in order to help other objectives and provide some examples of these classifications and relationships for eg. OMAS interfaces that might want such samples.)

I've also left `coco-pharmaceuticals` as a sub-directory of the general `data-samples`, so that indeed we can come up with other sample "packs" and provide a listing in the `README.md` at the `sample-data` level, with separate pack-specific READMEs in the respective directories providing any details for those specific samples (like the `README.md` under `coco-pharmaceuticals`).

cmgrote:
Latest update includes this split of IGC-specific portion into its own subdirectory under a more general `metadata` subdirectory -- and removes all IBM Information Server-specifics to the load script embedded in that subdirectory.  This should allow us to put alternative options there, just like with `database`, but still drive the loading across all of these from the `deploy.yml` (based on a couple of configuration variables in `defaults/main.yml`)

cmgrote:
@planetf1 are we OK to merge this?  I can then work on porting over the other database backends as well (and testing that they work with the automation).

PR #597, opened 2019-01-22
Ensures modification details are retrieved for IGC Repo Connector

cmgrote:
To fix empty createTime, updateTime, etc details on relationships and EntityProxies.


ruxandraRosu:
I've checked for a semantic assignment and atlas it's not throwing an exception anymore, create/updateTime are on both entities and on relationship as well

cmgrote:
Thanks!  Will merge as soon as branch is back up-to-date and checked.

Issue #602, opened 2019-01-22
bug in the the in memory repository find property value implementation


davidradl:
The code only will return entities that are deleted.

PR #603, opened 2019-01-22
#602 fix find property value in memory bug

davidradl:


cmgrote:
Before you merge, can you correct it in the other places, too?

PR #629, opened 2019-01-24
Fix debug logs

bogdan-sava:


bogdan-sava:
recheck

bogdan-sava:
It is not added. It was there.
Basically there were two implementations for slf4j. Using log4j and logback.
Just picked logback as it is better in my opinion.

cmgrote:
> It is not added. It was there.

Sorry, wasn't obvious from the diff but I see that now.

mandy-chessell:
The logback library was already in use in the IV OMAS.  I moved it to the top level pom yesterday to clear the SLF errors in the build.  This seems to have impacted the Spring server chassis.

The EPL license is OK - I do not understand the implications of a dual license. - Will check with Steve.  It is OK to merge the change.

If debug coming out on stdout by default then need to fix that.  However, it is OK for debug to come out on stdout if explicitly enabled.

bogdan-sava:
I don't think they were intentionally removed. They are debug log.  It was caused by the moving 2 dependencies ( log4j-to-slf4j and logback-classic ) in the parent pom to be available on all the modules.
Those 2 scope was test so they were only available for tests not included in build and no more logging lib for application.

pom.xml
```diff
@@ -137,6 +137,11 @@
                 <artifactId>plexus-utils</artifactId>
                 <version>3.1.1</version>
             </dependency>
+            <dependency>
+                <groupId>ch.qos.logback</groupId>
+                <artifactId>logback-classic</artifactId>
+                <version>1.2.3</version>
+            </dependency>
         </dependencies>
     </dependencyManagement>
```
mandy-chessell: Many dependencies appear to be outsince of the dependency management tag - is this right?

bogdan-sava:
Both 
log4j-to-slf4j (logging bridge) and logback-classic are compile dependencies of spring-boot. 
Both were there since spring-boot is used. 
Only the dependency scope was changed and caused log missing.

It was a pure maven configuration issue.

pom.xml
bogdan-sava: Yes. 
We start using it recently. DependencyManagement should be used carefully. It helps with the control of versions of artifacts used in transitive dependencies.

PR #659, opened 2019-01-28
Adds asset property mappings for DataFile + fixes OMRS bundle upload

cmgrote:


cmgrote:
@ruxandraRosu I think so far we're only using database tables & columns in our work (?)  If we end up using data files at some point as well, this update is necessary to populate their `name` and `description` properties (now that they're inheriting from `Asset` instead of only `Referenceable`).

PR #677, opened 2019-01-31
Fix links on sidebar

jmertic:


_layouts/default.html
```diff
@@ -20,21 +20,21 @@
          <div class="sideBartext">Open Metadata and Governance</div>
          <a href="{{ site.github.repository_url }}"><button>View on Github</button></a>
          <div class="links">
-            <a href="/egeria/">Home</a>
```
mandy-chessell: The changed shown here are not sufficient to correct the change of URL to the GitHub pages.  The link is used in the code as well as many markdown files.

jmertic: Talked about this today - sounds like it took a bit for GH URL changes to make it's way through ;-)

I did make an adjustment however to which should resolve in the odd case of a different url. Let me know what you think.

PR #684, opened 2019-01-31
Adds generated deployment directories for UI to ignore file

cmgrote:
After a `mvn clean install` these directories show up as unknown to git:

- `open-metadata-implementation/user-interfaces/access-services-user-interface/tools/`
- `open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/`

I'm assuming we want to re-generate them each time (ie. not check them into git), hence adding to `.gitignore` (?)


.gitignore
```diff
@@ -88,6 +88,10 @@ open-metadata-resources/open-metadata-deployment/sample-data/coco-pharmaceutical
 open-metadata-resources/open-metadata-deployment/sample-data/coco-pharmaceuticals/*.retry
 open-metadata-resources/open-metadata-deployment/sample-data/coco-pharmaceuticals/metadata/ibm-igc/generated_*
 
+# Ignore UI-generated / deployment content
+open-metadata-implementation/user-interfaces/access-services-user-interface/tools/**
+open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/resources/public/**
+
```
bogdan-sava: please add these too

open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/node_modules/**
open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/static/package-lock.json

PR #689, opened 2019-02-01
Splits samples credentials between database and files

cmgrote:
Allowing a split of the samples across different systems / containers / hosts.


cmgrote:
Hold on -- just found something I missed in the update

PR #691, opened 2019-02-01
Helm12

planetf1:
Updated to now correctly configure postgresql within cluster


planetf1:
Atlas is also updated with latest fixes (including for null metadataCollection)

planetf1:
DO NOT MERGE. Investigating summary report of 31000 files..

PR #732, opened 2019-02-07
Deployment tutorial update

cmgrote:
Updates the walkthrough of the VDC deployment based on Nigel's walkthrough in this morning's call.  Probably still some cleanups we need to do in the configuration itself (eg. hard-coded password / authentication methods), but this should be a start...

(Also fixes a couple minor bugs in the cleanup scripts for the Coco Pharma samples metadata for IGC.)


cmgrote:
I've tried to address the hostname hard-coding, but won't be fully-resolved until we can release updated Docker images (currently not possible per #739)...

Issue #670, opened 2019-01-29
Consider switching to adoptopenjdk/alpine base images for java apps in charts


planetf1:
See https://hub.docker.com/u/adoptopenjdk & https://developers.redhat.com/blog/2017/03/14/java-inside-docker/

Currently the Egeria vdc docker images are based on standard ubuntu & the ubuntu distributed openjdk & generally work.

However the images above may provide a better base environment that is more container aware in terms of resource limits within the container.

cmgrote:
Started on this for Egeria itself with #741 -- think we need to do the same for other images as well.

cmgrote:
Since I closed the PR to cater for a variety of approaches, adding discussion back here:

> I've tried to add containerization based issues into https://github.com/odpi/egeria/projects/8
#670 refers to wanting to switch to alpine to make the images smaller, but there are also images from AdoptOpenJDK which I haven't looked at yet.

The approach I took was to use OpenJDK's official image, one of which uses Alpine as its own base (so it's likely as minimal yet still "official" as they'll come).  This resulted in an Egeria OMAG Server Platform image at 130MB, same for the Egeria UI image, and an Apache Atlas image from Graham's managed fork (including embedded HBase and Solr) at approximately 1.3GB.  (That's down from 2.56GB and 10.3GB respectively -- so 10x+ reduction...)

> For the images themselves, as well as the base image to draw from there is refactoring needed in the scripts -- but I still have a few components to go, so will review based on that experience.

I think it makes sense to break the whole process down into different steps, like:

1. Build the component (Egeria, Apache Atlas, Apache Ranger, etc) as vanilla -- granted this could be just taking a release from Maven
1. Package the runtime into an image (without all the build toolchain, so 10x+ reduction in size)
1. (Use this runtime image (only) in all subsequent steps)
1. Define the composition of these runtime images (eg. via Helm charts).
1. Orchestrate the configuration of those runtime images (this bit I'm less opinionated on so far, as haven't seen a great approach for this yet; but the method at the moment of embedded init containers seems messy and high maintenance to me?)

> On the build - no question. But we need ODPi engagement to get the generic id and build support. Some could be built within the build, though a seperate process is better. Versioning will need to be automatic and that needs to extend to the helm chart definitions tool We will need to push to a public repo which may or may be be dockerhub (it was a tactical option to get something working quickly)

👍 

> It's also worth considering how we build the images.

[...]

PR #741, opened 2019-02-08
Initial source-to-image container builder for OMAG Server Platform

cmgrote:
This approach reduces the container image for the OMAG Server Platform from ~2.5GB to less than 150MB (using just the JRE and Alpine rather than a full Ubuntu build environment).  Using the commands outlined in the `README`, maybe it's even something we could add to our Jenkins pipeline to build a "nightly" snapshot from master to auto-publish into the public Docker hub under an official ODPi namespace?

(And of course we could do release-based publications for releases.)

We probably want to investigate a similar approach for the rest of the images, to significantly reduce the overall container footprint?


cmgrote:
Actually, will look at re-factoring this further to allow production of other images (not just OMAG Server Platform).

planetf1:
I have started a discussion with the ODPi build team 

I've tried to add containerization based issues into https://github.com/odpi/egeria/projects/8
#670 refers to wanting to switch to alpine to make the images smaller, but there are also images from AdoptOpenJDK which I haven't looked at yet.

For the images themselves, as well as the base image to draw from there is refactoring needed in the scripts -- but I still have a few components to go, so will review based on that experience.

On the build - no question. But we need ODPi engagement to get the generic id and build support. Some could be built within the build, though a seperate process is better. Versioning will need to be automatic and that needs to extend to the helm chart definitions tool We will need to push to a public repo which may or may be be dockerhub (it was a tactical option to get something working quickly)

It's also worth considering how we build the images.

For egeria, from source seems a good option - though we also need images from other projects where we either have source or don't. This could be at docker build time, or for more flexibilityt at deployment time - but that is awkward if the build time is extensive. 

Use of the maven artifacts is my favoured mechanism - as I originally tried with IGC, but it didn;t quite work. That could *optionally* allow the image to self-update at deployment time IF THE USER WISHED - most likely for the development use case.

My hope was to learn a little more at think, and then us do some research on the different possibilities, frameworks. IBM has microclimate for example -- but something open is needed (or there needs to be multiple options)

It's aj intriguing area and certainly one we can make many improvements. consider what we have today version 0.1 :-)

PR #742, opened 2019-02-10
Updates to OMRSRepositoryConnector

mandy-chessell:
Two changes to the OMRSRepositoryConnector:
 * Abilty to throw FunctionNotSupportedException from instance maintenance methods.
 * New parameters for find instance methods to separate fuzzy match from exact match specification.

The design has to cope with different repository connectored with different levels of support.

omrs-repository-connectors

Apache Atlas represents OMRSRepositoryConnectors outside of the Egeria build running on back level Egeria. @grahamwallis likely to update Atlas soon but other OMRSRepositoryConnectors will still exist.


cmgrote:
Thanks for adding these!  Hoping to have a more in-depth look tomorrow.

grahamwallis:
With regard to the colour coded levels of support - would it help if instead of using the word 'supports', we were to use the words 'exposes' and 'invokes' to indicate what level of API is exposed by each connector module and what level of underlying connector(s) it can invoke? The 'pink' colour code comes closest to this separation but I think it is still ambiguous as to which direction the 'translation' is taking place.

With regard to the lower example repository connector leg ('Apache Atlas'), would there not a problem coupling the 'green' OMRSRESTConnector (which supports just the new APIs to a 'yellow' OMRSRESTServices (which only supports the old APIs). As stated, the Atlas 'leg' can be updated soon after this change to Egeria, but I don't see how another fully yellow leg would continue to work without being updated. I may be missing something but I would have expected the OMAG Server to turn largely pink or blue for the time being (until the deprecated APIs are removed).

ruxandraRosu:
Hi,

I've done some tests calling findEntitiesByProperty with exactMatch(by qualifiedName) and fuzzyMatch(null) arguments. 
On atlas side the method call reaches OMRSMetadataCollection - findEntitiesByProperty where RepositoryErrorException is thrown as it is not yet implemented. 
The deprecated call (only fuzzyMatch) it's throwing the same exception as it reaches the same method.

mandy-chessell:
@cmgrote I can change the methods you suggest for FunctionNotSupported.  This design change did not change any of these methods - so this would be another change in behaviour - but a logical change?

mandy-chessell:
@ruxandraRosu   Am I understanding that the RepositoryErrorException is thrown within the Atlas Server from the original code, whereas before this change, the call worked?  So it is as if Atlas is receiving a different call?

cmgrote:
> This design change did not change any of these methods - so this would be another change in behaviour

Ah, I see...  I'd still be pro-change on this one, to get the default behaviour into v1.0 that if you don't override these methods your connector automatically indicates that it doesn't support those functions (?)  Guess the main question is to those it's going to hit (OMASes?) whether it would be something we could handle fixing on the consumption side of the methods in a short timeframe?

mandy-chessell:
@grahamwallis the OMASs are using a mixture of old and new calls so they are in white.  The ability for a green OMRSRESTMetadataCollection to call a Yellow remote partner is a requirement, and possible to make work - although I realize I have made a coding error by changing the property names in PropertyMatchFindRequest.  I have changed them back to their origina names.  This is probably the cause of the problem that @ruxandraRosu is seeing.

ruxandraRosu:
The issue I have is caused by the changes done in OMRSMetadataCollection - egeria code. findEntitiesByProperty with both match and fuzzy args is always throwing exception, it is not calling the localMetadataConnector.

cmgrote:
> findEntitiesByProperty with both match and fuzzy args is always throwing exception, it is not calling the localMetadataConnector.

I guess the underlying connectors (ie. Atlas in your original example) still need to be updated to handle _both_ match criteria...  Or are you saying this is the case for in-memory repo as well?

(Sounds like Mandy's change back to the original parameter names should at least mean the original calls, with only `matchCriteria` should work again?)

mandy-chessell:
@cmgrote the aim is that the underlying connectors should not need changing for the old behaviour to work.  They will not be able to support the new function obviously.  However the local OMRS connector should compensate for that as much as is possible.

open-metadata-implementation/repository-services/repository-services-apis/src/main/java/org/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/OMRSMetadataCollectionBase.java
```diff
@@ -15,6 +15,7 @@
 
 import java.util.List;
 import java.util.Date;
+import java.util.function.Function;
```
cmgrote: Are we using this?  Or just hit `Enter` too quickly (as I often do) in IntelliJ and got this magically? 😉

mandy-chessell:
I have changed the calls that @cmgrote listed to throw FunctionNotSupportedException in the base class.  This means that connectors such as the IGC one will not need to override these methods just so they can throw the right exception.  Ideally they should only need to override the methods that are supported.  It should be possible to provide support for the TypeDef/AttributeTypeDef query methods as long as the verify methods are implemented correctly, further reducing the effort to write a new connector.  I think that should be a different PR.

mandy-chessell:
@cmgrote removed Function import - was special present from IntelliJ.

mandy-chessell:
@ruxandraRosu ah - so the exception is not happening the in the Atlas server but locally?

mandy-chessell:
@ruxandraRosu could you send the details of the exception and message.  I am puzzled as all of the Egera OMRSRepositoryConnectors seem to override the new findEntitiesByProperty method so the exception it throws from OMRSMetadataCollection should not occur.   (OMRSMetadataCollection should not call any other connector - its methods are overridden by the implementing connectors.  We should only see its behaviour where the implementing connector has not overridden the method - so we need to find out which connector implementation has not got this overriding implementation.

ruxandraRosu:
Exception message is: EntityListResponse{entities=null, nextPageURL='null', offset=0, pageSize=0, relatedHTTPCode=501, exceptionClassName='org.odpi.openmetadata.repositoryservices.ffdc.exception.RepositoryErrorException', exceptionErrorMessage='OMRS-METADATA-COLLECTION-501-001 OMRSMetadataInstanceStore method findEntitiesByProperty for OMRS Connector org.apache.atlas.openmetadata.adapters.repositoryconnector.LocalAtlasOMRSMetadataCollection to repository type atlas is not implemented', exceptionSystemAction='A method in MetadataCollectionBase was called which means that the connector's OMRSMetadataInstanceStore (a subclass of MetadataCollectionBase) does not have a complete implementation.', exceptionUserAction='Raise a Github issue to get this fixed.', exceptionProperties=null}

ruxandraRosu:
Upper stacktrace section:
0 = "org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.OMRSMetadataCollection.findEntitiesByProperty(OMRSMetadataCollection.java:963)"
1 = "org.odpi.openmetadata.repositoryservices.localrepository.repositoryconnector.LocalOMRSMetadataCollection.findEntitiesByProperty(LocalOMRSMetadataCollection.java:1594)"
2 = "org.odpi.openmetadata.repositoryservices.rest.server.OMRSRepositoryRESTServices.findEntitiesByProperty(OMRSRepositoryRESTServices.java:2043)"
3 = "org.odpi.openmetadata.repositoryservices.rest.server.spring.RepositoryServicesResource.findEntitiesByProperty(RepositoryServicesResource.java:873)"

grahamwallis:
OK- I think with the reversion to original matchProperties parameter naming I can see how this works. Just to be sure can I just check... (sorry I am fighting off a cold and migraine so probably only operating at about 50%) ... the OMRSRESTMetadataCollection is packing all the parameters (existing and new) into the REST request body, and the pre-existing (old, yellow version) OMRS REST Service will only unpack the things it understands, which will be the existing matchProperties. Is that correct?

The change I will make in the Atlas OMRS Connector will build off the updated Egeria master so the whole 'Atlas-leg' will turn 'green'. i.e. it will pick up the new OMRS REST Service, Local OMRS Connector and my updated Atlas OMRS Connector.

For a.n.other back-level repository connector, that is not updated to use the most recent Egeria, the (yellow) OMRS REST Service will unpack the matchProeprties, pass them through to the (yellow) Local OMRS Connector and then to the (yellow) 'A.N.Other OMRS Connector' - which both treat them as 'fuzzy'.

One nit-pick - although we are using the term 'fuzzy' we are actually doing a substring match rather than a true fuzzy match.

cmgrote:
> One nit-pick - although we are using the term 'fuzzy' we are actually doing a substring match rather than a true fuzzy match.

Good point -- +1 to this 👍

mandy-chessell:
@cmgrote more than substring match as it handles numbers, maps and arrays.

planetf1:
I was able to run the current VDC helm chart with a special egeria build of this PR. Types/entities are being populated from VDC into Atlas via our 3 party cohort. I've not had time to do any thorough checks but if anyone wants to check, I can provide cloud URLs.

grahamwallis:
@mandy-chessell @cmgrote With regard to the 'fuzzy' discussion: I believe the current (pre #742) behaviour is that anything other than a string will be subject to an exact match, whereas String properties will be tested using a substring predicate. This is for find*ByProperty(). For find*ByPropertyValue() the searchCriteria parameter is only applied to String properties. My interpretation (which may be incorrect) is therefore that the current (pre #742) behaviour has distinct semantics for String properties - i.e. it is only Strings that are soft-matched, all other types of property are exact-matched. And the soft-matching of strings is substring not fuzzy.
In the post #742 implementation:

- Are we aiming to perform soft-matching for types other than String? If yes, then what the of soft-matching are we looking for?

- Are we changing to performing fuzzy string matches, rather than substring matches?

I'm just trying to nail down what the exact specification is. 

From an implementation point of view, some of these things should be  achievable with JanusGraph (although maybe not via Atlas) and some (e.g. contains, but not fuzzy) will be achievable with Tinkerpop 3.4.0.

cmgrote:
> anything other than a string will be subject to an exact match, whereas String properties will be tested using a substring predicate

Just double-checked my semantics:

`findEntitiesByProperty`:

- I'm breaking down non-string property values into their constituent pieces
- Any string (whether primitive or within a map, array, etc) is being substring matched
- Any non-string is being exact matched

`findEntitiesByPropertyValue`:

- Only searches against string properties (given description is "Return a list of entities whose _string based property values_ match the search criteria")
- All of which are substring matched

To be further pedantic, aside from just `substring`, what about case sensitivity?  I would assume that by "fuzzy" we're saying both substring (contains) and case-insensitive (?)

grahamwallis:
@cmgrote All good points and not pedantic.

For the Atlas connector I have not found a way to supply array or map values to the search processors. So the connector deals with primitives only (strings are treated as substrings, other types of primitive must be exact).

For find*ByPropertyValue - which uses searchCriteria, I request a search of the full text index - and this is case-insensitive.

Unfortunately, I cannot emulate that for find*ByProperty, which is more constrained and uses the specific property indexes (backing off to graph attribute checking if an index search is not possible). This is case-sensitive, and I can't currently see how to change that without some very deep changes in Atlas.

ruxandraRosu:
I am adding below the description for the current behaviour and the cause of the issue mentioned above:
All POST requests to findEntitiesByProperty (with or without fuzzyMatch) are handled by OMRSRepositoryRESTServices - findEntitiesByProperty.
This method calls localMetadataCollection.findEntitiesByProperty(with all params, including fuzzy) that further calls realMetadataCollection (LocalAtlasMetadataCollection)
Because LocalAtlasMetadataCollection extends OMRSMetadataCollection, but does not override findEntitiesByProperty with all the args(including fuzzy), the NotImplementedException is thrown.
Until the full implementation is done can we override in LocalAtlasMetadataCollection this method to call the old one in case fuzzyMatch is null(or change in OMRSMetadataCollection) so that we can still have old functionality when this is merged?

mandy-chessell:
@ruxandraRosu you are correct.  The current implementation of LocalAtlasRepositoryConnector is not compatible with this patch.  So it is not possible to run a mixture of this patch and the current Atlas connector in the same server.  This patch can be installed in an Egeria server that is calling an Atlas server using a back-level version of Egeria (as in the picture at the top of this patch).  Once this patch is committed then Atlas will need updating to build with the new master.

mandy-chessell:
@cmgrote @grahamwallis this is clarification on the search behaviour.

I am uncomfortable with the term "substring" as a way of describing the fuzzy search because it takes a regular expression allowing alternate values to be expressed as well as substrings.  It also allows the specification of whether it is matching on case or not.

What has been added is exact match support - this is where we use the equals() method to compare values rather then matches().   The previous function always assumed the property value is a regular expression on the match.  On the fuzzy match it also allows an exact match so that the deprecated version still works as before.

The implementation in the in-memory repository was flawed before this patch in that it only worked with string properties. Now it converts all property values to strings and performs either an equals or a matches (regex expression evaluation) on the result.  So now it is able to search for values in arrays, maps, structures and numbers.

The matching of properties that are arrays, maps and structures is done on the whole property value rather than breaking each of its elements into consituent pieces.  An array is converted into a string like this { value1, value2 }, a map to a string like this { key1 -> value1, key2 -> value2 } and a struct to { fieldName1 : fieldValue1, fieldName2 : fieldValue2 } and comparisons made on these resulting strings.

grahamwallis:
I would suggest we get together to discuss on a phone call.

cmgrote:
> I am uncomfortable with the term "substring" as a way of describing the fuzzy search because it takes a regular expression allowing alternate values to be expressed as well as substrings. It also allows the specification of whether it is matching on case or not...

Fair enough -- if the logic is indeed intended to be `match()` it makes sense to not restrict to just `substring`.  I must admit this is news to me, though!  I'm having trouble following where we would define these regular expressions...  The properties provided to the search methods are just InstanceProperties, having eg. xPropertyValues within them -- to handle regular expressions in general, we'd surely need to end up treating every property as a PrimitivePropertyValue of type String (as that's the only type that can represent a regular expression -- there isn't any "RegularExpression" PrimitiveDefCategory, and no regular expression Category in general?) Not sure how this would work when a property itself is defined as a Date or a Number (maybe we just treat everything in the matchProperties InstanceProperties is provided as String, assuming there's no validation that these properties match an actual defined property with the same name's type?)

This is a confusing overload of InstanceProperties to me: sort of feels like we might want a search-specific, regular expression-y means of passing this info to the findX methods...

> The implementation in the in-memory repository was flawed before this patch in that it only worked with string properties. Now it converts all property values to strings and performs either an equals or a matches (regex expression evaluation) on the result. So now it is able to search for values in arrays, maps, structures and numbers.

> The matching of properties that are arrays, maps and structures is done on the whole property value rather than breaking each of its elements into consituent pieces. An array is converted into a string like this { value1, value2 }, a map to a string like this { key1 -> value1, key2 -> value2 } and a struct to { fieldName1 : fieldValue1, fieldName2 : fieldValue2 } and comparisons made on these resulting strings.

This seems to be introducing a canonical string representation for these "complex" objects that must be adhered to within open metadata repositories, in order to do such an `equals()` or `match()` against (?)  I suspect that in practice this will mean repositories either need to store this canonical representation as a string directly (to optimally search against), or could otherwise only do a search by retrieving the entire metadata set, `toString()`ing the complex properties, and then iterating through every single object in Java -- which would be a performance disaster (?)

So storing a canonical string representation seems to be the only real choice.  Then again, even with that approach presumably we'd also need to cater for semantic similarity without an exact string match: eg.

- `{ key1 -> value1, key2 -> value2 }` is the same map as `{ key2 -> value2, key1 -> value1 }` (ordering being undefined in maps -- yet as strings they're not `equal()`)
- yet `[ value1, value2 ]` is not the same array as `[ value2, value1 ]` (ordering being important in arrays)
- a regex used for `match()` would presumably also need to handle this canonical representation to tokenize around commas, understand the bracket usage, potential ordering differences in unordered complex objects, etc

mandy-chessell:
We are out of time on this one - if it is to make the first release.  The search has always been a regex expression - we decided that when we agreed not to create a search language.  The regex to apply to each property value is entered in the property value.  This has not changed in this fix.  The request was to add the exact value match because of handling special characters in fully-qualified names.

I will change the name of the field from fuzzyXXX to regExXXX to make it clear and you are questioning the use of fuzzy search.  I will also add parameters for substring queries so they can be optimized in certain repositories.

cmgrote:
Agreed.

Apologies, I think I missed those earlier discussions on regexes. Personally I'm OK with the name change to regExXXX, but it does feel like we're rushing this bit. It also feels a bit strange that we have so many parameters for different kinds of criteria rather than one parameter which itself captures these different varieties of options within it (and any combination thereof, perhaps even including nesting possibilities?)...

The big change I was after to include in the first release was `FunctionNotSupported` -- which seems less contentious -- this search change was less critical for me. Others?

mandy-chessell:
I have now had time to read through the comments carefully and realise I made a mistake changing the API in the first place as the issue seems to be missing documentation on how the original interface works rather than a problem with its capability.   So I have removed the changes to the OMRSRepsitoryConnector find methods (including the new methods).

mandy-chessell:
New issues raised:
* #759 document find interfaces
* #756 add RegEx primitive type

PR #744, opened 2019-02-11
Fixes #743 JavaDoc issues

### LINKED ISSUE #743: Javadoc errors in IGC Client library
looks like Javadoc confused by the angle brackets

/Users/mandy-chessell/CloudStation/Drive/Code/ODPi/egeria-governance/egeria/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/igc-rest-client-library/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/igc/clientlibrary/model/common/Reference.java:545: error: unknown tag: String
     * @return List<String>
                   ^
/Users/mandy-chessell/CloudStation/Drive/Code/ODPi/egeria-governance/egeria/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/igc-rest-client-library/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/igc/clientlibrary/model/common/Reference.java:562: error: unknown tag: String
     * @return List<String>
                   ^
/Users/mandy-chessell/CloudStation/Drive/Code/ODPi/egeria-governance/egeria/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/igc-rest-client-library/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/igc/clientlibrary/model/common/Reference.java:579: error: unknown tag: String
     * @return List<String>
                   ^
/Users/mandy-chessell/CloudStation/Drive/Code/ODPi/egeria-governance/egeria/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/igc-rest-client-library/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/igc/clientlibrary/model/common/Reference.java:596: error: unknown tag: String
     * @return List<String>
                   ^

cmgrote:

### DIFF
--- a/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/SchemaElement.java
+++ b/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/SchemaElement.java
@@ -21,7 +21,6 @@ import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.PUBLIC_
  *         <li>SchemaAttribute describes the use of another schema as part of the structure within a bigger schema.</li>
  *     </ul>
  *     Assets are linked to a SchemaType.
- * </p>
  */
 @JsonAutoDetect(getterVisibility=PUBLIC_ONLY, setterVisibility=PUBLIC_ONLY, fieldVisibility=NONE)
 @JsonInclude(JsonInclude.Include.NON_NULL)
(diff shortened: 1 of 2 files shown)


cmgrote:
Since the smaller individual changes are likely lost in the update to the generated code, will outline here:

- @planetf1 a few of your JavaDoc headings used ampersand (`&`) -- this is problematic for HTML, so I've replaced with the word `and` (I think probably the convention we should go for in general, rather than needing to embed special html escapes like `&amp;` everywhere...)
- @mandy-chessell there were a few places where string descriptions / JavaDoc headings made use of a character encoding-specific apostrophe character, whilst elsewhere in the same file you just used a single quote -- I've replaced all with single quotes to avoid any errors generating the JavaDocs without explicitly providing the appropriate character encoding

There were also a couple places where an `@return` was specified for a `void` method (so just deleted the `@return`), where a `@throws Exception` was used (replaced with a number of `@throws` for each of the actual Exceptions that could be thrown), and one spot where the JavaDoc comment appeared below rather than above the object being commented.

mandy-chessell:
I think we have clashing fixes - I was hoping that this issue would fix the problems in the IGC connector.  I have also been fixing the other modules.

mandy-chessell:
Trying to think what would be the least effort to unpick this.  Maybe merge this fix and I will sort out the clashes in my branch.

cmgrote:
Darn, figured I'd take the hit / pain of fixing everything in one go -- but was afraid this might be the case...

mandy-chessell:
Better to have this attention to detail rather than a community that does not care :)

PR #755, opened 2019-02-16
Removes unsupported overridden methods from IGCOMRSMetadataCollection

cmgrote:
I had overridden all methods in OMRSMetadataCollectionBase in anticipation of working through several of them, but given recent change to allow unsupported methods for some of the write-specific ones I'm removing all of those that we do not intend to implement in the IGC repository connector.

(This way they automatically draw the unsupported status from OMRSMetadataCollectionBase.)


mandy-chessell:
This is a good change - it means as we enhance the parameter checking in the base then the IGC connector benefits from it.  Also easier to see what the IGC connector supports.

Issue #765, opened 2019-02-19
Add RID to qualifiedName for Connection for IGC


cmgrote:
Need to ensure that `qualifiedName` for Connections are unique -- the name alone in IGC is insufficient (can have multiple Connection objects with exactly the same name), so need to add something like RID into the `qualifiedName` to ensure it is unique.

PR #768, opened 2019-02-19
Adds RID to connection qualifiedName for IGC and implements new PurgeDelete events

cmgrote:
Qualified names for other assets should remain unchanged (no RID necessary to make them unique), but for `Connection` entities they now look like:

`(data_connection)=LOCALFS_b1c497ce.8e4c0a48.och6f7sjm.3idrs1v.75e0po.vc663oulk78vej4u5fj4p`


cmgrote:
Fixes #765

Issue #576, opened 2019-01-18
Update atlas docker image


planetf1:
Additional Atlas fixes have been made in PR (#575) for issue #562 by graham wallis

As we don't yet have our docker builds automated I need to rebuild manually.

planetf1:
done

Issue #345, opened 2018-11-02
Add openLDAP to helm charts


planetf1:
Add an example open ldap configuration to helm charts

Most real environments using Egeria are likely to make use of ldap for user authentication (and potentially for user role/group membership)

The existing charts should be extended to add ldap service, and to knit use of this service into the other deployed services (ie ranger etc)

planetf1:
LDAP setup is now performed by an additional container 

Note that the .ldif files previously checked in needed to be refactored.
Additionally there were some errors with the group definitions which have been corrected
Additional to limit the additional scope of changes the default openldap org of example.org has been left as is - this can be changed in a future PR

planetf1:
After builds updated, this has been verified to work ok. The VDC ldap server now has coco pharma personas

PR #778, opened 2019-02-21
Helm20 : Enable docker build by maven

planetf1:
See #576 for details on this change

Docker images can now be built using something similar to
mvn clean install -Ddocker -Ddocker.repo=planetf1

Note
 * docker needs to be installed locally
 * Significant disk space is needed 
 * The build may take an hour or more as atlas, egeria and other software all gets recompiled (again) from source with the current Dockerfiles (optimization is the subject of other issues)

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/docker/README.md
+++ b/open-metadata-resources/open-metadata-deployment/docker/README.md
@@ -26,6 +26,6 @@ In this example:
  * All docker builds will currently set the tag (version) to be the same as the maven version ie 1.1-SNAPSHOT. This means that when testing it is imperitive to always force pull fresh images, or an old version may be used. For example when using kubernetes ensure ImagePullPolicy = 'Always'
 . Previously every single change was versioned, but this led to significant overhead in storage as well as constant changes in the source code just to do a rebuild.
 
-* If DOCKER_HOST is set to a remote host connecton using ssh, the docker build may fail. Refer to https://github.com/spotify/dockerfile-maven/issues/272 for further information
+* If DOCKER_HOST is set to a remote host connecton using ssh, the docker build may fail. Refer to https://github.com/spotify/dockerfile-maven/issues/272 for further information 
 
 * If building on MacOS, the OSX credential store is not supported for docker logins (which is needed to push an image to the repo). To workaround this remove the line '"credsStore": "osxkeychain",' from ~/.docker/config.json and reissue 'docker login'. See https://github.com/spotify/dockerfile-maven/issues/273 for further info


planetf1:
This has also pulled in changes for ldap in #345 - which isn't tested yet, but doesnt cause breakage, and required the build to be addressed prior to continuing.

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -73,8 +73,8 @@ igc:
 image:
   ranger:
     repository: planetf1/ranger-admin-egeriavdc
```
cmgrote: These are all still hard-coded to your personal Docker repo...  Is there not some way in Helm to have a single parameter for repository, and then use `{{ repo }}` (or similar) in each of these to repeat it (so there's only one value to overload)?

cmgrote:
The other area I think we can optimise is this double-container setup everywhere for the "init" pieces...  Have you seen this (temporal) `initContainer` functionality that can be put into the k8s manifests (helm templates)?

https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

It would give us the basic "orchestration" to wait on bringing up a particular pod's real container until other pods / services it depends on are available and running, and also means we wouldn't need to have utilities in the runtime containers themselves to do things like check other services are running (eg. `curl`, `http`, etc). (Since these initContainers block bringing up that pod's real container until the dependencies are confirmed, though, they couldn't be used for post-up configuration that some of the existing "init-xyz" are doing. But I suspect there are alternative practices for that, where needed... having a separate long-running container in the same pod that's quickly finished doing any meaningful work seems clunky?)

planetf1:
Yes, I've come across init containers. 

I wasn't able to solely go with init containers initially, since I needed to 'configure'  the associated container. 

You are correct though that where we are waiting for an external resource, let's consider kafka for example, then an init container could be used, though in theory at least for egeria, the implementation does in theory allow for the server coming up even before kafka is present, and messages will be queued...

Another use for an init container could be to build configuration for the real container -- if something more complex than a simple file is needed. 

In terms of the current 'sidecar' approach -  we're likely to enhance egeria to allow a server to be auto-started based off configuration, and simplify how it's role is defined, and it may be simple enough to allow a basic file/configmap to be used. In the egeria case at least (and hence also for igc event mapping/connector, UI, security sync etc). So that could remove the need for this sidecar entirely

Another approach where there are dependencies between services is to use statefulsets - since then very clear assertions can be made about sequencing, dependencies, and parallelism. The downside (and upside) is a more tightly bound service, which seemed to add complexity for an experimental 'demo' environment. That being said it is a technique to consider on a case by case basis -- if we do have any tightly clusterered pieces.

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
planetf1: yes undersood -- actually the default is encoded for sure, but all these values can be overriden in helm already .

This issue was concerned with the maven build process only... and minimal changes to allow things to still work (hence why the tags got updated). refactoring of the helm charts will be in another change. Baby steps. small increments. There are so many moving parts. My plan is to simplify the override by allowing a user to specify the repository base name just once rather than multiple times, and probably similar for the tags

I'd propose to leave a default in place -- the images indicated are on docker hub so available to all, which means they would work for a chart consumer with no changes. At some point I hope we can get images under the 'odpi' account, at which point the default will change away from my account. But we don't have that capability yet,.

planetf1:
recheck

PR #795, opened 2019-02-28
Helm21 : Support Java Debug

planetf1:
See #576 for docs

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/README.md
+++ b/open-metadata-resources/open-metadata-deployment/README.md
@@ -12,7 +12,7 @@ Currently the focus is on Helm charts to facilitate deployment to Kubernetes clu
 
 Docker images are created independently from this project and hosted on docker hub.
 
-Helm charts will be reused from other helm repos when available
+Helm charts will be reused from other helm repos when available.
 
 ## Kubernetes install
 


cmgrote:
Just raises one question for me: do we want to have a single Egeria image in which we bundle all of the components (eg. OMAG Server Platform, UI, etc) and choose at runtime which one to startup? Or a different image per component?

The latter gives us smaller images, of course, but more images overall... But sticking with container ethos regardless of the number of images they should be started as separate containers, really (ie. one process per container) (?)

I'm sort of leaning towards an image per component, as I think it'll be easier to maintain and manage at runtime than a complicated entrypoint script / override (?) (I have a multi-stage build in the works already that will build Egeria only once as a stage, and then use that build image as the sourcing point for multiple runtime images -- so if it's the build process that's a concern that should be (nearly) overcome...)

planetf1:
I would agree one container per executing runtime.
Where the contents of the container need to be the same - or nearly, then we should use the same image.
An example would be egeria (main omag server) , IGC connector. Currently the design is that both run within the server chassis - so it's basically the same code, just configured (via REST or the config file) seperately.
The UI code does launch a seperate jar currently, but it's about to be added to server chassis, which is why I am also using the same image. It's also why the standard technique of specifying CMD (which can be overriden) is used in addition to entrypoint. But in the UIs case that override will likely go away -- but it's hard to know when or how exactly as it itself is a moving target which we need to flexibly adapt to.

I have removed the additionally image variables as they were superfluous since we are using the same image.

In this PR case the entrypoint was necessitated by wanting to evaluate environment variables AT RUN TIME.. This can't be done directly in CMD or ENTRYPOINT and needs to be done via a script - a common pattern. See the postgres container for example. I could have introduced the entire environment variable, but felt it was clearer and safer to just allow debug on/off

it's true we *could* extend entrypoint to also add other script snippets, or make extensions configurable - some containers for example use 'run-parts' to run extra bits, but I've not jumped to that yet

planetf1:
Let me know if you have other concerns. I'll plan to promote otherwise in a few hours to work through some of the debugging/deployment issues around ui,asset catalog.

planetf1:
For some reason the 'DCO check' (which just checks signatures and takes <1s) was 'stuck' in pending status. I just pushed a readme dummy update to force a recheck so that I can merge...

PR #803, opened 2019-03-01
Demo deployment optimisations and cleanups

### LINKED ISSUE #626: Move IGC user/password to k8s secret
Currently the igc user/pasword is stored in values.yaml of the vdc chart and must be manually edited before deployment as needed

User/passwords should be managed using kubernetes secrets.

### LINKED ISSUE #627: Refactor k8s initialization scripts (igc, atlas, egeria etc)
Currently these components use a simple shell script to wait for things to be ready and then configure their co-located component.

There are common capabilities across all these scripts
 - check a REST endpoint is alive
 - Send a REST call & check results
 - check kafka is alive
 - retry loop

These scripts could a) be refactored internally b) use shared components and possibly c) be rewritten in python or another more appropriate language if not needed.

Not an immediate issue to resolve

### LINKED ISSUE #670: Consider switching to adoptopenjdk/alpine base images for java apps in charts

### LINKED ISSUE #780: Split postgresql data loading from IGC configuration in VDC helm chart
Currently the postgresql-init container runs ansible scripts that both
 * Load sample data into the VDC environment (postgresql)
 * Load sample data into IGC
 * Invoke IGC's 'IMAM' capability to import sample data metadata into IGC 

It would be desirable to separate the two, so that IGC is not required, yet sample data can still be loaded
Additionally since IGC sits outside k8s currently it may be better to invoke the IGC data loading activities externally - clearer separation

### LINKED ISSUE #782: VDC helm chart - add support for external cloud kafka - such as Event Streams
Currently the VDC chart depends on a kafka chart, which means we deploy an instance of kafka as part of installing VDC.

It would be useful to provide -- as an option -- the ability to use an existing cloud based kafka service.
For example on IBMcloud this could be IBM Event Streams.
A similar approach could be taken for google, azure etc.

A simple optional setting in 'values' could either enable internal or external kafka
Additional settings could provide the fixed endpoint and authorization data

cmgrote:
Main changes (aside from documentation cleanups):

- Removes (now unused) older duplicate data files under `SampleData/CocoPharma/dbdata`
- Enables switching between cluster-internal (Helm-deployed) and external (pre-existing) Kafka configs to fix #782 
- Consolidates the `Deployment` and `Service` documents into a single file per module (was constantly forgetting to update a port number in one or the other otherwise)
- Removes the redundant `-deployment` at the end of every pod name (just takes up logging screen real estate)
- Removes the app name prefix from the deployed container names (they are namespaced within the pod already, so no need to have a variable name inside the pod)
- Cleans up `configmap` to remove unused variables, add conditionality for internal / external services, and remove the fully-qualified internal DNS names (short names work fine, and then should not cause problems if people deploy into other namespaces) to fix #626 
- Uses a standard `initContainer` in most places to check availability of dependencies before running the pod's actual container to fix #627 
- Adds `readinessProbes` and `livenessProbes` where possible
- Adds `Job`s for one-time tasks like data loading, sending config to Egeria, etc (will auto-restart on failure until successfully completed or retry limit exceeded; Helm will auto-clean once successful) to fix #780 and fix #627 
- Replaces custom postgresql with standard (and minimal) postgresql container to fix #780 
- Uses multi-stage builds to produce significantly smaller (10x) runtime images to fix #670 
- Changes default build registry and repository for docker images to be one everyone can build and push to (and updating instructions to indicate how to do so)
- Removes redundant `README.md` in every docker image sub-folder -- if we want to document how to build images outside Maven we should probably just do this once, and not try to maintain the same instructions in so many places...
- Adds ability to use either a cluster-external IGC or a container-based one; as commercial software also does not setup any IGC by default
- Now embeds the Egeria IGC Proxy as a container in the same pod as IGC itself (if deploying as a container) to get around any hostname resolution issues (same pod = same hostname); if external IGC will automatically add `hostAlias` entry to allow IGC Proxy to properly resolve IGC environment's hostname irrespective of DNS

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -187,11 +187,6 @@
                 <artifactId>spring-boot-starter-data-jpa</artifactId>
                 <version>${spring-boot.version}</version>
             </dependency>
-            <dependency>
-                <groupId>org.springframework.boot</groupId>
-                <artifactId>spring-boot-maven-plugin</artifactId>
-                <version>${spring-boot.version}</version>
-            </dependency>
         </dependencies>
     </dependencyManagement>
 
@@ -284,7 +279,12 @@
                    <groupId>org.jacoco</groupId>
                    <artifactId>jacoco-maven-plugin</artifactId>
                    <version>0.8.3</version>
-                </plugin>
+	   </plugin>
+	        <plugin>
+                    <groupId>org.springframework.boot</groupId>
+                    <artifactId>spring-boot-maven-plugin</artifactId>
+	    	    <version>${spring-boot.version}</version>
+	        </plugin>
         </plugins>
 
     </pluginManagement>


cmgrote:
Somewhat interesting conundrum for testing: the new `Job`s that initialise things set themselves up by pulling just what they need from `master` at startup; but until the PR is merged, the content they're after doesn't exist in `master` (so the first job, which does the data load to postgresql, will just go into a failure loop)...

Not sure what the ideal way to handle such a scenario would be?  Doesn't seem great to pre-build all the content into the docker image at build-time, either (requires re-building the image itself any time the content used for configuration changes, which is very heavy...)

cmgrote:
Latest docker image sizes:

name | size | notes
---|---|---
localhost:5000/odpi/egeria | 199MB | chassis + UI
localhost:5000/odpi/ranger-admin-egeriavdc | 4.34GB | (still to be optimised)
localhost:5000/odpi/omrsmonitor-egeriavdc | 83.6MB | (probably already optimised?)
localhost:5000/odpi/ldap-init-egeriavdc | 383MB | (probably should be replaced by `initContainer`?)
localhost:5000/odpi/kafka-monitor-egeriavdc | 630MB | (still to be optimised)
localhost:5000/odpi/gaian-egeriavdc | 628MB | (still to be optimised)
localhost:5000/odpi/egeria-configure | 64.7MB | db clients + git + any other utilities
localhost:5000/odpi/apache-atlas | 1.07GB | Solr + HBase + Atlas
... total ... | 7.5GB | 

I'd guess we can probably get the footprint to less than 5GB overall assuming similar optimisations for the un-optimised containers (eg. Atlas went from 10GB to 1GB)...

planetf1:
Thanks. I will try and read through this in the next day or two, I think it's important we go through that as it's a lot of changes in a PR, and then we can look at also testing on IBM cloud; looks like a very useful set of changes.

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
```diff
@@ -0,0 +1,171 @@
...
+              port: 9446
+              scheme: "HTTPS"
+            initialDelaySeconds: 180
+            periodSeconds: 20
+            timeoutSeconds: 3
+            failureThreshold: 3
+          resources: {}
+          securityContext:
+            capabilities:
+              add: ["IPC_OWNER"]
```
planetf1: A little uncomfortable, but given DB2 needs it may be difficult to address. Where this isn't permitted external IGC can be used anyway. I will experiment (after merge) to see how this can be done with ibm cloud (we should add notes to the readme with anything we find)

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ingress.yaml
```diff
@@ -1,72 +1,69 @@
...
 
-        #ingress.bluemix.net/rewrite-path: |
-        #  serviceName={{ include "myapp.fullname" . }}-ranger-service rewrite=/;
-        #  serviceName={{ include "myapp.fullname" . }}-atlas-service rewrite=/;
+    #ingress.bluemix.net/rewrite-path: |
+    #  serviceName={{ include "myapp.fullname" . }}-ranger-service rewrite=/;
+    #  serviceName={{ include "myapp.fullname" . }}-atlas-service rewrite=/;
 
-    name: {{ include "myapp.fullname" . }}-ingress
+  name: {{ include "myapp.fullname" . }}-ingress
```
planetf1: I wasn't able to use ingress satisfactorily on IBM cloud so intend to return to it.. so can't comment on whether this works or how it works across platforms

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/job-01-data-loader.yaml
```diff
@@ -0,0 +1,83 @@
+# SPDX-License-Identifier: Apache-2.0
```
planetf1: The official PG image has the option to run db scripts. there was a bug I reported, now supposedly fixed. BUT I think we have better flexibility doing it this way (and the job is a good change)

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/job-01-data-loader.yaml
```diff
@@ -0,0 +1,83 @@
...
+        - name: wait-for-db
+          image: busybox
+          command:
+            - "sh"
+            - "-c"
+            - >
+                until nslookup {{ include "myapp.fullname" . }}-postgresql-service; do
+                echo waiting for {{ include "myapp.fullname" . }}-postgresql-service;
+                sleep 2;
+                done;
```
planetf1: Will this be sufficient? JUst the name being available? Perhaps using 'netcat' to check the port can be connected to is one step better? (obviously valid postgresql request even better). I found netcat generally good enough in the initial scripts

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/job-02-init-egeria.yaml
```diff
@@ -0,0 +1,64 @@
...
+          envFrom:
+            - configMapRef:
+                name: {{ include "myapp.fullname" . }}-configmap
+          command:
+            - "/bin/bash"
+            - "-c"
+            - >
+                curl -f --verbose --basic admin:admin -X POST
+                --header "Content-Type: application/json"
+                ${EGERIA_ENDPOINT}/open-metadata/admin-services/users/${EGERIA_USER}/servers/${EGERIA_SERVER}/event-bus
```
planetf1: To support external kafka we likely need ot add support for authentication. I suggest we open a separate issue on this

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -46,42 +47,21 @@ service:
...
-
-# Images ending -init are typically containers that perform configuration,
-# often in the same container as the component they are configuring
-
-# Note that sometimes the same docker image is used for multiple purposes - for example
-# Both the Egeria UI, and Egeria core use the same docker image -- for now
-
 image:
   ranger:
-    repository: planetf1/ranger-admin-egeriavdc
```
planetf1: Neither of these are perfect. at least planetf1 is externally resolveable, though it *should* be ODPi. but it can be yet Suggestions? I think if nothing better we should revert to the original as at least anyone can then take the chart. I will chase up build team to get an ODPi account on docker hub.

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -91,74 +71,95 @@ image:
...
-    repository: planetf1/omrsmonitor-egeriavdc
+    repository: localhost:5000/odpi/omrsmonitor-egeriavdc
     tag: 1.1-SNAPSHOT
     pullPolicy: Always
 
   ldapinit:
-    repository: planetf1/ldap-init-egeriavdc
+    repository: localhost:5000/odpi/ldap-init-egeriavdc
     tag: 1.1-SNAPSHOT
     pullPolicy: Always
```
planetf1: We should probably revert the pullPolicy (which I changed!) to if not present, also tag probably to latest. But this has implications on the current maven-oriented build due to the way tagging is done. Separate issue?

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -91,74 +71,95 @@ image:
...
+# see 'cp-kafka' below for config options
 kafka:
+  internal:
+    enabled: true
   external:
-    enabled: false
-    type: NodePort
+    hostname: ""
+    ip: ""
+    port: ""
```
planetf1: will need credentials I think (mentioned above). i can figure this out in a future PR as needed

open-metadata-resources/open-metadata-deployment/docker/atlas/pom.xml
```diff
@@ -41,7 +41,7 @@
                     </execution>
                 </executions>
                 <configuration>
-                    <repository>${docker.repo}/atlas-1773-egeria</repository>
+                    <repository>${docker.registry}/${docker.repo}/apache-atlas</repository>
```
planetf1: Ah! missed that

open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
```diff
@@ -1,51 +1,24 @@
...
-WORKDIR /root/egeria
+WORKDIR /root/egeria/
 # Add Pull request if specified (requires git-extras package)
 RUN if [ ! -z "$pr" ] ; then git pr $pr ; fi
-
 RUN mvn clean install -DskipTests
 
+FROM openjdk:8-jre-alpine
+COPY --from=build /root/egeria/open-metadata-distribution/open-metadata-assemblies/target/egeria-*-omag-server/servers/omag-server-*/server-chassis-spring-*.jar /server-chassis-spring.jar
+COPY --from=build /root/egeria/open-metadata-distribution/open-metadata-assemblies/target/egeria-*-distribution/egeria-omag-*/user-interface/access-services-user-interface-*.jar /access-services-user-interface.jar
```
planetf1: Ah got it (ref q above). The UI should move into the server chassis but will have to deal with that when it happens

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/apache-atlas.yaml
```diff
@@ -0,0 +1,83 @@
...
+        - name: apache-atlas
+          image: "{{ .Values.image.atlas.repository }}:{{ .Values.image.atlas.tag }}"
+          imagePullPolicy: {{ .Values.image.atlas.pullPolicy }}
+          ports:
+            - containerPort: 21000
+          livenessProbe:
+            httpGet:
+              path: /login.jsp
+              port: 21000
+            initialDelaySeconds: 120
```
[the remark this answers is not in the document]
cmgrote: Yes, I found the timeouts inconsistent, but also that it was better to let the pod restart itself due to failing the probes rather than waiting for very long times (even after which certain containers, like Atlas, didn't start due to what appeared to be some internal race issue).

From reading about the probes, this sounded like precisely what they're there for =)

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/configmap.yaml
```diff
@@ -12,37 +12,36 @@ metadata:
...
+  # Below values for IBMIGC need to reference either your container or external environment
+  # See also: ibmigc... in values.yaml
+{{ if .Values.ibmigc.internal.enabled }}
+  IBMIGC_ENDPOINT: https://infosvr:9446
+  IBMIGC_KAFKA_ENDPOINT: infosvr:59092
+{{ else }}
+  IBMIGC_ENDPOINT: https://{{ .Values.ibmigc.external.hostname }}:{{ .Values.ibmigc.external.port }}
+  IBMIGC_KAFKA_ENDPOINT: {{ .Values.ibmigc.external.hostname }}:59092
+{{ end }}
+  IBMIGC_AUTH: aXNhZG1pbjppc2FkbWlu
```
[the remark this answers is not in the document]
cmgrote: This is the default for most setups of IGC (`isadmin` / `isadmin`), including indeed the Ansible deployment. The README already documents that these configmap values need to be confirmed for IGC, as well, when using IGC as part of the setup.

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
```diff
@@ -0,0 +1,171 @@
...
+            - >
+                cd /samples && git init egeria &&
+                cd egeria &&
+                git remote add origin https://github.com/odpi/egeria.git &&
+                git config core.sparsecheckout true &&
+                echo 'open-metadata-resources/open-metadata-deployment/sample-data/*' >> .git/info/sparse-checkout &&
+                git pull --depth=1 origin master
+          volumeMounts:
+            - mountPath: /samples
+              name: samples-volume
```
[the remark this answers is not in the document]
cmgrote: Raised #810

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
```diff
@@ -0,0 +1,171 @@
...
+      imagePullSecrets:
+        - name: regcred
+      initContainers:
+        - name: init-git
+          image: "{{ .Values.image.configure.repository }}:{{ .Values.image.configure.tag }}"
+          command:
+            - "/bin/bash"
+            - "-c"
+            - >
+                cd /samples && git init egeria &&
```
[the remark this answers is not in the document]
cmgrote: Raised #810

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
cmgrote: Precisely: for external IGC this is all skipped, for container-based we don't really have a choice (DB2 won't work, and we require DB2 to run IGC...)

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ingress.yaml
cmgrote: I've only changed indentation to 2-space to match all the other YAMLs...

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/job-02-init-egeria.yaml
cmgrote: Raised #811

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
cmgrote: The problem is that with that approach no one else can *build* (and push / re-use) the charts (ie. updated ones)...

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
cmgrote: Raised #812

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/job-01-data-loader.yaml
cmgrote: Good catch -- I'd replaced most of this with a more robust check (basically introspecting k8s itself to see that there is at least one pod available and ready behind the service) before continuing -- just pushed a change for this one as well.

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
planetf1: I know :-) I was focussed on the consumer in the first case ie anyone can install our charts.
I've opened up https://github.com/odpi/ci-management/issues/44 to address this with the ODPi/LF team so let's chase that.

Issue #815, opened 2019-03-05
Maven docker build fails after refactoring


planetf1:
```
20:20 $ mvn clean install -Ddocker -Ddocker.repo=planetf1
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] Resolving expression: '${docker.registry}': Detected the following recursive expression cycle in 'docker.registry':
[docker.registry] @
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project org.odpi.egeria:open-metadata-docker:1.1-SNAPSHOT (/home/jonesn/src/egeria/open-metadata-resources/ope
n-metadata-deployment/docker/pom.xml) has 1 error
[ERROR]     Resolving expression: '${docker.registry}': Detected the following recursive expression cycle in 'docker.registr
y': [docker.registry] -> [Help 2]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/ProjectBuildingException
[ERROR] [Help 2] http://cwiki.apache.org/confluence/display/MAVEN/InterpolationCycleException
```

planetf1:
Note this is an optional step - does not affect main build, only build of refactored docker images

cmgrote:
I made registry a required option (see README) — if there’s a way to default it to Docker public but otherwise still include it (where it was previously missing) happy to do that?

planetf1:
Tested with
mvn clean install -Ddocker -Ddocker.repo=planetf1 -Ddocker.registry=localhost:5000 -> pushes to docker registry
(having run docker run -d -p 5000:5000 --restart=always --name registry registry:2)
mvn clean install -Ddocker -Ddocker.repo=planetf1 -> pushes to docker hub as default
mvn clean install -Ddocker -> fails with error that repo isn't specified

PR #816, opened 2019-03-06
#815 fix docker build

planetf1:
See #815 


planetf1:
thanks... will merge. however the build infrastructure is stalled again, so have chased support team and will merge once able to. thanks for testing.

Issue #739, opened 2019-02-08
Docker repo for VDC images is non-ODPi


cmgrote:
The docker repository for the VDC images is currently hard-coded to a non-ODPi location (Nigel's): `planetf1/egeria-egeriavdc`, etc.

(See `image` map inside `open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml`)

Needs to be changed to ODPi, so that images can be updated as-needed by the broader team.

planetf1:
Indeed. This is part of an overall requirement to look at the build pipeline for the docker images.
This needs to be automated, quite likely via the linux foundation jenkins server
I'll likely also add an additional maven profile to allow an easy user-initiated build, and allow the repo name to be overriden
(one of the shortcuts getting a demo env ready quickly)

cmgrote:
Wondering if for development work it would be handy to be able to use a local registry (`localhost:5000`) for re-deployment of changes to some containers to integration test, etc -- presumably part of what #771 will enable (you mention parameterising the repo, I guess / hope this is for the chart as a whole)?

planetf1:
I think configuration of the image repository is important -- I initially chose dockerhub as a quick way to get started. 

In a corporate k8s environment I might well want to push images to my local container registry - perhaps in openshift, or ICP. On IBM cloud I might want to push to the container repo there.

I hadn't thought too much about the local case (I try and avoid this - my machine is too full!) - but no reason why not

The image repository is in fact already configurable on a per-image basis, but it might help to have a single setting for the default registry to use also - meaning if you want all images from one place, just set it as a single parameter for the helm command (or config file). 

And for container build in maven, the intent was indeed to parameterize the repo to write to.

planetf1:
See #771 for further details on the changes made to docker build.

I will keep this issue to track changing the default repository from planetf1 to odpi (or similar -- or even updating the default registry) once we have the required build process & shared ids in place

planetf1:
See https://github.com/odpi/ci-management/issues/44 for a request to the build team to get access to ODPi.

Suggest we use this issue also to parameterize the repo to be used for images.

it's already possible to update the image to be used, but would be useful if by default we could just set a single value for all images (still allowing for individual images to override this)

planetf1:
I'll assign, and update once process in place

planetf1:
See the ci-management issue above for latest updates.
Basically the build is now working but we are trying to get access to the odpi account on dockerhub.
[...]

PR #824, opened 2019-03-07
Parameterises container registry and pull policy with defaults

cmgrote:
To address #739

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/charts/vdc/templates/rbac.yaml
+++ b/open-metadata-resources/open-metadata-deployment/charts/vdc/templates/rbac.yaml
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Egeria project.
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    app.kubernetes.io/name: {{ include "myapp.name" . }}
+    helm.sh/chart: {{ include "myapp.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/component: vdc-rbac
+  name: {{ include "myapp.fullname" . }}-vdc-api-role
+rules:
+- apiGroups: [""]
+  resources: ["endpoints","pods","services"]
+  verbs: ["get", "list", "watch"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  labels:
+    app.kubernetes.io/name: {{ include "myapp.name" . }}
+    helm.sh/chart: {{ include "myapp.chart" . }}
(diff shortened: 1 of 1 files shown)


open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -55,71 +55,61 @@ debug:
 logging:
   level: NONE
 
-
-# The following section defines all of the DOCKER images being used by this
-# chart. Normally they should be left as is, but are exposed here in case the
-# user wishes to extend.
+imageDefaults:
+  registry: "localhost:5000"
```
planetf1: I am considering reverting the default registry to dockerhub (and making the repository configurable).
I am still awaiting the LF build team, but in the meantime I've had a couple of people ask about using the charts.
The easiest way of doing that is to ensure the images work 'out of the box' MUCH though I dislike the non-official repo id which I will again try and push forward. For them to build, and setup a repo, is not as easy - though it is important to have the capability

I can do this under an additional PR if easier - we don't need to change this one /miss out on the other changes

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -55,71 +55,61 @@ debug:
...
+    registry: "docker.io/library"
     repository: mariadb
-    tag: latest
-    pullPolicy: Always
 
   egeria:
-    repository: localhost:5000/odpi/egeria
-    tag: 1.1-SNAPSHOT
-    pullPolicy: Always
+    repository: odpi/egeria
```
planetf1: the repo name needs to be configurable (see above comment) - clearly it already is, but we can default it too, for the egeria custom images (ie not mariadb)

open-metadata-resources/open-metadata-deployment/docker/configure/pom.xml
```diff
@@ -32,17 +32,45 @@
                 <artifactId>dockerfile-maven-plugin</artifactId>
```
planetf1: I think this should work - the two step process is needed to get the latest approach working.
I can't test in the next few hours as the full build is too long, so will raise a new issue if any issues observed

planetf1:
See review above. Despite review comment merging now is probably the best approach, and a new PR can cater to the other changes mentioned?

open-metadata-resources/open-metadata-deployment/docker/configure/pom.xml
cmgrote: I tested on my side and it worked, but would indeed be good to have a peer-reviewed test =)

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
cmgrote: Raised #829

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
cmgrote: Included in #829

Issue #831, opened 2019-03-08
Updates to helm charts to better support local & cloud


planetf1:

planetf1:
A couple of minor changes
 - <! xml marker in yaml comment
 - external IGC configurable port
 - IGC auth base64 string moved out of configmap into values
 - default to no debug (egeria policy -- I override it locally)
 - a few lint issues noticed whilst debugging (Note: there are more.....)
 - default IGC to disabled (most users won't have access to the image)
 - comment on how to rebuild docker images

planetf1:
* Introduced new _kafkaauth.tpl. This is prefixed _ so that helm does not interpret as a k8s resource. Instead it provides a few common templates for external kafka. At this time these are supporting IBM IKS, but a conditional could be added here to support other environments
* Changed external kafka values to improve structuring of external config. I did consider adding individual parms, or a custom map to allow for any extra kafka parms, but this makes configuration for known services more complex. Skipped for now, but in future we may want provider: custom & allow for any custom values
* modified jobs/deployments to include the k8secret (from template above), and make use of the new kafka fragment when configuring eventbus. To reduce risk and improve readability the old parms were left in-situ - new code used for external kafka only
*Fixed up a few template parsing issues (based on parms)
* Removed basic kafka monitor

planetf1:
Note that the secret is assumed to contain two values
 username
 password

to create this you can use
```
echo -n 'token' > username
echo -n '<APIKEY>' > password
kubectl create secret generic auth-vdc-kafka --from-file=./username --from-file=./password
```

For IBM Event Streams you can use user/password (cloud) or token/key (cloud and on-prem) - other providers are expected to work similarly.

PR #832, opened 2019-03-08
#831 updates to work better in cloud & local

planetf1:


open-metadata-resources/open-metadata-deployment/charts/vdc/Chart.yaml
```diff
@@ -1,7 +1,8 @@
 # <!-- SPDX-License-Identifier: Apache-2.0 -->
```
cmgrote: Worth dropping the XML comment surround here as well?

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
```diff
@@ -174,7 +174,7 @@ spec:
                 name: {{ include "myapp.fullname" . }}-configmap
           env:
             - name: "logging.level.org.odpi.openmetadata.adapters.repositoryservices.igc"
-              value: "DEBUG"
+              value: {{ .Values.logging.level }}
```
cmgrote: We should probably just drop this environment variable from the default config (or inherit the top-level logging rather than this IGC-specific-level) (?)

planetf1: Currently that fragment does inherit a common logging value from the template -- obviously if the user of that variable within the code changes then we should update accordingly.

The value is defaulted to no logging on the principle that debug should be possible via audit logs.

open-metadata-resources/open-metadata-deployment/charts/vdc/Chart.yaml
planetf1: Agreed - cut/paste error.
Corrected and repushed.

open-metadata-resources/open-metadata-deployment/charts/vdc/templates/ibm-igc.yaml
planetf1: suggest we push add-is, and if a change is needed go with new issue.
I think the proposed change is a step forward.

PR #842, opened 2019-03-13
Recover the old version of Virtualizer

wyf-1116:


planetf1:
I thought we were deleting the old virtualizer.... is it still in use?

GuyAckermans:
Hi Nigel,
Slight misunderstanding on my end: The igc  proxy was supposed to be removed, but the deprecated virtualizer has to remain until the new one is completed.

PR #844, opened 2019-03-13
Ranger Connector implementation Ranger Connector extends Connector Base.

danielaotelea:
It provides an interface that is called from the Security Sync in order to persist the tags and resources to the Ranger Server.
In the Ranger Connector contructor is provided a connection object that contains the Ranger Server URL and the authorization token.
The Security Sync Component received the events and the processing is done in the implementation. Ranger connector is used to build the specific objects for the Ranger Server and to push the changes.


mandy-chessell:
The connector name should be something like "SecurityServiceConnector" and there should be an implementation of the connector called something like "RangerSecurityServiceConnector".

PR #845, opened 2019-03-13
FindEntityByPropertyValue should accept the null and empty string as …

danielaotelea:
…a search criteria

FindEntitiesByPropertyValue should accept null or "" as search param.
The validation from the OMRS should be removed.
Also, the validation from the Atlas connector should be removed.
All the entities with the given type should be returned from the find method.


cmgrote:
I'll need to make some other changes to the IGC handling for the case where `searchCriteria` is null...  Just to confirm: `null` or `""` should be treated the same (ie. retrieve all entities of the type in both cases)?

cmgrote:
I'd also suggest we add a sentence to the JavaDoc comments (or `searchCriteria` parameter itself) indicating that when null / empty this is the expected behaviour, just for clarity down the road =)

mandy-chessell:
Agree with @cmgrote that the JavaDoc commets should be updated - not just in the interface but all implementations - Base, Local, Enterprise, Atlas, IGC, Graph, Inmemory ...

cmgrote:
@danielaotelea -- I think all that was left on this one now is to update the JavaDoc comments to indicate this as expected behaviour for `null` / `""`...  Hopefully then we can get this committed and closed? 😁

bogdan-sava:
is this PR still needed ?

bogdan-sava:
As discussed with @danielaotelea I'll close this PR

PR #849, opened 2019-03-14
Fix the documentation of the VDC demo

wyf-1116:


GuyAckermans:
VDC documentation has been restored

PR #856, opened 2019-03-15
Helm charts - improved support for public cloud

planetf1:


planetf1:
See 831 for more info

planetf1:
See #831 for more info

This PR fixes up the helm charts to work in a public cloud environment, and in particular has enabled 
the use of one cloud based provider (IBM Event Streams). 

additional support may be added in future - for example
 - use of simple local kafka
 - custom to allow ad-hoc K/V pairs to be passed (flexible but harder to configure)
 - support for other major providers such as Azure, Google etc

As these are worked through other generalizations/refactoring may be needed, but this is a start,

PR #859, opened 2019-03-15
Data engine - update process creation

popa-raluca:
* Add API for creating a process based of an array of Ports/DeployedAPIs/Assets
* Add API api for linking an array of ports to an existing process


bogdan-sava:
recheck

bogdan-sava:
recheck

Issue #861, opened 2019-03-15
OutOfMemory errors caused by accumulation of incoming Kafka messages


dkantor:
We have seen OutOfMemory errors with the following stack trace.  Analysis of heap dump showed there was a backlog of 53,079 unprocessed kafka messages in `KafkaOpenMetadataTopicConnector`, taking 412 MB of heap space.
```
java.lang.OutOfMemoryError: Java heap space
    at java.lang.StringBuilder.ensureCapacityImpl(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at java.util.AbstractCollection.toString(Unknown Source)
    at java.util.Collections$SynchronizedCollection.toString(Unknown Source)
    at java.lang.String.valueOf(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at org.odpi.openmetadata.adapters.eventbus.topic.kafka.KafkaOpenMetadataTopicConnector.checkForEvents(KafkaOpenMetadataTopicConnector.java:246)
    at org.odpi.openmetadata.repositoryservices.connectors.openmetadatatopic.OpenMetadataTopicConnector.run(OpenMetadataTopicConnector.java:96)
    at java.lang.Thread.run(Unknown Source)
```
Please see [Jeff Hagelberg's comment about Kafka polling](https://github.com/odpi/egeria/issues/860#issuecomment-473388268).

> The issue is that you can see in `KafkaOpenMetadataEventConsumer` that the code polls Kafka once every second and adds what it finds to a list. `KafkaOpenMetadataEventConsumer.run()` needs to be changed to check the size of the kafka message backlog before polling kafka, and poll only if the number of queued events is small. That should be done as part of this issue. It is really the root cause of the issue. Once that is done, fixing the logging becomes less important (although it does need to be done as well)

jnhagelberg:
The easiest way to fix this would be to add a method to `KafkaOpenMetadataTopicConnector` to get the current size of `incomingEventsList`, such as `getIncomingEventListSize()`.  Then, you would update the code in `KafkaOpenMetadataEventConsumer.run()` to make the Kafka polling conditional.  Something like:
```

        while (isRunning())
        {
            try
            {
                if (connector.getIncomingEventListSize() > 100 ) {
                    continue;
                }
                ConsumerRecords<String, String> records = consumer.poll(defaultPollTimeout);
                log.debug("Found records: " + records.count());
[...]

Issue #860, opened 2019-03-15
OutOfMemory errors caused by debug logging of incoming Kafka messages


dkantor:
We have seen OutOfMemoryErrors with the following stack trace.  Analysis of heap dump showed a `StringBuilder` instance holding 275MB of memory that is a local variable in `java.util.AbstractCollection.toString()`.
```
java.lang.OutOfMemoryError: Java heap space
    at java.lang.StringBuilder.ensureCapacityImpl(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at java.util.AbstractCollection.toString(Unknown Source)
    at java.util.Collections$SynchronizedCollection.toString(Unknown Source)
    at java.lang.String.valueOf(Unknown Source)
    at java.lang.StringBuilder.append(Unknown Source)
    at org.odpi.openmetadata.adapters.eventbus.topic.kafka.KafkaOpenMetadataTopicConnector.checkForEvents(KafkaOpenMetadataTopicConnector.java:246)
    at org.odpi.openmetadata.repositoryservices.connectors.openmetadatatopic.OpenMetadataTopicConnector.run(OpenMetadataTopicConnector.java:96)
    at java.lang.Thread.run(Unknown Source)
```

`KafkaOpenMetadataTopicConnector` at line 246 is logging the contents of `incomingEventsList`:
```
 log.debug("checking for events {0}" + incomingEventsList);
```
At a minimum, the debug logging should be changed to pass incomingEventsList as an argument rather than concatenating it, to avoid the overhead of `incomingEventsList.toString()` if debug is not enabled on the logger.
```
log.debug("checking for events {0}", incomingEventsList);
```
In addition, it would make sense to implement some logging code for efficiently logging the contents of `incomingEventsList` which avoids concatenating the entire contents of the list in a `StringBuilder`.  Perhaps it could log a subset of the events if the list exceeds a certain size.

guptaneeru:
Agree with @dkantor to atleast check the log level. but I would even say to not log this at the debug level. 
On the side note, the real question is why is this collection holding so many messages in memory. Is it something to do with how Kafka was setup?

jnhagelberg:
[...]

PR #862, opened 2019-03-18
#861 - Limit maximum size of kafka event queue

jnhagelberg:
See #860, #861

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
+++ b/open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
@@ -128,11 +128,12 @@ public class KafkaOpenMetadataEventConsumer implements Runnable
             
             	
                 	
-            	if (! pollRequired && connector.getNumberOfUnprocessedEvents() > defaultMaxQueueSize) {
+            	int nUnprocessedEvents = connector.getNumberOfUnprocessedEvents();
+            	if (! pollRequired && nUnprocessedEvents > defaultMaxQueueSize) {
             		//The connector queue is too big.  Wait until the size goes down until
             		//polling again.  If we let the events just accumulate, we will
             		//eventually run out of memory if the consumer cannot keep up.
-            		log.warn("Skipping Kafka polling since unprocessed message queue size {} is greater than {}", connector.getNumberOfUnprocessedEvents(), defaultMaxQueueSize);
+            		log.warn("Skipping Kafka polling since unprocessed message queue size {} is greater than {}", nUnprocessedEvents, defaultMaxQueueSize);
             		awaitNextPollingTime();
             		continue;
             	


open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
```diff
@@ -29,6 +29,7 @@
 
     private static final long recoverySleepTimeSec = 10L;
     private static final long defaultPollTimeout   = 1000;
+    private static final long defaultMaxQueueSize = 100;
```
guptaneeru: Should this size come from a configuration?

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
```diff
@@ -102,6 +103,21 @@ public void run()
...
+            	//from being declared "dead" if the polling does not occur often enough?  There is a Kafka
+            	//setting "max.poll.interval.ms" that may have some bearing on this.  The default value
+            	//of this setting is Integer.MAX_VALUE, so unless we are overriding that property
+            	//we *should* be ok.            	
+            	if (connector.getNumberOfUnprocessedEvents() > defaultMaxQueueSize) {
+            		//The connector queue is too big.  Wait until the size goes down until
+            		//polling again.  If we let the events just accumulate, we will
+            		//eventually run out of memory if the consumer cannot keep up.
+            		log.warn("Skipping Kafka polling since unprocessed message queue size {} is greater than {}", connector.getNumberOfUnprocessedEvents(), defaultMaxQueueSize);
+            		awaitNextPollingTime();
```
guptaneeru: There is kafka property 'max.poll.records'. This identifies the number of records which will be processed on each poll. By default in CAMS it was set to  500. We are reducing it to 100. So that will reduce the number of messages to come in. Essentially each poll will only give us 100 messages. Do we still want to block the thread? 
I think it is  un necessary.

As you said, if we are putting delays in code, then we need to be aware of session timing out as well.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
```diff
@@ -171,14 +187,7 @@ public void run()
             }
             finally
             {
-                try
-                {
-                    Thread.sleep(1000);
```
guptaneeru: Why are we delaying here again? This will slow down the polling thread.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
```diff
@@ -347,4 +347,14 @@ public String getPrintableProperties(Properties   liveProperties)
 
         return printableProperties;
     }
+    
+    /**
+     * Gets the number of events that have not been processed yet.
+     * 
+     * @return
+     */
+    public int getNumberOfUnprocessedEvents() {
```
guptaneeru: I am not very familiar with the egeria code. But I glanced at it. It looks like, events from this list is copied to another list and all events from this list is removed. That does not necessarily mean that events were processed. So there are still large number of un porcessed events while we poll again.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
jnhagelberg: We are delaying here so that we only run this check or do a poll every 1 second.  I know it is odd to have the sleep in 2 different places, but only one of them actually is used.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
jnhagelberg: @guptaneeru Yes, this is needed.  We cannot allow the consumer to run out of memory.  The egeria code has no knowledge of what the kafka parameters are, and it needs to work in all kafka configurations.  In addition, tweaking kafka parameters does not eliminate the possibility of the event queue consuming all of the available heap memory.  This change does do that.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
jnhagelberg: Ideally, yes.  So should the defaultPollTimeout, etc.  It seems like that can be done in a follow on PR.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
jnhagelberg: @guptaneeru Good point.  For the fix to be effective, the property that limits the number events returned by polling also needs to be used.  It was this list that was seen growing unchecked in the heap dump, to the fix I've added is targetted at limiting its size.  I did not see the copy of the list being processed in the profile.

guptaneeru: Right, my point was we may shift the problem from one place to another. But again, some body more familiar with egeria code should evaluate that.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventConsumer.java
planetf1: FYI it appears that for egeria the defaults we end up using are:

        max.poll.interval.ms = 300000
        max.poll.records = 500

planetf1: Personally I'd be a little worried if the consumer didn't poll within 300s ... but note that we also have session.timeout.ms . This could be kept lower, whilst the max.poll.timeout.ms is increased - could be prudent configuration. However I agree with Jeff that there does need to be some limits here to prevent any java client running out of heap if message backlog gets very large (there could be a huge burst for example), so it is certainly better to leave many more of those in the Q.

I don't know the kafka connector well... but There may also be a reliability consideration here. If messages are read off the queue but buffered in the client there is a bigger exposure as the messages will likely not be read should the client/machine die & come up elsewhere, whilst if left in the topic they would?

In my current setup I see auto commit is true, and the auto commit interval is 1s - so reducing the max queue size (as per this PR) also reduces the exposure to lost events? I can imagine in some scenarios we may want them tightened up even further

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
jnhagelberg: @guptaneeru I don't think we are.  Basically, with this change we are limiting the number of kafka message that can be in memory.  The change makes the value of `max.poll.records` more important.  In the worst case, if there is only one consumer and max.pool.records > 100 (which will be most of the time), there will be at most 2*max.poll.records in memory -- one set that is being processed in the other thread and one in queue in this class.  If max.poll.records < 100, then there will be at most 200 in memory -- 100 being processed and 100 in the queue.  There is no way we would end up with a situation where there are 50,000 events in memory.

Issue #865, opened 2019-03-19
Add @generated to IGC generated code


planetf1:
The sonar scan is currently showing very high stats on duplicated code
See https://sonar.odpi.org/component_measures?id=org.odpi.egeria%3Aegeria&metric=duplicated_lines_density

Originally the subject area was a big factor (removing this from scans reduced dup counts to 5% or so). However now the generated IGC code is a large factor.

We can then set sonar to skip duplicate scanning on all lines where @generated is used?

This is far from perfect, and if it is awkward I will stick to exclusion based on selected file paths - it just seems @generated is probably a more generic approach to cater for other scanning tools as it's effectively standard java

cc: @cmgrote 

See also https://github.com/odpi/ci-management/issues/34

cmgrote:
Very happy to make this change — any example you can point me to so I can be sure I get the syntax right? =)

planetf1:
https://docs.oracle.com/javase/8/docs/api/javax/annotation/Generated.html

it seems to still be in Java 11 (which we should compile with)

https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/annotation/processing/Generated.html

though there appear to have been some changes done around standard annotations after java 8 -- so hopefully things won't break (not tested myself yet)

cmgrote:
I'm testing this out on my own (in my own Sonar instance), but it doesn't seem that `@Generated` is automatically causing duplication checks to be ignored.  Is there something somewhere I need to configure to ensure that `@Generated` is used to ignore duplicated lines?

(Also as a heads-up, it also seems to have increased the processing time of the Sonar report by 10x -- hopefully avoidable by whatever extra configuration I need for the above point?  Otherwise I'd suggest we're probably back to just ignoring paths -- eg. anything whose code package has `generated` in it...)

cmgrote:
... in fact, this documentation seems to imply that for duplicate code checks, the only way to ignore them is to exclude the files themselves (via a file pattern):

https://docs.sonarqube.org/latest/project-administration/narrowing-the-focus/#header-3

(?)

planetf1:
Indeed - by default that is the case. In my personal nexus scan I do just that.
However since @generated is fairly standard java, the sonar docs give a tip on defining a pattern that (a little crudely) removes @generated tagged files from the scan
I explain the thought in odpi/ci-management#34

cmgrote:
[...]

PR #876, opened 2019-03-25
Adds @Generated annotation to address #865

cmgrote:
Still needs other changes to the Sonar check / rules themselves to actually do something with this `@Generated` annotation, per #865 ...


planetf1:
The sonar build changes were made under https://github.com/odpi/ci-management/issues/34 - so the sonar scans are now looking better, with generated code removed from dup checking ( and other tests) - not perfect but seems to be a prudent approach

PR #883, opened 2019-04-02
IGC samples load fix

cmgrote:
Fixes samples load to IGC for environments where the tiers are split across various servers (previously worked only when all tiers were co-located on a single machine).


cmgrote:
Already tested -- only changes are to Ansible playbooks used to load IGC...

PR #884, opened 2019-04-02
Fix open-metadata-conformance-suite

bogdan-sava:

### DIFF
--- a/open-metadata-conformance-suite/pom.xml
+++ b/open-metadata-conformance-suite/pom.xml
@@ -65,29 +65,39 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-assembly-plugin</artifactId>
-                <version>3.1.0</version>
-                <configuration>
-                    <descriptorRefs>
-                        <descriptorRef>jar-with-dependencies</descriptorRef>
-                    </descriptorRefs>
-                    <archive>
-                        <manifest>
-                            <addClasspath>true</addClasspath>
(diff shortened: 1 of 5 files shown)


bogdan-sava:
To run Open Metadata Conformance Test, after successful build, run in terminal:

`java -jar open-metadata-target/open-metadata-conformance-suite-1.1-SNAPSHOT.jar {server-name} {server-url}`

Ex:

`java -jar open-metadata-target/open-metadata-conformance-suite-1.1-SNAPSHOT.jar omas http://localhost:8080`

Result is:


```
===============================
Open Metadata Conformance Test  
===============================
Conformance Report for server: omas (http://localhost:8080)

Number of tests: 799
Number of tests passed: 799
Number of tests failed: 0
Number of tests skipped: 0

Congratulations, omas server at http://localhost:8080 is an open metadata repository
```

mandy-chessell:
@bogdan-sava This fix does not seem to change the conformance suite itself.  Instead it is removing the use of GSON in other modules.  Is that right?  Does that mean we need to add a explicit requirement to the egeria method to ban the use of GSON?

bogdan-sava:
Yes. It does two things: removes gson and replace the assembly plug-in with mavenj-jar-plug-in. After removing gson, worked from calling main method but still not from jar generated with assembly.
I know from experience that jar plug-in is more reliable, so I use it without more investigation.

I think mixing Jackson with gson is not good idea. However, if it is really needed, we can investigate about incompatibility and how to make them work.

mandy-chessell:
Thanks for fixing this :)

PR #920, opened 2019-04-15
Enable specification of egeria repo mode for VDC chart  

planetf1:
* Supports use of repo mode to enable the local graph repository (for example by specifying 'egeriarepomode: local-graph-repository')
* Fixes VDC chart to support non-default namespace (queries and rbac)
* Save git branch/log information in docker image for aiding debug


cmgrote:
Looks like the `check-availability.sh` script can now determine the k8s namespace automatically and dynamically? 👍  Probably worth removing the `NAMESPACE` env var from the `configure` image's `Dockerfile` to avoid any confusion that it is picked up and used.

planetf1:
I did notice namespace was in the container, though it's not a standard k8s environment var .. so it shouldn't clash. it's just not any use with the current script...

I didn't touch CACERT - but it is said in the dockerfile. If we think it's pretty much a constant (which I suspect) I'd be inclined to remove.

I could add a commit with that removal if we think it's a good idea (consider it tidyup - shouldn't affect the working of the chart in my scenario at least)

cmgrote:
We should definitely remove `NAMESPACE` now that the script has changed, as otherwise it will be misleading if people look at the `Dockerfile` and assume that it can be overridden (which it cannot as of this change).  I'd be inclined to do the same with `CACERT` (updating the shell script to include its location directly, as was done with `NAMESPACE`) so that they're all consistent -- it's strange to have some configurable at runtime and others not when they're effectively all part of the same secrets area of k8s (?)

Only question will be that if we're using some other k8s distribution that _may_ not retain identical locations for these things that we'll then need to externalise them again (no idea if that could be something different in OpenShift?)...  What's the rationale for embedding them directly in the script rather than feeding in via `ENV`?

planetf1:
I'm going to propose both to avoid confusion (as you suggest) but not to work on the overriding behaviour. i do hope to try openshift at some time in the not too distant future. a single node install is easy - a proper multinode cluster isn't the best (i have on good advice) though the latest beta has rewritten the installer....

I figure we'll have a few potential changes then anyway so we can focus on it then

i'll make the env changes tomorrow if this pr - or a new one if someone else commits it...

planetf1:
Note that whilst testing I've encountered some issues with building apache atlas - kafka initially, but then some dependency problems with ubuntu and solr (as updates have changed).

I will add a fix to this PR -

planetf1:
To summarise needed atlas changes so far
 1. Uplift kafka version from 1.0 to 2.1.1 (without this atlas will fail to connect to Egeria kafka topic at runtime)
 2. uplift findbugs from 3.0.1 to 3.0.5 (needed to avoid ERROR in build)
 3. Changed mvn invocation in container to use -DskipEnunciate (otherwise build fails)

planetf1:
We now have 'official' odpi builds starting. ( https://github.com/odpi/ci-management/issues/44 )
That is now failing hitting the same error found with the latest fixes here.
I will do some final verification today then would like to merge fairly quickly.
@cmgrote are you ok with the changes so far? Anything else before merging (but I suggest a short delay until I've just had a chance to test the new graph repo is working correctly and there's no major breakage). I don't want to hold up too far though as actually having builds will be a big step forward.

cmgrote:
> @cmgrote are you ok with the changes so far? ... I don't want to hold up too far though as actually having builds will be a big step forward.

Agreed, go for it!

planetf1:
Tested on IKS.
 Atlas, IGC proxy, egeria seem to be working fine (simple sanity test)
 -> Can create a term in IGC, and it is reflected to Atlas, and to Egeria (graph repository)

I did note the samples were not getting loaded anyway (though no job failures). Will persue as a separate issue

cmgrote:
The samples are not loaded as part of the k8s deployment -- it's left as a separate (manual) step to kick-off.

PR #951, opened 2019-04-23
Update GraphEntityMapper as it is throwing "org.janusgraph.core.SchemaViolationException

ruxandraRosu:
Property value cannot be null" when null values is passed for an attribute


grahamwallis:
Thanks Ruxandra - that looks like a good change to me.
I noticed that the branch is out of date - could you update it please?

Issue #952, opened 2019-04-23
Allow Subject Area FVT to be runnable at cli


planetf1:
I wanted to run the subject area fvt against my installation.
However the 'jar' under /Users/jonesn/src/egeria/open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/target was not directly runnable ie:

```
➜  subject-area-fvt git:(master) ✗ java -jar target/subject-area-fvt-1.1-SNAPSHOT.jar                                              git:(master|✚1
no main manifest attribute, in target/subject-area-fvt-1.1-SNAPSHOT.jar
```

I therefore enhanced the pom.xml to use the maven assembly plugin so that I can now run:
```
➜  subject-area-fvt git:(safvt1) ✗ java -jar target/subject-area-fvt-1.1-SNAPSHOT-full.jar                                         git:(safvt1|✚1
Enter a URL. Press enter to get the default (http://localhost:8080.)):
```

We can of course change the names as needed, plus also we may want to package up the tests into a new bundle - which itself will require some assembly work.

However as an interim measure I think this makes a sensible step forward for anything we may wish to run in isolation, outside the ide, and without a complex invocation

Does it look reasonable?

PR #953, opened 2019-04-23
#952 allow SA FVT to be CLI runnable

planetf1:
See #952 for a description of the change

Basically - allow SA FVT to be directly runnable


planetf1:
@bogdan-sava - I proposed this change around the same time you have been removing some ... ;-)
I missed the issues/PR

If there's an issue open we can continue the discussion there, but in essence I think there are reasons for both lean and fat approaches

* A fat all-inclusive jar/archive makes it easy to just 'run' a task or process - probably good when you may want to move the tool around, and likely run it alone against a server (in my case it was to test helm deployments)
* A leaner 'install' with shared libraries where you are likely to amass a grouping of capability & want to avoid duplication -- certainly on a server & in some other environments too.

If you point me to the relevant isssue(s) [or I can create one?] we can extend this to a more general discussion?

planetf1:
I've not noticed a problem with the plugin

Do you mean that when you rebuild this particular module, the resulting jar is not always runnable - ie inconsistent
Or do you mean that when the plugin is used to build a variety of modules, sometimes a runnable jar isn't created

In the latter case it could be a combination of incorrect useage, or bugs - but without evaluating a specific issue it's impossible to know.

I think the approach suggested here works when you have a single runnable you want to easy distribute/run. It's less desirable when combined with other tools used by the same user - but if we don't bundle like this we should consider documenting how to use(run), and/or create a deployment package.

Issue #954, opened 2019-04-23
Fix invocation parms for Subject Area FVT


planetf1:
As per review comment on #748 there is a minor typo in interpreting the command line arguments for the SA FVT. Both the userid and servername are picked up from the second parm 

I've made the minor correction so that one can now issue with the parms
http://cloudserver:1234 myservername myuser

and these are correctly interpreted

PR #955, opened 2019-04-23
#954 Fix up use of command line arguments in SA FVT

planetf1:
See #954 - fixed minor typo in pulling cli parms
Also added a space on the interactive prompt, as otherwise when displaying URL the ':' may be mistaken for the actual url string


planetf1:
@davidradl I think we can merge this?

PR #963, opened 2019-04-25
Update Data Engine types

popa-raluca:


open-metadata-implementation/repository-services/repository-services-implementation/src/main/java/org/odpi/openmetadata/repositoryservices/archivemanager/opentypes/OpenMetadataTypesArchive.java
```diff
@@ -7724,6 +7593,7 @@ private void addArea2Types()
         this.add0265AnalyticsAssets();
         this.add0270IoTAssets();
         this.add0280ModelAssets();
+        this.add0290PortsAndWires();
```
cmgrote: I can't remember: do we still have "wires"?  If not, would suggest changing the name of this method to just `add0290Ports()` to avoid any future confusion...

cmgrote:
Otherwise looks good from my scan through the code...  @mandy-chessell 👉 ?

Next on my list is to attempt to map the DataStage metadata across to this: but first need these types to map against 😉

Issue #960, opened 2019-04-24
Migrate IGC connector to its own repository


cmgrote:
Objectives:

- Demonstrate building an external proxy connector (outside core Egeria codebase)
- Demonstrate how to dynamically include / bind such a connector into a running environment (eg. Helm-orchestrated containers)

cmgrote:
It looks likely that we'll need to change the packaging mechanism for Spring Boot in the server chassis to use `ZIP` instead of the default `JAR` -- to be able to load external dependencies (like connectors) at runtime (using `loader.path`) rather than forcing everything to be pre-bundled into a single jar file.

Unless we want to have some interim deployment step that re-bundles the jar file by injecting external connector jars into its `BOOT-INF/lib`?

See:

- https://stackoverflow.com/questions/30745132/external-library-folder-for-spring-boot
- http://weyprecht.de/2018/05/02/how-to-really-package-and-deploy-a-spring-boot-application/
- https://docs.spring.io/spring-boot/docs/current/maven-plugin/usage.html

planetf1:
This will presumably also affect the ability to do java -jar XXX.jar which is how the current docker containers launch the spring boot app, so will need to change there? However reading the docs it does seem as if the change is needed almost for this very reason - to get additional flexibility in the loading process...

cmgrote:
You can still run the executable jar file using `java -jar XXX.jar`, this approach just allows you to additionally provide a `-Dloader.path=...` to the same command.  So you can run:

```
java -Dloader.path=... -jar XXX.jar
```

Without using this approach, the `-Dloader.path=...` is ignored, as is any `-cp`, `CLASSPATH` environment variable, etc.

So this way we could even still use the same 'vanilla' Egeria Docker image for even these custom connector proxies -- just by init'ing the container first to mount in the additional connector jar(s) and changing the entrypoint / CMD args to include the mounted path with `-Dloader.path=...`

PR #964, opened 2019-04-25
Removal of IGC connector from this repository

cmgrote:
... as part of migration to its own dedicated repository, per #960.

@bogdan-sava and @mandy-chessell can you please review the change in https://github.com/odpi/egeria/commit/08f7bb6a1c6bf318c2c8d1022bebb2ad6c947729 and comment on whether this may cause problems anywhere?  This change is to ensure that we can include external libraries (drivers, other connectors, etc) as part of starting up the OMAG Server Platform (ie. via Spring).  It basically changes the manifest of the jar file to use Spring's `PropertiesLauncher` rather than the `JarLauncher` (the latter doesn't seem to allow any classes outside the jar file of OMAG Server Platform itself).  With the `PropertiesLauncher` we can include additional classes (jar files) at startup of the OMAG Server Platform via a `-Dloader.path=...`, allowing us to at runtime include non-core connectors (like IGC, Atlas, etc).

All of the other changes simply remove code that was added specifically for IGC connectivity...

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/charts/vdc/templates/apache-atlas.yaml
+++ b/open-metadata-resources/open-metadata-deployment/charts/vdc/templates/apache-atlas.yaml
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright Contributors to the Egeria project.
 ---
+{{ if .Values.atlas.enabled }}
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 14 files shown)


cmgrote:
See also the PR in new IGC connector repository, which should probably be merged first for continuity: https://github.com/odpi/egeria-connector-ibm-igc/pull/1

planetf1:
I added a note to the PR... the invocation of the server chassis in our helm charts will need to be updated accordingly.

PR #995, opened 2019-05-03
Updates to IGC Helm template

### LINKED ISSUE #976: Update IGC proxy configuration in Helm chart
Post-migration of the IGC connector, update the proxy configuration of the Helm chart for IGC to dynamically load the connector jar as part of the initialisation of an otherwise vanilla Egeria (core) container image.

cmgrote:
Updates the IGC Helm template for VDC chart to fix #976:

- Points to static ODPi GitHub repo location for obtaining IGC repository connector (for now)
- Simplifies the IGC template (removes duplicate sections across conditional statements)
- Updates ConfigMap, Values and Proxy Init job to use new parameters of the connector

Note: the updated template will not actually work until a new Egeria (OMAG Server Platform) Docker image has been built and released that uses the new packaging mechanism included in #964.

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
+++ b/open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
@@ -15,8 +15,10 @@ RUN mvn clean install -DskipTests
 RUN git log > git.log && echo ${branch} > git.branch && git remote -v >> git.remote
 
 FROM openjdk:8-jre-alpine
-COPY --from=build /root/egeria/open-metadata-distribution/open-metadata-assemblies/target/egeria-*-omag-server/servers/omag-server-*/server-chassis-spring-*.jar /server-chassis-spring.jar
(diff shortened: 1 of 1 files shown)


planetf1:
Have started a new build of images from master (official builds/id not quite ready yet).

cmgrote:
I tested the contents of this PR locally against a local Docker image built against master, and the PR worked prior to submitting; but please do (/ happy to) re-test once we have the latest images publicly available on Docker Hub 👍

planetf1:
Have hit some issues with building (inc resource, travel etc) but FYI current block is
https://github.com/odpi/egeria/issues/1006

Will try and investigate when time allows, but I think we need to hold off merging this change until the base images are in place

cmgrote:
> Will try and investigate when time allows, but I think we need to hold off merging this change until the base images are in place

That's fine, just be aware that the IGC adapter code is already removed from this repo, so the base images will no longer have any IGC connectivity embedded once built & put into Docker Hub.  (So we should try to merge these very shortly thereafter for continuity of the IGC piece 😉)

planetf1:
I've now made the other nec essary changes so suggest we merge these all together

cmgrote:
@planetf1 I'll let you "orchestrate" the merges, if that works? 😄

planetf1:
Yes, and I'll also deploy/test

PR #1019, opened 2019-05-14
Avro support for asset owner - part 1

planetf1:
Initial support for avro file format for asset owner
 - based on CSV support
 - only supports string identifier for avro schema
 - url and name fields added to types archive (but both set same)
 - avro connector added, but many methods disabled (commented) along with unit tests

Opening PR to facilitate integration. However further work needed
 - proper avro schema support
 - Completion of connector
 - Unit testing
 - General code cleanup

See also planetf1/assetowneravro2 for a working integration using some test client code
cc: @grahamwallis @davidradl


schannamallu:
recheck

planetf1:
This code will be refactored & new PR opened

Issue #1041, opened 2019-05-15
Separate solr into its own image and re-use


cmgrote:
Both the Apache Ranger and Apache Atlas images need to make use of Solr (the latter requires v5.5, and the former also supports that as a minimal version).  Would therefore be good to separate out Apache Solr into its own container and service, and have both the Apache Atlas and Apache Ranger containers (pods) make use of the separate Apache Solr service.

cmgrote:
When I looked at this a bit more closely it looked like it could be a lot of work to separate things out in the containers (as Solr collections must be initialized / created before pointing Atlas / Ranger at it), creating sequencing challenges, version challenges (ensuring the right Solr objects are created depending on the precise Atlas / Ranger version), etc -- combined with the likelihood that the savings here are probably fairly minimal (~100 MB memory, < 1GB image footprint) I'd suggest putting this as a fairly low priority...  Might be more effort than it's worth 🤷‍♂

planetf1:
I'd suggest we just close if we don't think we'll do it... There's a lot that could be improved with ranger, but we'll likely only tackle it when we want a real focus on building some scenarios - and even then this is probably low down in the list.

PR #1043, opened 2019-05-15
Container image optimisations

cmgrote:
Egeria image changes:

- Fixes permissions on the Egeria image to run as non-root "egeria" user, and moves all contents into `/opt/egeria/...` rather than the root filesystem

Apache Ranger image changes:

- Moves Apache Ranger password setup into configmap, rather than hard-coding
- Uses postgresql database backend rather than mariadb
- Uses initContainer approach for ensuring database is running before starting (removing need for dockervize)
- Applies multi-stage build to obtain a minimal runtime image of ~580MB (vs previous 4.4GB)
- Executes Ranger and Solr processes as dedicated non-root users (unfortunately must run the container as root still due to the tightly-coupled setup scripts of Ranger: for creating various log directories and deploying database entries)

Also adds an Apache Solr pod based on the Docker Official image (as basis for future optimisation under #1041).

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/SubjectAreaErrorCode.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/SubjectAreaErrorCode.java
@@ -303,6 +303,11 @@ public enum SubjectAreaErrorCode {
             "Cannot update a Line if no properties are supplied",
             "The system is unable to process the request.",
             "Correct the code in the caller to only issue updates if there is something to update."),
+    PROJECT_CREATE_WITHOUT_NAME(400, "OMAS-SUBJECTAREA-400-068 ",
+            "Cannot create a Project without a name",
+            "The system is unable to process the request.",
+            "Correct the code in the caller to create a Project with a name."),
+
 
     OMRS_NOT_INITIALIZED(404, "OMAS-SUBJECTAREA-404-001 ",
             "The open metadata repository services are not initialized for the {0} operation",
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/exceptions/InvalidParameterException.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/ffdc/exceptions/InvalidParameterException.java
@@ -3,12 +3,15 @@
 package org.odpi.openmetadata.accessservices.subjectarea.ffdc.exceptions;
 
 
+import java.util.Map;
+
 /**
  * The InvalidParameterException is thrown by the Subject Area OMAS when a parameter is null or an invalid
  * value.
  */
 public class InvalidParameterException extends SubjectAreaCheckedExceptionBase
 {
+    private Map<String,Object> relatedProperties;
     /**
      * This is the typical constructor used for creating a InvalidParameterException.
      *
@@ -39,5 +42,13 @@ public class InvalidParameterException extends SubjectAreaCheckedExceptionBase
     public InvalidParameterException(int  httpCode, String className, String  actionDescription, String errorMessage, String systemAction, String userAction, Throwable caughtError)
     {
         super(httpCode, className, actionDescription, errorMessage, systemAction, userAction, caughtError);
+    }
+
+    public Map<String, Object> getRelatedProperties() {
+        return relatedProperties;
+    }
+
+    public void setRelatedProperties(Map<String, Object> relatedProperties) {
+        this.relatedProperties = relatedProperties;
     }
 }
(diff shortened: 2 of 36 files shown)


askb:
recheck

open-metadata-resources/open-metadata-deployment/docker/ranger/Dockerfile
```diff
@@ -1,62 +1,26 @@
-# SPDX-License-Identifier: CC-BY-4.0 
-# Copyright Contributors to the Egeria project. 
-# ---
-# Docker configuration file for Ranger
-
-# Let's use UBUNTU
-FROM            ubuntu:18.04
-
-# Image metadata following http://label-schema.org/rc1/
```
planetf1: I notice this change has dropped the metadata labels.
One could discuss how relevant they are, though we should probably have them in all or images, or none
Some tips at https://medium.com/@chamilad/lets-make-your-docker-image-better-than-90-of-existing-ones-8b1e5de950d
We could perhaps replace the version with a simple build date, as that is easy to automate.
Probably the best approach is to simply open up a new issue to correct the tags, and allow this change to go through.

open-metadata-resources/open-metadata-deployment/charts/vdc/values.yaml
```diff
@@ -135,11 +146,9 @@ image:
     name: infosvr
     tag: v11.7.0.2
 
-  configure:
-    name: egeria-configure
-
-  nameOverride: ""
-  fullnameOverride: ""
+# These should not be embedded under 'image' (?)
```
planetf1: correct. (they are only used to override the base chart name)

open-metadata-resources/open-metadata-deployment/docker/ranger/Dockerfile
```diff
@@ -66,109 +30,67 @@ RUN mvn -Pall -DskipTests=true clean compile package install assembly:assembly
...
-RUN sed -i "s|SOLR_DOWNLOAD_URL=|SOLR_DOWNLOAD_URL=http://www.mirrorservice.org/sites/ftp.apache.org/lucene/solr/${solrver}/solr-${solrver}.tgz|" install.properties
-
-#Update path to include java, so that when we run the container it is available
-#default shell isn't bash :-(
-
-#RUN echo "export JAVA_HOME=/usr/lib/jvm/default-java" >> /root/.bashrc
-RUN echo "export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64" >> /root/.profile
-RUN echo "export PATH=${JAVA_HOME}/bin:$PATH" >> /root/.profile
-RUN chmod +x /root/.profile
+RUN sed -i "s|SOLR_DOWNLOAD_URL=|SOLR_DOWNLOAD_URL=http://www.mirrorservice.org/sites/ftp.apache.org/lucene/solr/7.7.1/solr-7.7.1.tgz|" install.properties
```
planetf1: the solr version can be a little fragile - usually the last point release is left. Just a note.

planetf1:
A useful set of changes -- thanks. I think we should open an issue to consolidate/discuss image metadata. I'll pull and build to test.

cmgrote:
Agreed — consistent image metadata is needed.  I’m happy to be assigned that in a new PR.

open-metadata-resources/open-metadata-deployment/docker/ranger/Dockerfile
planetf1: Opened https://github.com/odpi/egeria/issues/1048 to follow up

planetf1:
@cmgrote failed rat check

open-metadata-resources/open-metadata-deployment/docker/ranger/dist/ranger-docker-start.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
planetf1: Missing license declaration

open-metadata-resources/open-metadata-deployment/docker/ranger/Dockerfile
```diff
@@ -66,109 +30,77 @@ RUN mvn -Pall -DskipTests=true clean compile package install assembly:assembly
...
+RUN sed -i "s|^SQL_CONNECTOR_JAR=/usr/share/java/mysql-connector-java.jar|SQL_CONNECTOR_JAR=/usr/share/java/postgresql-jdbc.jar|g" install.properties
 RUN sed -i "s|^audit_solr_urls=|audit_solr_urls=http://localhost:6083/solr/ranger_audits|g" install.properties
-RUN sed -i "s|^rangerAdmin_password=|rangerAdmin_password=${rootpw}|g" "${dadir}/install.properties"
-RUN sed -i "s|^rangerTagsync_password=|rangerTagsync_password=${rootpw}|g" "${dadir}/install.properties"
-RUN sed -i "s|^rangerUsersync_password=|rangerUsersync_password=${rootpw}|g" "${dadir}/install.properties"
-RUN sed -i "s|^keyadmin_password=|keyadmin_password=${rootpw}|g" "${dadir}/install.properties"
 
-# Expose web ports
-EXPOSE          6080 6182 6083 6183
+# Create minimal runtime image
```
planetf1: One example here ... we could probably leave this issue open -- I think we could probably look through how a consumer may use the metadata. One specific example I was thinking of was the build date (even though we have the git history on disk)

cmgrote: Per #1048 the build date is already tracked by Docker itself (as are the exposed ports)...  I'm not keen on tracking the same metadata in duplicate ways that then either must be kept in sync or risk confusing consumers as to which of the conflicting metadata is correct (?)

planetf1: will close as review and consider discussion in #1048

planetf1:
testing shortly and will merge later if good.

planetf1:
changes look good. Was able to deploy, access ranger UI, manipulate policies. Updating and merging

PR #1084, opened 2019-05-29
Egeria release 1.0 changes for release to maven central

planetf1:


planetf1:
DO NOT merge yet. Waiting to get very build verification enabled for release 1.0 stream. See https://github.com/odpi/ci-management/issues/43

planetf1:
> It does feel strange that we need to have `-SNAPSHOT` against a release, but I guess if it works fine by me 🤷‍♂

Yes... but the linux foundation scripts are setup to assume that. They are used by a number of different teams. As part of preparing a build for staging the scripts strip the snapshot part & create ready-to-go release images ie 1.0 1.1 etc. Then when we're ready a request to the helpdesk will release what is in staging to maven central. That's the theory....

PR #1086, opened 2019-05-30
Add documentation on submitting report metadata

ruxandraRosu:
Signed-off-by: ruxandraRosu <ruxandra-gabriela.rosu@ing.com>


open-metadata-implementation/access-services/information-view/docs/user/java-events/semantic-assignment.json
```diff
@@ -0,0 +1,22 @@
+{
+  "class": "SemanticAssignment",
+  "eventVersionId": 1,
+  "tableColumn": {
+    "guid": "b1c497ce.60641b50.0v9mgsb23.pa2dds7.77ucei.kgbl5vs0vlcvh54p1cr2d",
```
planetf1: These are slightly unusual GUIDs - as normally the standard is hex (and diff format) but presumably this is a representation of another external tool which has more odd GUIDs?

ruxandraRosu: The guid format is IGC specific in this case.

PR #1100, opened 2019-06-03
Lineage: Deletion of relationships

GuyAckermans:


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
```diff
@@ -216,4 +236,22 @@ private void processUpdatedEntityEvent(EntityDetail originalEntity, EntityDetail
...
+            GlossaryTerm glossaryTerm = glossaryHandler.getGlossaryTerm(relationship.getEntityTwoProxy(), serverUserName);
+
+            deletionEvent.setGlossaryTerm(glossaryTerm);
+            deletionEvent.setEntityGuid(relationship.getEntityOneProxy().getGUID());
+            deletionEvent.setEntityTypeDef(relationship.getEntityOneProxy().getType().getTypeDefName());
+            deletionEvent.setOmrsInstanceEventType(OMRSInstanceEventType.DELETE_PURGED_RELATIONSHIP_EVENT);
+
+            publisher.publishRelationshipEvent(deletionEvent);
+        } catch (Exception e) {
+            log.error(e.getMessage());
```
bogdan-sava: catch Exception in general is not a good practice: swallow all the exceptions, including ones that might be handled in an upper level but they never reach it.

Also error message of log suppose to include some details that makes it easy to identify. Something like:
`log.error("Error process semantic assignment deletion for "+ relationship.getEntityOneProxy().getGUID() + ". Exception message is:  "+  e.getMessage());`

.gitignore
```diff
@@ -80,6 +80,9 @@ omag.server.*.config
 .java-version
 *.graphml
 
+# Ignore Lineage repository
+egeria-lineage-repository/berkely/**
```
planetf1: One obervation that applies across egeria -- we really need to ensure that, without too much config, all data (writing) is constrained to a single/small number of folders. this is essential when working with containers which are basically read-only, but with specific read-write volumes mounted at a specific location. Each volume has overhead - sometimes sys admin created, so having files in many places would be tricky.

I will raise a seperate issue to track this, since it's much broader than just lineage - which is only following the existing pattern! .. and there isn't really precedent set yet.

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/util/GraphConstants.java
```diff
@@ -0,0 +1,25 @@
...
+
+    public static final String PROPERTY_NAME_GUID                             = "guid";
+    public static final String PROPERTY_NAME_QUALIFIED_NAME                   = "qualifiedNameÁ";
+    public static final String PROPERTY_NAME_NAME                             = "name";
+
+    /*
+     *  Elements
+     */
+
+    public static final String PROPERTY_KEY_PREFIX_Element                  = "ve";
```
bogdan-sava: should it be all capital letters?

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/util/GraphConstants.java
```diff
@@ -0,0 +1,25 @@
...
+
+package org.odpi.openmetadata.governanceservers.openlineage.util;
+
+public class GraphConstants {
+
+
+    // Short names for core properties
+
+    public static final String PROPERTY_NAME_GUID                             = "guid";
+    public static final String PROPERTY_NAME_QUALIFIED_NAME                   = "qualifiedNameÁ";
```
bogdan-sava: typo in the string

planetf1:
Docker build should not be a blocking check on PR. See referenced issue above. For now will force
recheck as the build failed due to infra issues

planetf1:
recheck

planetf1:
recheck

PR #1113, opened 2019-06-07
Changes Confidentiality terms to a naming convention that embeds the level

cmgrote:
Per email exchange -- to reflect the change that there is no longer a ConfidentialityLevel enumeration, and we need to store the integer value for the level somewhere (suggestion here is to embed it in the name of our confidentiality level terms).

To clarify: this is **only** for the IGC sample data (where we need special ways to represent classifications).


cmgrote:
@danielaotelea this change to the sample data, and the one linked (odpi/egeria-connector-ibm-igc#15) should give you the fix for the Confidentiality change.

I'm not sure if you'll be able to simply re-load this changed XML file into your environment to get the changes, or you'll need to manually rename the terms under Confidentiality to pre-pend them with numbers (might be quicker / easier anyway as a one-off?)  Just be sure to pull in the changes in the connector (linked PR above) and re-build the connector to get the new mapping behaviour 😄

PR #1115, opened 2019-06-10
Support multiple semantic assignments for a table, report or dataview column

ruxandraRosu:


cmgrote:
I think that would be a problematic assumption in the grand scheme of things, given the `SemanticAssignment` relationship itself allows many-to-many...  (If we wanted a single assignment I'd say we need a separate relationship (type) to capture it?)

mandy-chessell:
You are correct, we need to allow for multiple semantic assignments in the model, OMRS and OMASs.  Event if an organization wants to restrict the assignment to only one active at a time, with effectivity dating, there may be more than one attached event if only one is active.

davidradl:
@planetf1 I think the change is fine. I am curious as to what prompted the need for this omas change. I assume it was not effectivity considerations.

ruxandraRosu:
We discussed previously that single semantic assignment is ING specific and should not be part of egeria logic. I think this restriction should be handled on virtualizer side, this is why I've added the TODO comment. I've made this change in order to be consistent with the model logic and to support use cases that should allow for multiple assignments, as requested in hackatons. Please let me know if any concerns regarding this change.

planetf1:
Thanks for clarification @ruxandraRosu - that sums up the then and now well :-)

open-metadata-implementation/access-services/information-view/information-view-api/src/main/java/org/odpi/openmetadata/accessservices/informationview/ffdc/exceptions/runtime/AddEntityException.java
```diff
@@ -0,0 +1,14 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.accessservices.informationview.ffdc.exceptions.runtime;
+
+public class AddEntityException extends InformationViewUncheckedExceptionBase {
```
bogdan-sava: I suggest renaming InformationViewUncheckedExceptionBase to InformationViewExceptionBase
It is obviously unchecked

ruxandraRosu:
recheck

planetf1:
recheck

bogdan-sava:
recheck

bogdan-sava:
recheck

planetf1:
The build issue has been reported to, and initially responded to. We are waiting for the PR https://github.com/odpi/ci-management/pull/112 to be approved/merged. This is in ci-management which is managed by the linux foundation build team.

bogdan-sava:
recheck

Issue #1117, opened 2019-06-12
findbugs/spotbugs/pmd scans during build


planetf1:
* Investigate adding findBugs or spotBugs (it's effective successor) into the maven build
 - with appropriate security rules
 - this tool can analyze the generated virtual machine code for security issues

* Also look at PMD for further analysis
* See if they can be nicely used within IntelliJ (for developer productivity) as an option
* See if sonarcloud.io scans can make use of the analysed pmd/findbugs data

planetf1:
See also https://community.sonarsource.com/t/spotbugs-configuration-for-sonarcloud-io/10704

planetf1:
See https://medium.com/@prasincs/open-source-static-analysis-for-security-in-2018-part-2-java-f26605cd3f7f for some opinions on available tools.

PMD & spotbugs/securitybugs now optional in build
OWASP checking not run in maven build (yet?) as we have our own nexus scan

PR #1125, opened 2019-06-13
Add optional profile to run spotbugs & pmd code analysis 

planetf1:
See #1117 

This PR enables an OPTIONAL profile that will run 'spotBugs', a bytecode scanner (& the successor to findBugs) with the security bugs plugin. Together they are regarded as good practice , however more evaluation is needed to determine if this should be run by default since we already sun Sonar. The current security focussed rules are covered at https://find-sec-bugs.github.io/bugs.htm

The PR is necessary to make it easy to augment our Sonar reports with the result of spotBugs which is done by integrating to sonarcloud.io as an external analyzer.

This PR also allows anyone to easily run an analysis from the command line via
`mvn clean install -DfindBugs`
but note the scan is intensive and will take significant time (10s of minutes)

Future PR may add support for PMD which is an alternative scanner more focussed on source code analysis

We should not run these by default until/if it's decided which offer the best value, and in which phase which will require further analysis of the results generated


planetf1:
PMD has also been added

PR #1126, opened 2019-06-14
Dockerbuild

akhilreddyjirra:
The additional Dockerfiles.build files will reduce the build time. As in normal scenario, the build store the images in docker local repo. If you are using Jenkins or any ci automation tools to build every time takes hours to complete the build.  

I am assuming this patch will reduce the build time by calling docker build images from odpi repo during docker build. 
Signed-off-by: Rama Krishna akhilreddyjirra@gmail.com


planetf1:
So far we've been building our docker images using
mvn clean install -Ddocker -Ddocker.repo=planetf1
or similar (for me to push to dockerhub under my id, the default is odpi)

This is run under jenkins, then pushed to the odpi repo.

Since these scripts are those which create the odpi repo, I can't see that we can use the odpi repo?
We've been wary of caching when the underlaying code used may change, and we pull down the archive to build.

Will comment further in the PR and really appreciate any suggestions and ideas to improve the process, and to cater for a broader range of use cases too

planetf1:
Thanks for the contribution.

From reviewing the proposal I see you are splitting the docker build in two -- one for the intensive build/compile process, and then another for the customization of the runtime.


Firstly let me take the egeria image. The build step involves extracting source code from egeria master, which we are developing. So this is changing many times a day. As such we want the egeria docker image to be rebuilt with this latest content. That means we can't cache, but must either re-run the git clone/build, or we inject the built components from another jenkins build. We opted for a full build as cleaner. In this scenario I'm not clear on where the time saving is with the new approach? WHen developing, the developer's docker environment can cache layers in any case?

We then have other components we use. Unfortunately there's no obvious candidates for prebuilt images we can just configure and consume (if only!), so we followed the same approach - rebuild. Apache Atlas we will track v2, and so need to regularly rebuild. Ranger a little less so. If we could determine what triggers mean we want a build it would reduce the frequency, and this is certainly an area we could look at

open-metadata-resources/open-metadata-deployment/docker/atlas/README.md
```diff
@@ -0,0 +1,22 @@
+
```
planetf1: This would fail our license check as we need a header such as
```
# SPDX-License-Identifier: Apache-2.0
# Copyright Contributors to the Egeria project.
```

planetf1:
In terms of our ODPi build process specifically......>

If we are to consider splitting & caching of the base layer, we would *ideally* want to only build if there have been git commits in the branch we are building within the image.

For apache-atlas-fork this will be rare, as it's a stable fork we're managing ourselves
For apache atlas & ranger it's constant as the apache teams work
... so no optimization 

Pragmatically we could reduce the frequency of build of atlas & ranger to - perhaps - weekly, assuming there was also a way for one of the egeria maintainers to trigger a new build

We change the customization typically no more than this in any case, which again would mean splitting wouldn't really help..
For egeria we should leave as is.

akhilreddyjirra:
> We change the customization typically no more than this in any case, which again would mean splitting wouldn't really help..
> For egeria we should leave as is.

Okey. i thoght got building

cmgrote:
@akhilreddyjirra thanks for the ideas!

@planetf1 actually for Apache Atlas the new (non-fork) build uses the release source itself -- which as far as I'm aware shouldn't ever change.  So this could be a worthwhile optimisation to put in just for the `apache-atlas/Dockerfile` split, as the runtime image is more likely to be the one that we make some minor changes to going forward (eg. changing parameters in its config), but the build process itself for the sources -- being built from the source tarball of the release and not git `master` -- should be quite stable.

@akhilreddyjirra could you open a new PR with just the changes you'd made to split the `Dockerfile` under `apache-atlas` directory?

PR #1127, opened 2019-06-14
Apache Atlas docker and k8s changes

cmgrote:
- Configures vanilla Apache Atlas container to tail all log files by default (not only `application.log`)
- Adds container-external accessibility to embedded Apache Atlas Kafka event bus
- Adds parameters for Apache Atlas container (eg. credentials, endpoints, etc)
- Swaps forked Apache Atlas container as default to the vanilla Apache Atlas container
- Configures Apache Atlas proxy as part of `egeria` container itself, rather than a separate container

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/docker/atlas/Dockerfile
+++ b/open-metadata-resources/open-metadata-deployment/docker/atlas/Dockerfile
@@ -19,6 +19,7 @@ ENV PATH /usr/java/bin:/usr/local/apache-maven/bin:/usr/local/sbin:/usr/local/bi
 WORKDIR /root
 
 # Pull down Atlas and build it into /root/atlas-bin.
+ADD https://api.github.com/repos/grahamwallis/atlas/git/refs/heads/master version.json
 RUN git clone https://github.com/grahamwallis/atlas -b master
 
 WORKDIR /root/atlas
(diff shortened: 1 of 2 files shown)


planetf1:
From a development perspective it's a good change... I'll float the suggestion with a few people I currently have pulling the charts to check for impact before merging

planetf1:
I've confirmed at least one user is making use of atlas ui for browsing. Likely terms/glossaries. I noted current UI is failing also so opened up https://github.com/odpi/egeria/issues/1129

I suggest we need to hold off merging just yet.

cmgrote:
Sure thing -- let's leave open for now.  I was under the impression that the forked Atlas build may be broken already, in which case this was at least something.  But if not, let's leave as-is for now 👍

planetf1:
Ok. I will double check if it works now (it was ok), but have a  few people confirm they are using it. Obviously if we make it optional then we can merge now. I think the availability of Egeria UIs will allow us to make the change finally as I completely agree in not wanting to prolong the use of forked atlas much longer.

schannamallu:
recheck

planetf1:
Can confirm 'old' atlas is working propertly, so continuing to wait until new UI is available in lieu of atlas UI for simple search

Issue #966, opened 2019-04-26
Build warnings due to minor maven problems


planetf1:
```
$ mvn clean install
[INFO] Scanning for projects...
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:open-metadata-conformance-suite-api:jar:1.1-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ org.odpi.egeria:open-metadata-conformance-suite:1.1-SNAPSHOT, C:\Users\nigel\IdeaProjects\egeria\open-metadata-conformance-suite\pom.xml, line 80, column 21
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:open-metadata-conformance-suite-client:jar:1.1-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ org.odpi.egeria:open-metadata-conformance-suite:1.1-SNAPSHOT, C:\Users\nigel\IdeaProjects\egeria\open-metadata-conformance-suite\pom.xml, line 80, column 21
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:open-metadata-conformance-suite-server:jar:1.1-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ org.odpi.egeria:open-metadata-conformance-suite:1.1-SNAPSHOT, C:\Users\nigel\IdeaProjects\egeria\open-metadata-conformance-suite\pom.xml, line 80, column 21
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:open-metadata-conformance-suite-spring:jar:1.1-SNAPSHOT
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ org.odpi.egeria:open-metadata-conformance-suite:1.1-SNAPSHOT, C:\Users\nigel\IdeaProjects\egeria\open-metadata-conformance-suite\pom.xml, line 80, column 21
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.odpi.egeria:open-metadata-conformance-suite:pom:1.1-SNAPSHOT
[...]

PR #1154, opened 2019-06-20
#966 add missing maven-jar version (via PluginManagement)

planetf1:


schannamallu:
recheck

Issue #1167, opened 2019-06-25
Local graph repo - regression? NPE in deployed k8s 


planetf1:
When using the graph repo with the vdc charts the following exception is seen by atlas when starting to process typedefs from the archive:
```
[...]

Issue #1114, opened 2019-06-10
Rationalise poms 


planetf1:
Currently we are pulling too many unnecessary maven artifacts as dependencies for our component. This increases size, complexity, scope for security issues/evaluation.

- 'scope' of dependencies in poms could be made more precise in some cases - ie test/compile/runtime/provided - to avoid drag through into built artifacts
- maven plugins don't need to be dependencies
- whilst versions should be declared in top level pom , specific dependency must be at the very lowest possible level to avoid inclusion
 - transitive dependencies may still need addressing through dependency management vs properties.

A few other issues relate to pom issues but of narrower scope #322 & #22 in particular.

An example of the impact of dependencies at the wrong level can be seen by examining one of our clients:
 
Taking asset-catalog we note:
 
The default scope for a dependency is 'compile' - which means it's used in the build AND is a runtime dependency.
We should carefully use test, runtime, provided etc where we can (see http://andresalmiray.com/maven-scopes-vs-gradle-configurations/ ) and must stop inheriting all from parent
 
We see that this client
 - pulls in lots of spring web/core/beans -- shouldn't be needed for a client
 - includes TEST jars -- not needed by runtime code, only test process
 - includes plugin code - these should be PLUGINS not dependencies
 - includes enunciate for javadoc - this is only build time
 - includes kafka (really? )
 
The visualization (right click on the maven project in IntelliJ) also shows that we have some conflicting dependencies - these need further checking but should be resolved via a combination of DependencyManagement & exclusions
 
There's probably reason to start going through all the poms and attept some cleanup
 - remove plugin dependencies
 - apply correct scoping
 - push down expression of dependency to lowest leaf where ALL submodules use them
 
Our first focus should be the clients. 
 
 
 
[INFO] ----------------< org.odpi.egeria:asset-catalog-client >----------------
[INFO] Building Asset Catalog OMAS Client 1.1-SNAPSHOT                 [80/252]
[INFO] --------------------------------[ jar ]---------------------------------
Downloading from odpi-snapshots: https://nexus.odpi.org/content/groups/public/org/odpi/egeria/asset-catalog-api/1.1-SNAPSHOT/maven-metadata.xml
Progress (1): 782 B^M                   ^MDownloaded from odpi-snapshots: https://nexus.odpi.org/content/groups/public/org/odpi/egeria/asset-catalog-api/1.1-SNAPSHOT/maven-metadata.xml (782 B at 3.5 kB/s)
[...]

PR #1169, opened 2019-06-25
#1167 Fix regression in local graph repo - berkeleydb restored

planetf1:
Fixing up regression caused by pom refactoring removing dynamically loaded berkeley for janusgraph

Also added janusgraph data files to gitignore & rat exclusions


planetf1:
Note: I intend to merge this immediately due to regression caused by #1114 and the scope of this update is restricted to broken code.

Issue #1171, opened 2019-06-26
Ranger docker : update solr


planetf1:
```
[INFO] Wed Jun 26 11:53:28 UTC 2019|INFO|Downloading solr from http://www.mirrorservice.org/sites/ftp.apache.org/lucene/solr/7.7.1/solr-7.7.1.tgz
[INFO] --2019-06-26 11:53:28--  http://www.mirrorservice.org/sites/ftp.apache.org/lucene/solr/7.7.1/solr-7.7.1.tgz
[INFO] Resolving www.mirrorservice.org (www.mirrorservice.org)...
[INFO] 212.219.56.184, 2001:630:341:12::184
[INFO] Connecting to www.mirrorservice.org (www.mirrorservice.org)|212.219.56.184|:80...
[INFO] connected.
[INFO]
[INFO] HTTP request sent, awaiting response...
[INFO] 404 Not Found
[INFO] 2019-06-26 11:53:29 ERROR 404: Not Found.
```

Currently the ranger build uses local docker. At some point in the future we can switch to using our separate solr container. However currently making a minimal fix for now to allow the container to build

PR #1172, opened 2019-06-26
#1171 uplift solr version for embedded use in ranger to prevent build…

planetf1:
… breakage


planetf1:
recheck

PR #1177, opened 2019-06-26
Correct 'scope' for test libraries

planetf1:
Corrected scope of test libraries to be 'test' only (not compile)


planetf1:
recheck

planetf1:
recheck

PR #1178, opened 2019-06-26
Docker split

akhilreddyjirra:
1.  Ranger Docker build is trying to download the solr version 7.2.1 which is updated to 7.2.2 and added argument  value during the build
Ranger docker: update solr #1171 
2. Modification in the Apache Atlas Docker.build to Dockerfile.build and change the apache-atlas to apache-atlas: build
3. Atlas is split to Dockerfile.build and Dockerfile

I am assuming we can trigger the build images often like once in a week. This split will reduce the build time in the main Dockerfile.
Hope this will helpful for devops people like me.


planetf1:
recheck

akhilreddyjirra:
> I like these ideas of splitting the images + adding version numbers as args; I guess these splits will require build infra changes (@planetf1)?
> 
> The Apache Atlas image split we merged already was sort of a special case as that entire image is still a sort of work-in-progress state anyway... For these other images we may want to wait until we have the necessary build infra changes in place (?)


the Apache-Atlas having build issues :
```
16:37:40 [INFO] Step 13/21 : COPY --from=build --chown=atlas:atlas /root/atlas-bin/ /opt/apache/atlas/
16:37:40 [INFO] 
16:37:40 [ERROR] COPY failed: stat /var/lib/docker/overlay2/f53981c60108e05ee6aff77b1ca80b44a4604857d8815a13bf50a1e1a3a64475/merged/root/atlas-bin: no such file or directory
16:37:40 [WARNING] An attempt failed, will retry 1 more times
16:37:40 org.apache.maven.plugin.MojoExecutionException: Could not build image
```
reason : the build image should be **apache-atlas:build** not **apache-atlas:lates**t in docker.build

This PR has the fix along with other splits

planetf1:
I took at the PR (sorry for delay).

Trying to build locally into my repo with 
`mvn clean install -Ddocker -Ddocker.repo=planet` worked for ranger.
Of course it is dependent on odpi:ranger whilst the previous process was more self contained (though dependent on ubuntu, alpine etc)

I think we should have the base ranger build buildable from these same scripts (maven) - perhaps using '-Ddocker.buildbaseimages' or similar? (and we run those less frequently/only on demand) 

Looking at atlas - this failed with


```
[ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-atlas: Could not build image: manifest for odpi/atlas:build not found -> [Help 1]
```

which kind of supports the point above - the end user should be able to build to their own repo easily - and they won't have access to 'odpi' to write to?

I then tried 'apache-atlas' (ie the new, base atlas only image) and got a failure as mentioned in an earlier comment
```
[ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-apache-atlas: Could not build image: COPY failed: stat /data/docker/overlay2/cc0496f7c1f8f95d2418c856b5cab46e6e809fc8349a4381adba39821cedc026/merged/root/atlas-bin: no such file or directory -> [Help 1]
```
- So that fix wasn't in the PR though could be easily added.

My thoughts so far
 * I do like the idea of doing the split
 * Anyone should be able to build their own images, targetting their own repo
 * I think easiest approach is to extend the pom.xml in the relevant directory to build both images (add an extra execution, using Dockerfile.build) - and also provide an option to skip the lengthy base build

schannamallu:
> I took at the PR (sorry for delay).
> 
> Trying to build locally into my repo with
> `mvn clean install -Ddocker -Ddocker.repo=planet` worked for ranger.
> Of course it is dependent on odpi:ranger whilst the previous process was more self contained (though dependent on ubuntu, alpine etc)
> 
> I think we should have the base ranger build buildable from these same scripts (maven) - perhaps using '-Ddocker.buildbaseimages' or similar? (and we run those less frequently/only on demand)
> 
> Looking at atlas - this failed with
> 
> ```
> [ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-atlas: Could not build image: manifest for odpi/atlas:build not found -> [Help 1]
> ```

seems odpi/apache-atlas:build image is not present in docker-hub 

> which kind of supports the point above - the end user should be able to build to their own repo easily - and they won't have access to 'odpi' to write to?
> 
> I then tried 'apache-atlas' (ie the new, base atlas only image) and got a failure as mentioned in an earlier comment
> 
> ```
> [ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-apache-atlas: Could not build image: COPY failed: stat /data/docker/overlay2/cc0496f7c1f8f95d2418c856b5cab46e6e809fc8349a4381adba39821cedc026/merged/root/atlas-bin: no such file or directory -> [Help 1]
> ```

seems odpi/apache-atlas:build image is not present in docker-hub
i will use this PR to make images available in odpi dockerhub
 
> * So that fix wasn't in the PR though could be easily added.
> 
> My thoughts so far
> 
> * I do like the idea of doing the split
> * Anyone should be able to build their own images, targetting their own repo
> * I think easiest approach is to extend the pom.xml in the relevant directory to build both images (add an extra execution, using Dockerfile.build) - and also provide an option to skip the lengthy base build

planetf1:
See also #1224. If we could fix the secondary build stage as part of the PR/split, then in 1224 we (I, if you like!) can look at updating the maven config to also allow build of the base image.

schannamallu:
> > I took at the PR (sorry for delay).
> > Trying to build locally into my repo with
> > `mvn clean install -Ddocker -Ddocker.repo=planet` worked for ranger.
> > Of course it is dependent on odpi:ranger whilst the previous process was more self contained (though dependent on ubuntu, alpine etc)
> > I think we should have the base ranger build buildable from these same scripts (maven) - perhaps using '-Ddocker.buildbaseimages' or similar? (and we run those less frequently/only on demand)
> > Looking at atlas - this failed with
> > ```
> > [ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-atlas: Could not build image: manifest for odpi/atlas:build not found -> [Help 1]
> > ```
> 
> seems odpi/atlas:build image is not present in docker-hub
> 
> > which kind of supports the point above - the end user should be able to build to their own repo easily - and they won't have access to 'odpi' to write to?
> > I then tried 'apache-atlas' (ie the new, base atlas only image) and got a failure as mentioned in an earlier comment
> > ```
> > [ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-apache-atlas: Could not build image: COPY failed: stat /data/docker/overlay2/cc0496f7c1f8f95d2418c856b5cab46e6e809fc8349a4381adba39821cedc026/merged/root/atlas-bin: no such file or directory -> [Help 1]
> > ```
> 
> seems odpi/apache-atlas:build image is not present in docker-hub
> i will use this PR to make images available in odpi dockerhub
> 
> > * So that fix wasn't in the PR though could be easily added.
> > 
> > My thoughts so far
> > 
> > * I do like the idea of doing the split
> > * Anyone should be able to build their own images, targetting their own repo
> > * I think easiest approach is to extend the pom.xml in the relevant directory to build both images (add an extra execution, using Dockerfile.build) - and also provide an option to skip the lengthy base build

@planetf1 
the build images are two heavy
```
$docker images | grep build
odpi/atlas                                               build               7cbbcbce5d4e        1 days ago          10.1GB
odpi/apache-atlas                                 build               c1a5d366816b        1 days ago          10.2GB
```
the build image size is too heavy..,in order to push and pull  builder images take time i think.
Better we can cut down on what modules to build.

cmgrote:
> the build image size is too heavy..,in order to push and pull builder images take time i think.

The build images are really just a step in the process -- not something we'd expect anyone to actually pull from Dockerhub (I think?)  The idea here is just to not take the "hit" of building each time, if we need to make some change to the runtime image (much more likely)...  (Locally this is resolved by Docker itself keeping around its own "cache" of layers that it just re-uses, but I don't think there's any equivalent for Jenkins?)  Could there be some other repository alternative where they could be placed that would make more sense than Dockerhub?

schannamallu:
> > the build image size is too heavy..,in order to push and pull builder images take time i think.
> 
> The build images are really just a step in the process -- not something we'd expect anyone to actually pull from Dockerhub (I think?) The idea here is just to not take the "hit" of building each time, if we need to make some change to the runtime image (much more likely)... (Locally this is resolved by Docker itself keeping around its own "cache" of layers that it just re-uses, but I don't think there's any equivalent for Jenkins?) Could there be some other repository alternative where they could be placed that would make more sense than Dockerhub?

@cmgrote i agree with you but i am worrying about Jenkins. For every time jenkins build job on fresh node .This makes build duration in Hours

cmgrote:
> For every time jenkins build job on fresh node .This makes build duration in Hours

Exactly what we're trying to avoid: having the build image in some repo would mean that it just needs to be downloaded rather than re-built each time; though I guess downloading a 10GB image (download itself will likely be more like 3GB from compression) could still be fairly time-consuming -- and may have other implications as well (?)

planetf1:
Ok a few thoughts

Firstly I think need to enable 2 different scenarios
a) Someone just wants to consume our helm charts -and use the docker images we've already published to dockerhub
b) They want to build all from source and then use those in our charts

If we take atlas and ranger, our main problem is that there's no readily 'built' image of either atlas or ranger that we can quickly bundle up into a docker image - nor is there an existing supported image we can just extend.. so we end up building from source.

If we needed master, even a docker build cache is problematic since the git clone would be different each time. Though we can mitigate this by only relying on release versions of atlas/ranger. That may be ok for ranger - it might not for atlas given we may be trying to get new features into the codebase.

Splitting the image does allow us to build at two different frequencies - infrequent (manual?) for the base atlas/ranger, and more frequent for our, light, customizations/configuration. We were just missing some steps to support someone wanting to build themselves.

If the 'base' is still too heavy
 - We could consider ensuring the 'Dockerfile is suitable for dockerhub's OWN automated build - though this may require a fork.. It would mean odpi doesn't worry about the base
 - We could work with the base atlas/ranger teams by contributing Dockerfile's etc that are suitably generic, but also good enough for us, and see if they can be persuaded to integrate into their build infrastructure at apache -- arguably this is the 'best' solution IMO
 - We can look around for other sources of 'built' code -- for example maven artifacts contain much of what is needed, but not all. Could we take those and augment them with any additional files needed. Maybe get packages modified in base atlas/ranger if needed - that would cut build time
 - we can try and improve the build team. I've contacted the ranger team (mailing list) to see how viable this might be. same for atlas?

planetf1:
@akhilreddyjirra Sorry for the backlog... I've merged some recent docker changes and I'm about to remove the old atlas image entirely.. So if you could merge the ranger dockerfile & remove the atlas update (or wait for me to push #1236 ) that would be super.

From now on we should be more agile as we've decided to drop the old image and just move forward with the new

akhilreddyjirra:
> image entirely



> @akhilreddyjirra Sorry for the backlog... I've merged some recent docker changes and I'm about to remove the old atlas image entirely.. So if you could merge the ranger dockerfile & remove the atlas update (or wait for me to push #1236 ) that would be super.
> 
> From now on we should be more agile as we've decided to drop the old image and just move forward with the new

@planetf1 I will wait for me to push #1236

planetf1:
Ok over to you. I won't make further changes just yet.
Any other changes you want to get in?

Then once in I'll give testing a go and rapidly fix any issues.
The backlog wasn't helping our progress....

Issue #22, opened 2018-06-07
dependencyManagement vs version properties


planetf1:
In the top level POM we have a set of properties such as
        <log4j.version>2.4.6</log4j.version>

Then when we have a <dependency> flag in either this pom, or any child we can use:
            <version>${log4j.version}</version>

This would seem to work, but I wondered whether it was simpler to use the standard maven mechanism for expressing the use of consistent versions, where we would use
<dependencyManagement> flags in the top level pom to declare our default versions, and then use <dependency> as today where each plugin needs to be used

We would do the same for <pluginManagement> to ensure consistent use of plugin versions across the project (in exceptional cases overriding can be done at any level). 

Atlas did use the explicit property mechanism, but I would think going with standard maven is simpler, 

Happy to do a patch if the community feels this is a better approach

mandy-chessell:
What would the flags look like ?

planetf1:
See example in https://github.com/GaianRangerPlugin/ranger-gaian-plugin/blob/master/pom.xml

Basically this uses the 'DependencyManagement' to express the version that should be used in all modules, without actually pulling in as a dependency. In the 'Dependency' section the version is then omitted. A similar pattern applies to the plugins with 'PluginManagement'. This is the 'standard' maven mechanism

One main difference is the effective on transitive dependencies - the maven scheme arguably provides a little more assurance as to what is used.

There's some further docs at https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html

mandy-chessell:
I think we should do this - as the code gets bigger, dependency management is going to become more important.  I would like to take you up on your offer to code the patch (when you are back from vacation :)

planetf1:
Updating top level pom.
Removing any DependencyManagement elsewhere
Added validation - as an example if an inconsistent version is used we now fail the build with
```
[INFO] --- maven-dependency-plugin:3.1.1:analyze-dep-mgt (depmgmt) @ access-services-user-interface ---
[INFO] Found Resolved Dependency/DependencyManagement mismatches:
[INFO] 	Dependency: org.springframework.security:spring-security-web:jar
[INFO] 		DepMgt  : 5.1.4.RELEASE
[INFO] 		Resolved: 5.1.5.RELEASE
```

There's no exception handling yet - will need to add on a case by case basis (overriding <plugin> configuration in local pom.

Discrepancies so far
[...]

PR #1190, opened 2019-06-27
Use DependencyManagement

planetf1:
See #22

This update pulls all Dependency requirements (plugins and dependencies) to the top level pom.
Dependencies (but not plugins) are checked, and any discrepancy will fail the build
Existing poms not updated in this PR - tests pass as effective version is the same (a couple of minor updates - see #22)

A batch pom update will be done in a subsequent PR
cc: @mandy-chessell


planetf1:
@GuyAckermans @DimitriosMaimaris Note the minor JanusGraph version change by 1 mod level to be consistent with the level being used for the egeria graph respository
@bogdan-sava Note a minor correction to spring-security-web, correcting what appears to be a type

Issue #1228, opened 2019-07-01
slf4j error in Asset Management Client (regression after pom refactoring)


planetf1:
After refactoring poms in #1114 and related pom updates, an slf4j runtime error is seen when running the asset management sample (and others).

In the example below a pom change has also been made to build a complete jar with dependencies, which assures a more defined environment than running in IDE
```
➜  asset-management-samples git:(fixslfdep) ✗ java -jar target/asset-management-samples-1.1-SNAPSHOT-jar-with-dependencies.jar fred
===============================
CSV File Reader Sample
===============================
Running against server: cocoMDS1 at http://localhost:8080
Using userId: garygeeke

SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
The connector can not be created from metadata.  Error message is: OMAG-PLATFORM-404-001 The OMAG Server cocoMDS1 is not available to service a request from user garygeeke
===============================
Accessing file: fred
The connector is unable to retrieve the requested record because the file is too short.
```

This was caused when the slf4j-api dependency was removed from 'compile' scope - the mvn dependency:analyze goal does not deem that component as being referenced in the java object code

planetf1:
This error is produced when no slf4j binding can be found on the runtime classpath

Docs at https://www.slf4j.org/manual.html explain in more detail the operation of slf4j. 

One para of note
```
As of SLF4J version 1.6.0, if no binding is found on the class path, then slf4j-api will default to a no-operation implementation discarding all log requests. Thus, instead of throwing a NoClassDefFoundError because the org.slf4j.impl.StaticLoggerBinder class is missing, SLF4J version
1.6.0 and later will emit a single warning message about the absence of a binding and proceed to discard all log requests without further protest.
```
We are prior to 1.6...

It further goes on to provide a recommendation
```
BASIC RULE Embedded components such as libraries or frameworks should not declare a dependency on any SLF4J binding but only depend on slf4j-api. When a library declares a transitive dependency on a specific binding, that binding is imposed on the end-user negating the purpose of SLF4J. Note that declaring a non-transitive dependency on a binding, for example for testing, does not affect the end-user.
```

This appears to suggest that
 * One should NOT include a binding when building components that may be reused
 * Having the same behaviour as no-op -- but without the error -- is CORRECT behaviour
[...]

PR #1231, opened 2019-07-02
#1228 Configure a default logger for access services samples

planetf1:
See #1228 for details


planetf1:
Note that this PR makes two changes
 - Adds logback to access client sample. Configures for >INFO. Result is no debug logging no error
 - Adds a jar with dependencies for each so they can be run directly

PR #1245, opened 2019-07-04
Add validators for report and data view handlers and change data view to support nested levels

ruxandraRosu:


open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/omrs-rest-repository-connector/pom.xml
```diff
@@ -39,6 +39,11 @@
             <artifactId>open-connector-framework</artifactId>
          </dependency>
 
+        <dependency>
```
planetf1: This doesn;t seem to be needed, and is what is causing the PR compile to fail. spring-web is not used in the java code so should not be a dependency

ruxandraRosu:
#recheck

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/omrs-rest-repository-connector/pom.xml
ruxandraRosu: Hi, I fixed this PR, the issue is that locally it is complaining about this dependency. I have deleted .m2 repo, but the issue persists. Any other suggestion I could try for my local build?

ruxandraRosu: Hk Nigel, this is the dependency tree:
[INFO] --- maven-dependency-plugin:3.1.1:tree (default-cli) @ omrs-rest-repository-connector 
[INFO] org.odpi.egeria:omrs-rest-repository-connector:jar:1.1-SNAPSHOT
[INFO] +- org.odpi.egeria:rest-client-factory:jar:1.1-SNAPSHOT:compile
[INFO] |  +- org.slf4j:slf4j-api:jar:1.7.26:compile
[INFO] |  \- org.odpi.egeria:spring-rest-client-connector:jar:1.1-SNAPSHOT:compile
[INFO] |     +- org.springframework:spring-web:jar:5.1.8.RELEASE:compile

ruxandraRosu: Issue is fixed locally, thanks for help!

pom.xml
```diff
@@ -193,6 +193,8 @@
 	    <sonar.coverage.exclusions>**/generated/**</sonar.coverage.exclusions>
 	    <sonar.cpd.exclusions>**/generated/**</sonar.cpd.exclusions>
 
+        <hibernate-validator.version>6.0.10.Final</hibernate-validator.version>
+        <spring-context.version>5.1.8.RELEASE</spring-context.version>
```
lcpopa: Version not needed, since there is already defined the spring-web.version

open-metadata-implementation/access-services/information-view/information-view-spring/pom.xml
```diff
@@ -39,6 +39,17 @@
             <artifactId>information-view-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <version>${spring-context.version}</version>
```
lcpopa: Version needs to be specified in central pom, not here

pom.xml
```diff
@@ -193,6 +193,8 @@
 	    <sonar.coverage.exclusions>**/generated/**</sonar.coverage.exclusions>
 	    <sonar.cpd.exclusions>**/generated/**</sonar.cpd.exclusions>
 
+        <hibernate-validator.version>6.0.10.Final</hibernate-validator.version>
```
planetf1: hibernate-validator needs to be added within `<dependencyManagement>` , using this property as the version.

Consider carefully what version - 6.0.18 seems current. If there's a particular reason to go older, fine -- but generally I'd err towards current levels

It's worth checking dependencies too as at https://mvnrepository.com/artifact/org.hibernate/hibernate-validator - though nothing interesting there.

planetf1: see https://github.com/odpi/egeria/blob/master/developer-resources/Dependency-Management.md

open-metadata-implementation/access-services/information-view/information-view-server/src/main/java/org/odpi/openmetadata/accessservices/informationview/reports/BasicOperation.java
```diff
@@ -230,4 +288,216 @@ protected void deleteRelationships(List<Relationship> existingRelationships) {
...
+                                                                                                    EntityNotKnownException,
+                                                                                                    RelationshipNotKnownException,
+                                                                                                    RelationshipNotDeletedException {
+
+        List<Relationship> relationships = omEntityDao.getRelationships(Constants.SCHEMA_QUERY_IMPLEMENTATION, columnGuid);
+        List<String> relationshipsToRemove = new ArrayList<>();
+        if (relationships != null && !relationships.isEmpty()) {
+            relationshipsToRemove = relationships.stream().map(e -> e.getEntityTwoProxy().getGUID()).collect(Collectors.toList());
+        }
+        for (Source source : sources) {
```
lcpopa: Try to change the 'for' into a stream and reduce the landscape

planetf1:
Interestingly I'm seeing the a test failure in the same module when building from *master* though I don't explicitly see a test failed, even though the module test phase ends with failure . Will update and retry. Oddly in the above checks the java11 build worked, but java8 failed. My test was with java11 - so perhaps something intermittent/inconsistent in that component. Will try some more.

ruxandraRosu:
> Interestingly I'm seeing the a test failure in the same module when building from _master_ though I don't explicitly see a test failed, even though the module test phase ends with failure . Will update and retry. Oddly in the above checks the java11 build worked, but java8 failed. My test was with java11 - so perhaps something intermittent/inconsistent in that component. Or perhaps more likely resource issues - as I was also running -DfindBugs. We also know the formal builds are more resource constrained than local builds. Will try some more.

Hi Nigel,
The test is failing randomly, I don't think it is related to java version. I will update you once I figure out the issue.

Issue #967, opened 2019-04-26
FVT test generation code is taking 60% of overall build time


planetf1:
Currently the open types test generator is taking up 60% (7 minutes) of the overall build time - taking a typical macos build to around 11:30. This can impact developer productivity and has been observed by a number of people.

It would be worthwhile looking into:
 * Why it is taking so long. Can the code be improved?
 * Adding an option to skip - at least for developers building their own code locally

The first would be my preferred option if we do find a significant issue as more validation is good. We should understand what the generator is doing, and whether it is effective - though the second is ok since the open types are updated less often. I would suggest keeping the tests in the commit build regardless.

On skipping, note that I experienced an issue with profiles & submodules when working on https://github.com/odpi/egeria/issues/771 and instead adopted the '-Ddocker' flag to *enable* docker builds (here we want a flag to disable IMO if we go down that route)

cc: @davidradl @bogdan-sava

innerTide:
I'm have the same issue as well. Is there any update? Thanks!
![image](https://user-images.githubusercontent.com/17440495/57459383-6e5e7380-7273-11e9-9ec9-a839702803c7.png)

planetf1:
I see a PR is proposed to support skipping (which I agree with). Let's leave this issue open to improve the performance when this step is undertaken
see https://github.com/odpi/egeria/pull/1008/files

planetf1:
@innerTide as an aside i notice your build is taking 32 mins - that's about 3 times my duration/official build. Maybe you have other issues/limitations too. 
Also note that when I tested previously a windows build was about 2.5x slower than either MacOS or linux (though a build in a vm on windows or using WSL was close to linux)

innerTide:
@planetf1 This is my first build so it took some time to download dependencies from Maven central. But I will try WSL :) Thanks!

planetf1:
Verify builds are now down to 14 minutes . Builds with FVT (master after merging, daily) remain at 40 mins or so on official LF systems

planetf1:
Suggest leaving this open to look at actual test code and confirm it needs to take this long when run

bogdan-sava:
In order to have dev builds I do recommend you to run:
`mvn -pl '!org.odpi.egeria:open-metadata-assemblies' clean install -DskipTests -DskipFVT `

It excludes assemblies also. On my computers finishes under 2 min.

bogdan-sava:
-pl param is a built in functionality in maven

bogdan-sava:
[...]

PR #1258, opened 2019-07-05
#967 run fvt open types generate in separate java vm

planetf1:
This change launches the open types generator in a separate vm using the 'exec' goal of the maven exec plugin rather than 'java'. 
See #967 for original issue


planetf1:
recheck

planetf1:
This is not a good change. The build appears successful, and returns successful, but the last few modules are missing . In retrospect running java via the command line is too subject to environment differences between developers/build (even though it works great locally). 

Need to look at other reasons/approaches for the extended run time when using maven exec plugin to launch the test class generator

PR #1261, opened 2019-07-08
Data platform OMAS & Virtualizer to support multiple busienss terms

cong78:
This PR is about to support multi business terms during semantic assignment, which is the change has been made on Information View OMAS.


ruxandraRosu:
#recheck

PR #1264, opened 2019-07-08
DP OMAS & Virtualizer support multi business terms

cong78:
This PR is about to support multi business terms during semantic assignment, which is the change has been made on Information View OMAS.

### DIFF
--- a/open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test-generator/pom.xml
+++ b/open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test-generator/pom.xml
@@ -22,34 +22,34 @@
 
     <dependencies>
 
-            <dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
-            <version>${slf4j.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
         </dependency>
 
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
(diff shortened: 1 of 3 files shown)


open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/views/ViewHandler.java
```diff
@@ -186,8 +189,9 @@ private EntityDetail addDerivedColumn(EntityDetail tableTypeEntity, String quali
             addRelationship(tableTypeEntity.getGUID(), derivedColumnEntity.getGUID(), Constants.ATTRIBUTE_FOR_SCHEMA);
             addRelationship(derivedColumnEntity.getGUID(), derivedColumn.getSourceColumn().getGuid(), Constants.SCHEMA_QUERY_IMPLEMENTATION);
             addRelationship(derivedColumnEntity.getGUID(), columnTypeEntity.getGUID(), Constants.SCHEMA_ATTRIBUTE_TYPE);
-            addRelationship(derivedColumnEntity.getGUID(), derivedColumn.getSourceColumn().getBusinessTerm().getGuid(), Constants.SEMANTIC_ASSIGNMENT);
-
+            for(BusinessTerm businessTerm : derivedColumn.getSourceColumn().getBusinessTerms()){
```
ruxandraRosu: In a future PR you should consider that in case the column exists, the business terms previously assigned could be different than the assigned business terms from the event.

cong78: Ok thanks! I will mark it as a TODO in the next PR.

PR #1268, opened 2019-07-08
Speed up open types test generation massively 

planetf1:
See #967 for analysis and description

In summary
 * Fixed 'clean' target (typo)
 * Switched test execution to use maven-antrun plugin (flexibility)
 * Corrected launch path to ensure execution is the same whether building one module or all
 * Return failure from test builder & propagate to maven to abort the build if .well.. it fails (before would continue without error) 
 * Add local optimization for entity references to bring 15-90 minute processing down to <1 minute
 * Added logging (may remove some more later... though other build phases log plenty too -- needed to confirm fix is good in all environments)


planetf1:
Note: See #967 - the formal build is currently seeming to abort before these tests being run... 
This is NOT a regression for this fix as prior builds are showing same issue
Will raise new issue to track.

planetf1:
@davidradl Appreciate any comments you may have

planetf1:
Ignore above comment - FVT tests are not currently run in the PR build - hence skipped.
So fix should be good for developers.

PR #1270, opened 2019-07-10
Glossary View OMAS - referenceable entity retrieval, fixed param validation, simplified conversion

lcpopa:


lcpopa:
Recreated, without other contributors commits displayed

PR #1276, opened 2019-07-12
Glossary View OMAS - added navigation on relations specific to spine objects, junits

lcpopa:


planetf1:
Java 11 build fails but looks like an infrastructure problem. re-starting check

planetf1:
recheck

lcpopa:
recheck

PR #1279, opened 2019-07-15
Changed the readme or added default classes into @JsonTypeInfo. And fixed defect in method "isValidClassificationForEntity"

immmus:
Changed the readme since the request in postman did not work without a class. 
You can also just add default values ​​of classes in @JsonTypeInfo and the old query will work. Attached two commits, the choice to make you :)


immmus:
`<dependency>`
    `<groupId>commons-collections</groupId>`
    `<artifactId>commons-collections</artifactId>`
`</dependency>`

schannamallu:
recheck

Issue #1252, opened 2019-07-05
Create info.yaml for egeria


planetf1:
Create info.yaml as requested by linux foundation infrastructure team
see https://github.com/odpi/ci-management/issues/118#issuecomment-507796141

planetf1:
done

PR #1287, opened 2019-07-17
Create info.yaml as per Linux Foundation requirements

planetf1:
As per #1252
@schannamallu :

I have ran against yamllint so should be valid yaml
I don't yet have answers to all the questions for permitted value
some sections are empty where it was unclear what to add - others I've taken a guess
please review/comment on the PR
@mandy-chessell
Please review as above. Also I have sourced maintainers from MAINTAINERS.md - I don't have access to review the github lists. please suggest any corrections needed


planetf1:
Bad PR - pulling in incorrect local test changes

PR #1305, opened 2019-07-19
Skip pmd/spotbugs for open types test

planetf1:
the pmd/spotbugs scan takes too much resource for our generated fvt code, and the build
environment is unable to complete the scan.

This PR configures pmd, spotbugs & owasp scan to skip the generated open types test code

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1923,34 +1923,7 @@
                     </execution>
                 </executions>
             </plugin>
-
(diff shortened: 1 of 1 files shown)


planetf1:
Also corrected a few cases where a plugin was defined with an explicit version. Instead moved the version declaration to `<pluginManagement>'

Issue #1221, opened 2019-06-28
Information view server is using backport-util-concurrent:backport-util-concurrent (very old, superceeded)


planetf1:
Information view server is using a very old library which was intended to support old versions of java.

We find it in org.odpi.egeria:information-view-server

From our license scan:
```
Dependency backport-util-concurrent:backport-util-concurrent:jar:3.1 located at Module org.odpi.egeria:information-view-server:jar:1.1-SNAPSHOT

Dependency backport-util-concurrent:backport-util-concurrent:jar:3.1 located at Module org.odpi.egeria:information-view-spring:jar:1.1-SNAPSHOT

Dependency backport-util-concurrent:backport-util-concurrent:jar:3.1 located at Module org.odpi.egeria:open-metadata-caller-package:pom:1.1-SNAPSHOT

Dependency backport-util-concurrent:backport-util-concurrent:jar:3.1 located at Module org.odpi.egeria:server-chassis-spring:jar:1.1-SNAPSHOT

backport-util-concurrent-3.1.jar located at open-metadata-distribution/open-metadata-assemblies/target/egeria-1.1-SNAPSHOT-distribution.tar.gz/egeria-omag-1.1-SNAPSHOT/server/server-chassis-spring-1.1-SNAPSHOT.jar/BOOT-INF/lib
```
This was created to implement JSR166 which I think went into java 1.5 so we should now be using standard java constructs

planetf1:
cc: @ruxandraRosu

PR #1310, opened 2019-07-22
#1221 - Information view server is using backport-util-concurrent; fix for excluding inactive relationships when building the context

ruxandraRosu:
Information view server is using backport-util-concurrent:backport-util-concurrent (very old, superceeded) #1221


planetf1:
thanks for fixing

PR #1313, opened 2019-07-22
Data Platform OMAS: external source registration

cong78:
This PR is about: 
- Functionality of registering external data platforms as Software Server Capability. 
- Refactoring of Error Codes. 
- Improve some server side codes. 

Any feedback is welcomed. :)

### DIFF
--- a/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/beans/InformationViewAsset.java
+++ b/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/beans/InformationViewAsset.java
@@ -10,7 +10,7 @@ public class InformationViewAsset {
     private EntityDetail endpointProperties;
     private EntityDetail connectionEntity;
     private EntityDetail connectorTypeEntity;
-    private EntityDetail dataStore;
+    private EntityDetail database;
     private EntityDetail informationViewEntity;
     private EntityDetail relationalDbSchemaType;
 
@@ -46,12 +46,12 @@ public class InformationViewAsset {
         this.connectorTypeEntity = connectorTypeEntity;
     }
 
-    public EntityDetail getDataStore() {
-        return dataStore;
+    public EntityDetail getDatabase() {
+        return database;
     }
 
-    public void setDataStore(EntityDetail dataStore) {
-        this.dataStore = dataStore;
+    public void setDatabase(EntityDetail database) {
+        this.database = database;
     }
 
     public EntityDetail getInformationViewEntity() {
--- a/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/handlers/InformationViewAssetHandler.java
+++ b/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/handlers/InformationViewAssetHandler.java
@@ -117,20 +117,20 @@ public class InformationViewAssetHandler implements Callable<InformationViewAsse
                 connectorTypeEntity.getGUID(),
                 new InstanceProperties());
 
-        String qualifiedNameForDataStore = QualifiedNameUtils.buildQualifiedName(qualifiedNameForSoftwareServer, Constants.DATA_STORE, event.getTableSource().getDatabaseSource().getName());
-        InstanceProperties dataStoreProperties = new EntityPropertiesBuilder()
-                .withStringProperty(Constants.QUALIFIED_NAME, qualifiedNameForDataStore)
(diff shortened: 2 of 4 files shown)


open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/handlers/RegistrationHandler.java
```diff
@@ -0,0 +1,105 @@
...
+                QUALIFIED_NAME_PROPERTY_NAME,
+                qualifiedName,
+                methodName);
+
+        EntityDetail retrievedEntity = repositoryHandler.getUniqueEntityByName(
+                userId,
+                qualifiedName,
+                QUALIFIED_NAME_PROPERTY_NAME,
+                properties,
+                SOFTWARE_SERVER_CAPABILITY_TYPE_GUID,
```
ruxandraRosu: you could use repositoryHelper.getTypeByName() to load the typeDef for SoftwareServerCapability. In this way you don't need to duplicate the guid in your code.

cong78: Ah great I will change it. Thx!

open-metadata-implementation/access-services/data-platform/data-platform-spring/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/server/spring/DataPlatformOMASResource.java
```diff
@@ -2,14 +2,54 @@
...
+    /**
+     * Return the unique identifier(GUID) from a software server capability definition.
+     *
+     * @param serverName    name of server instance to call
+     * @param userId        identifier of calling user
+     * @param qualifiedName qualified name of the software server capability
+     * @return unique identified of the software server
+     */
+    @GetMapping(path = "/software-server-capabilities/{qualifiedName}")
+    public GUIDResponse getSoftwareServerCapabilityGuidByQualifiedName(@PathVariable String serverName,
```
ruxandraRosu: Not sure about this, maybe it makes sense to return the softwareServerCapability other properties as well, not just the guid?

cong78: hm yes it makes sense from user perspective. I will change it.

open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/eventprocessor/EventPublisher.java
```diff
@@ -329,7 +331,8 @@ public void processConflictingTypeEvent(String sourceName,
      * @param event to be published
      * @return true/false based on the success of the operation
      */
-    public boolean sendEvent(DataPlatformEventHeader event) {
+    public boolean sendEvent(NewViewEvent event) {
```
ruxandraRosu: It makes sense to keep the parent abstract class as argument, it should work the same for any event type.

open-metadata-implementation/access-services/data-platform/data-platform-api/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/responses/DataPlatformOMASAPIResponse.java
```diff
@@ -24,7 +24,8 @@
 @JsonSubTypes(
         {
                 @JsonSubTypes.Type(value = GUIDResponse.class, name = "GUIDResponse"),
-                @JsonSubTypes.Type(value = VoidResponse.class, name = "VoidResponse")
+                @JsonSubTypes.Type(value = VoidResponse.class, name = "VoidResponse"),
+                @JsonSubTypes.Type(value = RegistrationRequestBody.class, name = "SoftwareServerCapability"),
```
ruxandraRosu: You could separate Response class from Request class. We could use the same naming for the same concept, maybe we should discuss about using softwareServerCapability instead of registration naming for consistency across all OMASs.

open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/eventprocessor/EventPublisher.java
cong78: yes somehow I misunderstood this part. Will change it.

cong78:
In this PR I also made the change on InformatiViewAsset that replaced DataStore to Database to make sure the type of software server is more specific. @ruxandraRosu

PR #1322, opened 2019-07-26
refactoring and fix bug

immmus:
Hello! I propose my own variant of refactoring methods "valueAsObject" and "valueAsString" in predefined classes.

Also, in the ArrayPropertyValue class, these methods did not work correctly due to two errors made, namely “Integer.getInteger (index)” on String return null, not a number, just the “add” method will throw an “IndexOutOfBoundsException” error when trying insert into index > list.size, which is what happened in this case.


immmus:
I propose to return the map in the “Value of an array property” just as elsewhere in order to clearly see the position of the object in the mass, since it is denoted as a key.

PR #1325, opened 2019-07-29
Security Officer OMAS - Processing of new events

danielaotelea:
Security Officer OMAS is able to process new events:
1. Classified Entity Event if the entity classified is a Glossary Term and the classification is Confidentiality. In this case, the schema elements that have a semantic assignment relationship with the BT, will be updated. Security Tag Connector resolves the security tag and the metadata repo is updated. Also, the propagation is done from the table level to assigned columns.
2. Re-Classified Entity Event for Glossary Term + Confidentiality
Schema elements are updated with the new classification. The propagation from table to column is done if the classification on the table level is more restrictive. (If a column has its own classification from a BT assigned directly, the security tag is updated if at the table level is a more restrictive classification, if not, the column keeps its classification)
3. De-Classified Entity Event for Glossary Term + Confidentiality
The security tags is deleted from schema elements (if a column contains a security tag from its own semantic assignment , it keeps the security tag, if the tag comes from the table level is deleted).
Added also: the default security tag if the Semantic Assignment relationship contains a Glossary Term without a Confidentiality classification

### DIFF
--- a/open-metadata-implementation/access-services/glossary-view/glossary-view-api/src/main/java/org/odpi/openmetadata/accessservices/glossaryview/rest/ControlledTerm.java
+++ b/open-metadata-implementation/access-services/glossary-view/glossary-view-api/src/main/java/org/odpi/openmetadata/accessservices/glossaryview/rest/ControlledTerm.java
@@ -1,0 +1,20 @@
+/* SPDX-License-Identifier: Apache 2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.accessservices.glossaryview.rest;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.NONE;
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.PUBLIC_ONLY;
+
+/**
+ * Simple POJO that holds the data of a ControlledTerm
+ */
+@JsonAutoDetect(getterVisibility = PUBLIC_ONLY, setterVisibility = PUBLIC_ONLY, fieldVisibility = NONE)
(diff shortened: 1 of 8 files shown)


open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/security-officer-connectors/security-officer-tag-connector/src/main/java/org/odpi/openmetadata/openconnector/governancedarmonconnectors/securityofficerconnectors/securitytagconnector/SecurityTagConnector.java
```diff
@@ -27,17 +27,32 @@ public SecurityClassification resolveSecurityClassification(SchemaElementEntity
...
 
-        return getDefaultSecurityTags();
+        return securityClassification;
     }
 
     private SecurityClassification getDefaultSecurityTags() {
         SecurityClassification securityClassification = new SecurityClassification();
-        securityClassification.setSecurityLabels(Collections.singletonList("C3"));
-        securityClassification.setSecurityProperties(Collections.singletonMap("user", "SecurityTagConnector"));
+        securityClassification.setSecurityLabels(Collections.singletonList("C99"));
```
cong78: So I assume the C99 is the default security tag?

danielaotelea: C99 is the default security tag. The confidentiality level = 99 is "Secret", by default, if the business term does not have a Confidentiality classification assigned, schema elements are hidden.

cong78: cool thanks!

PR #1332, opened 2019-07-31
Open Lineage Services client: Methods now take enum values instead of…

GuyAckermans:
… strings

### DIFF
--- a/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/eventprocessor/EventPublisher.java
+++ b/open-metadata-implementation/access-services/data-platform/data-platform-server/src/main/java/org/odpi/openmetadata/accessservices/dataplatform/eventprocessor/EventPublisher.java
@@ -23,11 +23,15 @@ import org.slf4j.LoggerFactory;
 public class EventPublisher  extends OMRSInstanceEventProcessor  {
 
     private static final Logger log = LoggerFactory.getLogger(EventPublisher.class);
+    private static final String eventPublisherName = "Data Platform OMAS Event Publisher";
+
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private OpenMetadataTopic dataPlatformOutTopicConnector;
     private OMRSAuditLog auditLog;
 
     public EventPublisher(OpenMetadataTopic dataPlatformOutTopicConnector, OMRSAuditLog auditLog) {
+        super(eventPublisherName);
+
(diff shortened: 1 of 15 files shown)


open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-client/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/client/util/Graphs.java
```diff
@@ -0,0 +1,20 @@
+/* SPDX-License-Identifier: Apache 2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.governanceservers.openlineage.client.util;
+
+public enum Graphs {
+
+    main("main"),
```
bogdan-sava: Because they are constants, the names of an enum type's fields are in uppercase letters.
Please see:
[https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html](url)

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-client/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/client/util/Queries.java
```diff
@@ -0,0 +1,19 @@
+/* SPDX-License-Identifier: Apache 2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.governanceservers.openlineage.client.util;
+
+public enum Queries {
+
+    glossary("glossary"),
```
bogdan-sava: same here.
Please use uppercase letters.

GuyAckermans: Hi Bogdan, 
I made a conscious decision here to deviate from the naming convention because these enum values are used to distinguish between the GET parameters of my endpoint. Having uppercase GET values would be even less ideal I think. See also the GraphServices class. Shall I leave it like this, or would you have a different idea on how to deal with this problem?

bogdan-sava: Uppercase value for http parameters is not an issue.

GuyAckermans:
@bogdan-sava Resolved by converting http parameters to match the capital case enum values

PR #1334, opened 2019-08-01
Security Sync Server is able to sync the policies defined in the External Enforcement tool

danielaotelea:
Ranger Security Connector is able to fetch the policies defined in Apache Ranger Server.
Based on the polling interval that is configurable using the Security Sync config, the Ranger server is asked for potential changes from the last know version of policies till the moment.


open-metadata-implementation/adapters/governance-engines-plugins/gaian-ranger-plugin/pom.xml
```diff
@@ -91,7 +91,7 @@
                 <artifactId>maven-shade-plugin</artifactId>
                 <executions>
                     <execution>
-			<id>assemble></id>
+			            <id>assemble></id>
```
planetf1: bad tab (minor!)

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/security-sync-connectors/ranger-connector/pom.xml
```diff
@@ -55,6 +55,56 @@
             <artifactId>commons-collections4</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.ranger</groupId>
+            <artifactId>ranger-plugins-common</artifactId>
+            <exclusions>
+                <exclusion>
```
planetf1: interesting - nice idea to try and slim things down :-)

danielaotelea: I think it will be better to use the dependencyManagement from Egeria's pom.xml. I will update the PR. I saw you added this management of dependecies after I have created the PR.

PR #1338, opened 2019-08-05
Issue1194

grahamwallis:


grahamwallis:
This PR makes the change to the graph repo such that it always takes a regex and matches it against the full value (string property of an entity or relationship) or classification name.

PR #1339, opened 2019-08-06
Adds helper methods to qualify strings to avoid regex interpretation

cmgrote:
As discussed on today's functionality call, these are helper methods to qualify strings to avoid them being interpreted as regular expressions.

Expectation is the following:

- OMASes wanting to do exact-match searches should call `getQualifiedLiteralString()` to get a string that can then be passed to the `find...` methods and will do an exact-match search
- Repository connectors that want to check whether the passed criteria is a regular expression should call `isLiteral()` from their `find...` methods -- true indicates the entire expression is an exact match, false indicates that at least some portion of it is a regular expression (so if the connector will do partial regular expression handling it needs to handle its own complexity on a `false` response; otherwise can simply short-circuit directly at the point at which a `false` is seen if it should not handle regular expressions at all)
- Repository connectors that cannot pass-down regular expressions to their back-end can call `getUnqualifiedLiteralString()` to "un-escape" the passed criteria to be able to push it down as an exact match (should only be done on the criteria or portion of the criteria that is _not_ a regular expression: ie. when `isLiteral()` returns true or when the other complex handling for partial regex handling has been done.)

### DIFF
--- a/open-metadata-implementation/repository-services/repository-services-apis/src/main/java/org/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/repositoryconnector/OMRSRepositoryHelper.java
+++ b/open-metadata-implementation/repository-services/repository-services-apis/src/main/java/org/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/repositoryconnector/OMRSRepositoryHelper.java
@@ -1075,18 +1075,25 @@ public interface OMRSRepositoryHelper
 
     /**
      * Retrieve an escaped version of the provided string that can be passed to methods that expect regular expressions,
-     * without being interpreted as a regular expression (i.e. the returned string will be interpreted as a literal).
+     * without being interpreted as a regular expression (i.e. the returned string will be interpreted as a literal --
+     * used to find an exact match of the string, irrespective of whether it contains characters that may have special
+     * meanings to regular expressions).
+     *
+     * Note that usage of the string by methods that cannot handle regular expressions should first un-escape the string
+     * using the getUnqualifiedLiteralString helper method.
      *
      * @param s - the string to escape to avoid being interpreted as a regular expression
      * @return string that is interpreted literally rather than as a regular expression
-     * @see #isLiteral(String)
+     * @see #isExactMatchRegex(String)
      * @see #getUnqualifiedLiteralString(String)
      */
-     String getQualifiedLiteralString(String s);
+    String getExactMatchRegex(String s);
 
     /**
-     * Determine whether the provided string should be interpreted literally (true) or as a regular expression (false).
-     * Note that this method relies on the use of the getQualifiedLiteralString helper method having been used to
+     * Indicates whether the provided string should be treated as an exact match (true) or any other regular expression
+     * (false).
+     *
+     * Note that this method relies on the use of the getExactMatchRegex helper method having been used to
      * qualify a string when it should be treated as a literal. That is, this method relies on the presence of the
      * escape sequences used by Java's Pattern.quote() method. The method is not intended to work on all strings in
      * general to arbitrarily detect whether they might be a regular expression or not.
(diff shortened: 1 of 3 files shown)


mandy-chessell:
How does isLiteral() know whether this is a regular expression or not?
If the repository connector removes the escape characters then they are going to return too much?  This is the cause of the original issue.

cmgrote:
> How does isLiteral() know whether this is a regular expression or not?

It is reliant on someone obtaining the literal by using the related helper method referenced in the JavaDoc for the `isLiteral()` method: `getQualifiedLiteralString()`.  This uses Java's native `Pattern.quote()` method to change the entire string into a literal (basically putting a `\Q` at the beginning of the string and a `\E` at the end of the string).  The `isLiteral()` method will therefore look for these presence of these to determine if the string is a literal (they're there) or not (they aren't there).

cmgrote:
There is no means to "detect" whether something is a regex or not -- every string (even without any regex characters) is technically a regex -- so it's up to the caller (OMAS, etc) to define that behaviour based on whether or not it "qualifies" its criteria before calling.

mandy-chessell:
The descriptions of the methods needs more explanation to say that the isLiteral assumes the presence of the escape characters.  For example string `"Drop Foot"` is literal and I would expect isLiteral() to return true but it won't.

I don't think the implementation needs changing - just clarification on the function.

cmgrote:
Brought another question to my mind, though: will OMASes have access to these OMRSRepositoryHelper methods?  Or do we need some other means to expose `getQualifiedLiteralString()` (or eg. make it static in the `OMRSRepositoryContentManager` class)?

cmgrote:
> I don't think the implementation needs changing - just clarification on the function.

New JavaDoc description for `isLiteral()`:

> Determine whether the provided string should be interpreted literally (true) or as a regular expression (false). Note that this method relies on the use of the getQualifiedLiteralString helper method having been used to qualify a string when it should be treated as a literal. That is, this method relies on the presence of the escape sequences used by Java's Pattern.quote() method. The method is not intended to work on all strings in general to arbitrarily detect whether they might be a regular expression or not.

> Primarily a helper method for methods that do not directly handle regular expressions (for those it
should be possible to just directly use the string as-is and it will be correctly interpreted).

cmgrote:
Hmm...  hold this thought -- thinking some other common helper functions for cases like `contains` might be handy as well, so considering some method renames to further help clarify things.

cmgrote:
Ok, thinking this is probably more logical now:

1. OMAS (or other caller) knows what behaviour it wants -- it either qualifies the string using one of the appropriate helper methods (`getExactMatchRegex()`, `getContainsRegex()`, `getStartsWithRegex()`, or `getEndsWithRegex()`), or builds its own custom regex, before calling to the enterprise connector. (We can expand these `getXYZRegex()` helpers if there are other common scenarios we come across.)
1. Enterprise Connector passes the string it receives as-is to the underlying repository connector(s).
1. Each repository connector can use the `isExactMatchRegex()` method to determine whether it is a literal string (should all be able to handle) or the other `isXYZRegex()` methods to check for a "relatively simple" regex that most repositories can likely handle (contains, startswith, endswith), or if none of those that it is a complex regex.
1. If it cannot handle regexes, the repository connector can decide at what point to short-circuit; if it can only handle the "simple" ones, it can use `getUnqualifiedLiteralString()` to remove the escapes that were put into the string by any of the `getXYZRegex()` methods (so it can pass along the result of that to its own search mechanism, even if it doesn't support regexes, after having determined what of its own proprietary wrapping it needs to do to achieve the "contains", "startswith" or "endswith" based on which simple regex was detected by the `isXYZRegex()` methods).

cmgrote:
Hi @mandy-chessell -- so with the "redesign" outlined directly above, hopefully it makes more sense? =)  This will make it much easier to make use of the methods from the perspective of connectors like Atlas and IGC (both of which support a limited set of "simple" regex-like mechanisms, but not full regexes) while still allowing full regexes to be used in general by other repos.

I'm still wondering about OMAS invocation of the helpers: should we have these as static methods within the OMRSRepositoryContentManager?  (I'm assuming the OMASes won't have an instance of an OMRSRepositoryContentManager to use them non-statically?)

mandy-chessell:
@cmgrote 
The repository content manager is private to the OMRS.  An OMAS should not call it directly.

Every OMAS is passed an Enterprise Repository Connector instance.  This connector provides access to the repository helper instance.  The repository helper instance is initialized with a link to the repository content manager instance for this server. 

Static methods are not needed.  We avoid using them to maintain maximum flexibility in the internal server implementation.

PR #1341, opened 2019-08-06
Data engine - update process creation to set the status to draft/active

popa-raluca:
- Create processes with DRAFT status and update them to ACTIVE after all the entities and relationships inside the process payload are created
- Extend RepositoryHandler to allow update of entity status
- Throw NoSchemaAttributeException when a schema attribute can not be retrieved based on the qualifiedName


popa-raluca:
> Only question I have is that I see we're doing parallel processing of the processes that are being sent through... Since within the payload of a given process we may include `PortImplementation` and `PortAlias` objects, it's expected that one process's `PortAlias` may refer to another's `PortImplementation`: if we're loading them all in parallel, how can we ensure that the lowest-level one (with the `PortImplementation`) is fully created before attempting to create the higher-level one (with a `PortAlias` that refers to the `PortImplementation` of the lower-level process)?

Hi Chris, 
For now, I only process the responses in parallel, after the all processes have been created.
The processes are being created in the method createProcesses(String userId, String serverName, List<Process> processes) - which does not create them in parallel yet. I'm working on this part now, I'll do a new pull request these days.

PR #1342, opened 2019-08-06
Updates jackson-databind to latest micro patch level

cmgrote:
To address a few CVEs, which hopefully would not affect us anyway (my limited understanding from past reading was that these are related to gadgets / polymorphism when using default typing, which Jackson only seems to be able to address through blacklisting...)  I'm expecting that we have default typing disabled (?)

Note that this does mean we need to track separate version properties for `jackson-databind` vs the rest, as the micro-patches apply only to this module and not the others (if we tried using the micro-patch level across the board we'll simply end up with Maven errors about non-existent modules).


planetf1:
The fix looks good. To be honest, jackson is rather a pain wrt CVEs. There's constant issues found release after release. As you say to some extent they are mitigated by the actual use we make of then in egeria, but that doesn't stop github nagging, CLM reports failing (which can be marked as acceptable), and of course the risk we misunderstand, and do indeed suffer from the issue (thinking apple, sqllite news just this week)

We'll obviously need to update both versions together, and 2.10.0 is now in preview, but on balance let's work with what we have and integrate this update for now. 

thanks :-)

PR #1346, opened 2019-08-07
Adds error code for any regex that isn't supported by a repository

cmgrote:
So we have a standard error to throw from any repository (connector) that is sent a regular expression it is not able to handle.  I'm guessing best to throw these as `RepositoryErrorException` as well (or is there a different preferred exception to use)?


mandy-chessell:
For FFDC, an error code should only be used in one place unless there is enough context information in the error message to be sure exactly where the error was detected.

Each connector implementation should have their own error code file for exceptions (and audit log messages) that it creates.

This particular message is not particularly well formed in that it does not give any context of the request.  It should also be implemented so that it is only output once - rather than every time a request is made - especially since there is no action to take.  If this message is output on every request where the regex is not supported then audit log will be swamped with thousands of messages that the operator can do nothign about.

It is better that notification of the limitation is brought to light in the conformance test.

cmgrote:
Closing -- to be replaced by repository-specific errors and `FunctionNotSupportedException`

PR #1139, opened 2019-06-17
Add mandatory javadoc & sources for maven central release (master) (NO MERGE YET)

planetf1:
This PR enforces the default javadoc build, excluding fvt (ie for opentypes)
It also adds empty jar files for sources & javadoc as this is mandatory for all artifacts for maven central

See #1110 for more info

NOTE: DO NOT MERGE until the javadoc errors are addressed.
cc: @mandy-chessell

PR #1357, opened 2019-08-09
JavaDoc error corrections

cmgrote:
Seemed to be a number of minor JavaDoc issues that were flagged as errors, and I think may be preventing some of the more recent builds from completing?

Have corrected directly, and I don't think anything has been lost (eg. removed `@throws` for exceptions that were undocumented and were not being thrown, `@param` that referred to non-existent arguments, etc)


planetf1:
Thanks, Once this is merged I'll redo my fix to force execution of javadoc on each build
See https://github.com/odpi/egeria/pull/1139 which also includes additional commentary.

However I wouldn't have thought these issues would cause build problems YET (until I implement 1139) as javadoc is normally only built by the mvn 'site' phase.

As above, I'll update/merge then revisit

PR #1360, opened 2019-08-09
Horizontal Lineage bufferGraph

DimitriosMaimaris:


planetf1:
I notice this is closed - but for info
 - If there's new dependencies they need explaining/justifying, if not currently used
 - Dependencies should be added in DependencyManagement in top level pom, then a dependency in the pom that needs them.
See developer-resources/Dependency-Management.md

Issue #1106, opened 2019-06-05
Update Lucene version


grahamwallis:

grahamwallis:
Lucene is included by Janusgraph. Egeria currently uses two different versions of Janusgraph (the Graph Repository uses 0.3.1 and Open Lineage Services uses 0.3.0). Both these JG versions pull in Lucene 7.0.0. 
This issue proposes that we align our use of Janusgraph at 0.3.1 (for both the Graph Repository and Open Lineage Services), exclude Lucene 7.0.0, and explicitly depend on Lucene 8.1.1.

grahamwallis:
I have asked the JG community whether it will be possible to upgrade to at least Lucene 7.1.0 in JG 0.3.2. Depending on answer and relative timing of releases, we might be avoid needing to explicitly exclude and include Lucene versions.

planetf1:
If we don't directly depend on lucene, only through JG, and they declare a dependency we can likely slip declaring a dependency.
That may not be possible though if we refer to lucene (not checked code). If so, using a 'provided' dependency may be appropriate so at least we only build with our declared version, and use theirs at runtime (this is what we do with gaian/derby)

planetf1:
Ref: CVE-2017-12629

planetf1:
Lucene is pulled in directly via
 * graph-repository-connector (via janusgraph-lucene)
 * open-lineage-services-server

With a further 16 maven artifacts indirectly (ie through the two above components - mostly the graph repo connector)

....

grahamwallis:
Can we close this issue now?

planetf1:
We are still reporting the CVE as using lucene 7.0.0 - see http://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2017-12629. I think we should establish the impact on Egeria, and document any potential exposure/mitigation if we can't realistically update

planetf1:
See https://github.com/odpi/egeria/pull/2772. Dependabot is now suggesting Janus 0.5 which has updated lucene to 7.7.2

planetf1:
See #1106

grahamwallis:
Sonarcloud appears to be clear now (for this issue). @planetf1 - if you agree I think we could now close this issue?

planetf1:
Agreed

PR #1362, opened 2019-08-12
AL OMAS, Open Lineage: Storing Lineage

DimitriosMaimaris:
Functionality for saving Entities and Relationships relating to Horizontal Lineage in bufferGraph.


open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/pom.xml
```diff
@@ -27,6 +27,13 @@
     </properties>
     <dependencies>
 
+
+        <dependency>
+            <groupId>org.quartz-scheduler</groupId>
```
planetf1: This is introducing a significant new dependency. Can this be brought up on the next Egeria call to ensure the core maintainers are happy with the additional dependencyu?

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/pom.xml
```diff
@@ -27,6 +27,13 @@
...
+            <artifactId>quartz</artifactId>
+            <version>2.2.1</version>
```
planetf1: Current version is 2.3.1 . Was there a reason the older version was chosen?

planetf1: Also the group/artifact/version should be added into the '<dependencyManagement>' section in the top level pom.xml, then here we remove the version. This helps to maintain consistency and is documented in the developer guidelines

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/pom.xml
```diff
@@ -78,6 +85,13 @@
             <artifactId>gremlin-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analyzers-common</artifactId>
+            <version>7.0.0</version>
```
planetf1: Current versions are 8.2.x. However we probably need to be consistent with the rest of egeria (and janus) if possible.  See also #1106 cc: @grahamwallis 

If a new dependency is needed - as with previous comment, please ensure it is added to dependencyManagement and the explicit version removed here. Also I would recommend running 'mvn dependency:tree' in this case, as we have transient dependencies coming in from Janus. (let me know if you need any help)

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/mockdata/MockGraphGenerator.java
```diff
@@ -41,8 +41,8 @@ public MockGraphGenerator() {
     /**
      * The parameters for the graph that is to be generated are hardcoded for now.
      * A "flow" constitutes of a path between columns of different tables, connected together by process nodes.
```
GuyAckermans: @code{} should remain

planetf1:
Just a FYI, JavaDoc is now run on every build. Whilst we don't check that all methods are documented (we could!), it is important that the javadoc copiler doesn't find errors. The latest PR build seems to be failing on the javadoc in the lineage modules. You can run mvn javadoc:javadoc just to run this stage.

PR #1364, opened 2019-08-12
update info.yaml as per LF build team request

planetf1:
Updated as per https://github.com/odpi/ci-management/issues/118


info.yaml
```diff
@@ -8,10 +8,10 @@ project_category: 'metadata'
...
     email: 'mandy_chessell@uk.ibm.com'
-    id: 'mandy-chessell'
+    github_id: 'mandy-chessell'
     company: 'IBM'
     timezone: 'Europe/London'
-primary_contact: *odpi_releng_ptl
+primary_contact: *odpi_egeria_ptl
 issue_tracking:
     type: 'github'
     url: https://github.com/odpi/egeria/issues'
```
cmgrote: The url for issue_tracking seems to have a dangling single quote?  Noticed the same under the list of repos (maybe worth scanning for any other spots as well?)

planetf1:
Thanks @cmgrote - corrected (and again ran yamllint)

planetf1:
I've changed the filename to INFO.yaml

Can you elaborate on what you need for the tsc. Technical Steering Group? 
Is it meant to be an id? email? web address?
And what do 'changes' signify (and then what does type/name/link mean?)

I see the schema at https://github.com/lfit/releng-global-jjb/blob/3fba00b0773a59eed76521103afb0a62deb79786/schema/info-schema.yaml but beyond requiring a 'tsc' entry, without docs/info I don't know the context, how the info is used, what is expected . Info on our meetings can be found at https://github.com/odpi/egeria

The TSC itself is the ODPi TSC. More info on this can be found at https://github.com/odpi/tsc

Perhaps you could make the changes that are needed & submit a PR? It is likely more efficient (I can merge my current changes first if it helps?)

PR #1366, opened 2019-08-12
Link updates for IGC connector GH repo rename

cmgrote:
@planetf1 looks like you were right, there were a couple of spots (including in the Helm chart) to include the rename 🤷‍♂


cmgrote:
Merging as only documentation changes + Helm chart for IGC

Issue #1373, opened 2019-08-14
Egeria docker image build failing


planetf1:
See https://jenkins.odpi.org/view/egeria-docker/job/egeria-maven-docker-stage-master/lastBuild/
```
13:15:36 [INFO] Removing intermediate container 1ab36459a9a1
13:15:36 [INFO]  ---> fd02b20f25b6
13:15:36 [INFO] Step 25/34 : RUN groupadd -r egeria -g 1000 && useradd --no-log-init -r -g egeria -u 1000 -d /opt/egeria egeria
13:15:36 [INFO] 
13:15:36 [INFO]  ---> Running in a0371786230f
13:15:37 [INFO] Removing intermediate container a0371786230f
13:15:37 [INFO]  ---> ee20ad7a5d97
13:15:37 [INFO] Step 26/34 : COPY --from=build --chown=egeria:egeria /root/egeria/open-metadata-implementation/server-chassis/server-chassis-spring/target/server-chassis-spring-*.jar /opt/egeria/server-chassis-spring.jar
13:15:37 [INFO] 
13:15:40 [ERROR] When using COPY with more than one source file, the destination must be a directory and end with a /
13:15:40 [INFO] ------------------------------------------------------------------------
13:15:40 [INFO] BUILD FAILURE
13:15:40 [INFO] ------------------------------------------------------------------------
13:15:40 [INFO] Total time: 27:35 min
13:15:40 [INFO] Finished at: 2019-08-14T12:15:40Z
13:15:40 [INFO] ------------------------------------------------------------------------
13:15:40 [ERROR] Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-egeria: Could not build image: When using COPY with more than one source file, the destination must be a directory and end with a / -> [Help 1]
13:15:40 org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.spotify:dockerfile-maven-plugin:1.4.10:build (build-and-tag-latest) on project open-metadata-docker-egeria: Could not build image
```

planetf1:
Egeria itself built ok, but may be caused by recent image split work

planetf1:
Debugging/fixing ...

planetf1:
The cause here is that the build now contains multiple files that match the pattern:
```
root@792e6d9604ad:~/egeria/open-metadata-implementation/server-chassis/server-chassis-spring/target# ls
[...]

PR #1376, opened 2019-08-14
#1373 fix egeria docker build (sources, javadoc need removal)

planetf1:
See #1373 

Fixes build failure due to additional javadoc, sources now in build


cmgrote:
Copying files over and then having multiple run commands means the layers end up left behind anyway, I think?  So we'll only be adding further space to the image size, rather than reducing it...  I'd prefer one of the following options instead:

- Change the build image so that it puts the various jar files into different directories (using `mv`, etc commands there, where we could then use the full regex we may need in the shell to do so); then the runtime just copies from the appropriate directory containing only the single jar it needs
- Change the runtime image to take as an argument the version number (which we should have via the pom anyway, so can pass in, since we tag with that version number as part of the Docker build process); and use this passed-in argument to construct a fully-qualified jar file name

planetf1:
I'm proposing to push this as the build is broken (and we have a LF conf coming up) but happy to then go back and refine at another point.

planetf1:
I opened up a new issue - clearer, different priority.

PR #1421, opened 2019-08-15
# add "/local" in path LocalRepositoryServicesResource

immmus:
Hi,
For the general concept, I suggest adding "/local" in path "Local Repository Services Resource" so that the url matches the general concept - "...repository-services/users/{userId}/resourceName".


immmus:
> This will break backward compatibility

Oh .. I did not think about it. I close it. :)

mandy-chessell:
Perhaps a little more background on this API.  It is used between servers to support federated queries.  Within the cohort there may be servers of different versions of egeria - this is why we need to preserve the interface.

This was the first API supported by the repository services.  If we where adding it today than this suggestion makes sense

immmus:
> Perhaps a little more background on this API. It is used between servers to support federated queries. Within the cohort there may be servers of different versions of egeria - this is why we need to preserve the interface.
> 
> This was the first API supported by the repository services. If we where adding it today than this suggestion makes sense

Yes, Yes I understand this. Thank you!)

Issue #1419, opened 2019-08-15
Lab tutorial notebook - default kafka port for local


planetf1:
Having deployed the 'lab' helm chart into a cloud deployment (IKS) I ran through the tutorials to note any adjustments/parameterization/defaults we need to add

initially this is to capture/document the changes, so that we can then better parameterize/default/document:

Some key info you need about your k8s deployment:
 * Ingress subdomain ie http://XXX.eu-gb.containers.appdomain.cloud
```
➜  charts git:(master) ✗ ibmcloud ks cluster get egeriadev | grep 'Ingress Subdomain'
Ingress Subdomain:              egeriadev.eu-gb.containers.appdomain.cloud
```
 * Information on deployed services (kubectl get services) ie
```
➜  charts git:(master) ✗ kubectl get services
NAME                        TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)             AGE
lab-cp-kafka                ClusterIP   172.21.6.89      <none>        9092/TCP            24m
lab-cp-kafka-headless       ClusterIP   None             <none>        9092/TCP            24m
lab-cp-zookeeper            ClusterIP   172.21.247.233   <none>        2181/TCP            24m
lab-cp-zookeeper-headless   ClusterIP   None             <none>        2888/TCP,3888/TCP   24m
lab-egeria-core-service     NodePort    172.21.188.153   <none>        8080:30080/TCP      24m
lab-egeria-dev-service      NodePort    172.21.38.199    <none>        8080:30082/TCP      24m
lab-egeria-lake-service     NodePort    172.21.146.21    <none>        8080:30081/TCP      24m
lab-jupyter-service         NodePort    172.21.91.206    <none>        8888:30888/TCP      24m
➜  charts git:(master) ✗
```
With these services the important info is the name, and then the second port (after :). All services of type 'NodePort' will be available on the node (the ingress subdomain) at that port number, ie 30888 for the jupyter service

To get started the notebook itself is typically accessed via
http://hostname:30888

Managing services
--
corePlatformURL     = "http://localhost:30080"
dataLakePlatformURL = "http://localhost:30081"
devPlatformURL      = "http://localhost:30082"

instead of the original values of

corePlatformURL     = "http://localhost:8080"
dataLakePlatformURL = "http://localhost:8081"
devPlatformURL      = "http://localhost:8082"

it's worth noting (cc: @mandy-chessell ) that
```
serverConfig=response.json().get('omagserverConfig')
[...]

PR #1422, opened 2019-08-16
#1419 Add environment variables for notebooks running in cloud

planetf1:
Adds environment variable for running tutorials from k8s


planetf1:
Her's an example proposal per #1419.

Other implementations are possible
 - pulling all envs out to a config map
 - keeping host/port seperate
etc.

Will test some more and keen to see if this works for notebook authors.

planetf1:
merging. Any changes needed when we test integration with updated notebooks will be a new PR.

PR #1424, opened 2019-08-16
#1419 Added load of notebooks on startup from github (as done for k8s)

planetf1:
Notebooks are now automatically populated in the Jupyter UI from github for the docker-compose implementation. This matches the approach taken in the helm lab chart


planetf1:
Note - container will run as root (and then will sudo to 'jovyan') - this is needed to allow apt-get to work. The alternative would be to build a new image based on this one, which includes git. However for a local lab environment I think this approach is fine.

PR #1432, opened 2019-08-19
Open lineage: convert graphson to valid json

GuyAckermans:


open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-spring/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/server/spring/OpenLineageResource.java
```diff
@@ -32,15 +32,15 @@
...
             @PathVariable("serverName") String serverName,
             @PathVariable("scope") String scope,
             @PathVariable("lineageQuery") String lineageQuery,
             @PathVariable("graph") String graph,
             @PathVariable("guid") String guid) {
-        return restAPI.initialGraph(serverName, userId, scope, lineageQuery, graph, guid);
+        return restAPI.queryLineage(serverName, userId, scope, lineageQuery, graph, guid);
     }
 
     /**
```
ruxandraRosu: Noticed class inTopicListener starts with lower i, please rename it to start with upper i

GuyAckermans: Fixed, thanks for spotting

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/services/GraphServices.java
```diff
@@ -96,45 +102,78 @@ private String glossary(Graph graph, String guid) {
...
-                        until(inE(edgeLabel).count().is(0)).
-                        repeat(inE(edgeLabel).subgraph("subGraph").outV()).
-                        cap("subGraph").next();
-        return janusGraphToGraphson(subGraph);
+        List<Vertex> sourcesList = g.V().has(GraphConstants.PROPERTY_KEY_ENTITY_GUID, guid).
+                until(inE(edgeLabel).count().is(0)).
+                repeat(inE(edgeLabel).outV()).dedup().toList();
+        Vertex originalQueriedVertex = g.V().has(GraphConstants.PROPERTY_KEY_ENTITY_GUID, guid).next();
+
+        Graph responseGraph = TinkerGraph.open();
```
ruxandraRosu: In this method a new temp graph is created where a subset of nodes/properties/edges are copied. This could be a performance issue as the memory and processor consumption could get high, creating and populating a graph is not trivial. We should investigate for a better way as we only need a subset of nodes and edges, with no manipulation operations on them. This could be done in future PR, this one can be merged as it is from my side.

PR #1457, opened 2019-08-29
Corrects Ansible group_names to address deprecation warning

cmgrote:
Replaces `-` with `_` in group_names to avoid a deprecation warning (`-` is technically an invalid character for a group_name, which up to now has automatically been replaced silently by Ansible -- that behaviour will change in an upcoming Ansible release, so correcting properly).


cmgrote:
Merging, as only changes to the Ansible playbooks that I've been managing...

Issue #1464, opened 2019-09-02
Minor cleanup of build warnings in the in-memory and graph repo connectors


mandy-chessell:
Both connectors were using a deprecated method for parameter validation.  The In-memory repository's graph handling was using unsafe casts.

mandy-chessell:
Also fixed calls to deprecated methods in the Kafka topic connector and added maxPageSize properties to OMRS instance management (means the maxPageSize from the config is now used).

Issue #1463, opened 2019-09-02
Community Profile OMAS Outbound Event Management


mandy-chessell:
This issue covers the support for the Community Profile OMAS sending events relating to changes in personal profiles and user identities.

There are new payload events, unit tests and support for a publisher and listener

mandy-chessell:
This enhancement is still in plan

mandy-chessell:
This is done

Issue #1462, opened 2019-09-02
New AuditLog components for OMAS In and Out topics


mandy-chessell:
When an instance of the audit log is created, it is assigned the component that it it logging on behalf of.  The audit log instance passed to the access service during the initialize method is configured with details of the access service as the component.

This change creates new component descriptions in OMRS for the in and out topic connectors so it is possible to separate issues with events arriving and leaving the server from the REST API work of the access service.

Issue #1461, opened 2019-09-02
Split access service registration from API module


mandy-chessell:
The server-side of an access service needs to register their admin class and other details with the admin services at OMAG Server Platform start up.  This is then used by the admin services to validate the configuration of individual admin services, set up default configuration and start up/shutdown the service within a server on request.

The code for registration is currently in admin-services-api.  This module should be usable both in the client and the server.  However, the common methods that I recently added to the AccessServiceAdmin base class includes calls to server-side functions.  This means the API module is no longer suitable for the client.

This issue covers the work to pull the registration code into its own module and change the current access services to inclide this dependency in its server-side POM.

PR #1465, opened 2019-09-02
Support for #1461, #1462, #1463, #1464 

mandy-chessell:
The event management is not yet complete


mandy-chessell:
Thanks Nigel - I will remove the scope from the Avro dependency

mandy-chessell:
recheck

Issue #1500, opened 2019-09-06
Building egeria within IntelliJ can fail on new install


planetf1:
Scenario
 - new install of intellij
 - new import of egeria project
 - mvn is configured as 3.6.2 in intellij
 - java is configured as 11.0.4 (openjdk hotspot) in intellij
 - mvn clean install at command line uses the same jvm/mvn and runs fine
 - using the intellij maven plugin and typing 'mvn clean install' fails with
```
Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:jar (attach-javadocs) on project open-connector-framework: MavenReportException: Error while generating Javadoc: Unable to find javadoc command: The environment variable JAVA_HOME is not correctly set.
```

This is because the environment within intellij is a little different to the cli

The following fragment should fix for the IDE without breaking other usage - added in the configuration for the javadoc plugin
```
<javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>

```
This is more noticeable now javadoc is part of the standard lifecycle (from install)

If a global JAVA_HOME is setup, the UI ay inherit and find javadoc, however since IntelliJ supports switching between JDKs as part of the project definition this could end up using the wrong version (or in this case not find it). java.home however is set correctly.. so the above fix provides for consistent results.

planetf1:
After some research
 - It is tricky to set the javadoc executable correctly. This is because 'java.home' gets set to the java environment /executing/ maven itself, not necessarily the compiler. Whilst in my IDE it was an SDK, in the build environment it was a JRE - and in any case could be the wrong version.
 - all intellij does is call maven. The relevant java environment set for the shell will prevail - NOT the version set in the intellij gui
 - this change was indeed introduced by the enabling of javadoc
 - it likely will only affect those trying to coordinate multiple environments through a tool like 'jenv' (my case)
 - if using jenv, the following commands will address the issue:

```
➜  egeria git:(master) java -version
openjdk version "11.0.4" 2019-07-16
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.4+11)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.4+11, mixed mode)
➜  egeria git:(master) jenv global 11.0
➜  egeria git:(master) echo $JAVA_HOME

➜  egeria git:(master) jenv enable-plugin maven
maven plugin activated
➜  egeria git:(master) jenv enable-plugin export
You may restart your session to activate jenv export plugin echo export plugin activated
[...]

PR #1501, opened 2019-09-06
#1500 fix javadoc build

planetf1:
See #1500
Fixes javadoc build in some environments


planetf1:
This works with jdk, but fails on build machine where maven *itself* is running using jre
Withdrawing PR to investigate this other scenario.

Issue #1512, opened 2019-09-10
OMAG servers startup list configuration 


bogdan-sava:
OMAG server platform suppose to have a list of server that starts on spring-boot application starts.
Default value for the list will be empty, and can be configured in java command, separate application file, env variable (spring-boot supported).
If the server have no configuration document, the error message will be displayed.
If server starts successfully the message will be shown in console.

bogdan-sava:
Current functionality is not affected.

ie. If a list with 2 servers is provided, one with existent configuration and one without, the message will be:

`
OMAG Server **'omas'** SUCCESS start , with message: Wed Sep 11 11:24:51 EEST 2019 omas is running the following services: [Open Metadata Repository Services (OMRS), Connected Asset Services, Asset Owner OMAS, Subject Area OMAS, Governance Program OMAS, Asset Lineage OMAS, Glossary View OMAS, Security Officer OMAS, Asset Consumer OMAS, Asset Catalog OMAS, Community Profile OMAS, Discovery Engine OMAS, Data Engine OMAS, Governance Engine OMAS, Information View OMAS, Data Platform OMAS]
OMAG Server **'test'** ERROR while startup, with error message: OMAG-ADMIN-400-018 OMAG server test has been called with a configuration document that has no services configured

Wed Sep 11 11:24:56 EEST 2019 OMAG server platform ready for more configurations`

PR #1513, opened 2019-09-10
OMAG servers startup list configuration #1512

bogdan-sava:


mandy-chessell:
I think it is useful to have this approach for environments where a platform's servers are fixed.  I am not sure that the platform should keep running if it can not start one of the servers listed in the application properties particularly related to configuration documents or userId..

mandy-chessell:
@bogdan-sava this is not quite the same issue.  If the server throws an exception from the instance startup message it is likely to be a permanent error - for example, the userId is not permitted to perform the action or there is a problem with the configuration document.

The failures once the server is running are more likely to be temporary and can be fixed by managing the resources.

In a production environment, the first type of error should never occur.  If the instance methods are called then the response says what is wrong.  When the servers are started from the property there is no way to return the response so the platform should fail.

bogdan-sava:
platform shuts down means spring boot application shutdown or just deactivateTemporarily all the servers ?

mandy-chessell:
Spring boot application should exit with a non-zero return code.  Ideally if it can deactivateTemporarily all of the running servers then that is a bonus.  No point in only shutting down the other servers as does not help the discoverability of the problem.

bogdan-sava:
delayed by deactivateTemporary issue #1587

bogdan-sava:
recheck

bogdan-sava:
recheck

Issue #1499, opened 2019-09-06
remove (empty) server configuration UI


planetf1:
We have a server-configuration-user-interface directly under user-interfaces. No source is contained, but we do have a module defined in the pom. It gets built into a no-op jar as part of the build process.

Since we've had no code here ever I would suggest we entirely remove this component as it just adds clutter and potentially confusion

bogdan-sava:
README says: The server Configuration User Interface provides the administration user
interface for an open metadata server.
If I remember right, it was designed to be a UI tool for rest configurations that we currently do with postman. 
On the other hand, it can be a view on the UI, accessible if the user has proper  role.
@mandy-chessell what do you think?

planetf1:
A change in directory will break the egeria container build as referenced in /Users/jonesn/src/egeria/open-metadata-resources/open-metadata-deployment/docker/egeria

I intend to change the build to be more resiliant (by using our packaged distro) but until then, if a PR is created I suggest we do not merge until the associated change has been made & tested in the Dockerfile

There are also assemblies in open-metadata-distribution/open-metadata-assemblies/src/main/assemblies . These should include the UI, but do not. My intent was to fix as part of the work mentioned above.

But the change sounds sensible

planetf1:
this has been done as part of ui refactoring

PR #1530, opened 2019-09-12
#1499 remove server-configuration-user-interface module

bogdan-sava:
Signed-off-by: bogdan-sava <bogdan.sava@gmail.com>


bogdan-sava:
yes. next PR, if this one goes to master.

planetf1:
I think it's a good idea to split out too -- as there will be changes affecting packaging / containers too and we should merge into master close in time or will break UI in containers

PR #1538, opened 2019-09-13
Use authorized user in rest requests

ruxandraRosu:


open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/java/org/odpi/openmetadata/userinterface/accessservices/api/OpenLineageController.java
```diff
@@ -91,9 +97,10 @@ public String getMockGraph(String userId){
     }
 
     @RequestMapping(method = RequestMethod.GET, value = "/entities/{guid}/source-and-destination")
-    public Map<String, Object> sourceAndDestinationLineage(String userId, @PathVariable("guid") String guid, @RequestParam View view){
+    public Map<String, Object> sourceAndDestinationLineage(@PathVariable("guid") String guid, @RequestParam View view){
```
mandy-chessell: I may be misunderstanding this.  The userId that goes in the security context is the NPA of user interface application.  The userId that flows in the URL is the end user's URL.  Is this what is happening here?  It looks like the end user's userId has been removed.

bogdan-sava: It just looks like that. :) 
Nope. This is replacing the hardcoded userId="demo" with the real  end-user/persona userId in the URLs.

PR #1547, opened 2019-09-13
Asset Lineage OMAS: New workflow for lineage events

DimitriosMaimaris:
After the last workshop changes on the flows has been discussed and this is initial implementation.


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/CommonHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private static final String GUID_PARAMETER = "guid";
+    private String serviceName;
+    private String serverName;
+    private RepositoryHandler repositoryHandler;
+    private OMRSRepositoryHelper repositoryHelper;
+    private InvalidParameterHandler invalidParameterHandler;
+
+    /**
+     * Construct the discovery engine configuration handler caching the objects
```
GuyAckermans: Correct javadoc

Issue #821, opened 2019-03-06
Enable docker images to be built from local content (esp. for egeria container)


planetf1:
Enable a developer working in their dev environment (macos, linux, windows) to test changes in a cloud based k8s service

For example 
 - build egeria under intellij on macos
 - Deploy & test on IBM cloud

scenarios I've felt the need for this in
 - working on a low power system (ie microsoft surface) when travelling
 - deploying full stack / larger images than local machine allows
 - testing cloud specific function integration (ie external kafka, ingress etc)

For example 
 - build egeria under intellij on macos
 - Deploy & test on IBM cloud

See also #810 which allows for local containers to directly use locally built artifacts

planetf1:
I have a reworked egeria dockerfile which builds from the latest maven repo version

By default this uses 1.1-SNAPSHOT & the nexus public repo (where our build snapshots go) but these are arguments which can be modified by the user.

This cuts the docker build time massively - to just over 30s for the build phase which was taking around 30 minutes. More importantly it means that
a) The artifacts used in the container environment are identical to that we build on the official build
b) We test the distribution archives that are intended for other consumers too.

For now I am using the open-metadata-assemblies maven artifact. This contains ALL of our .gz archives. The only one containing the user interface is the *distribution.tar.gz

This works but future changes might include
 * Considering whether we use one or more maven artifacts
 * Considering whether the .gz s should be pushed anywhere other than maven central
 * considering the set of archives contained within the packaging jar - for example a simpler 'runtime' archive.

So will leave this issue open for that consideration.

Now testing the new build approach....

planetf1:
The docker build fails following this change....

The reason is that the order the plugins executed in the pom is
 - Build the assemblies jar
 - package the files into .gz

This is a reasonable default as usually one would be building java code into a jar, then creating an external zip archive for download

However in our case we are trying the opposite - to take our assemblies, and package them into a special packaging jar for distribution via maven central.

In order to achieve this I have re-configured the maven assembly plugin to execute in the 'prepare-package' phase rather than 'package. Since the jar build is linked to the 'package' lifecycle phase this will correct the order. I have then added the 'default-jar' build of the jar which will take the archives and compose them into the module jar.

It likely worked locally due to not working in a from-0 build environment and testing steps in a unit test manner.

planetf1:
So we now have
```
[...]

PR #1588, opened 2019-09-19
Fix docker build (out of sequence jar/gz build)

planetf1:
See #821

Corrects the build order to
 -> package up assemblies into .gz
 -> include in jar


planetf1:
Merging due to build failure

PR #1591, opened 2019-09-19
Add web binders to handle enums in RequestParam/Body

ruxandraRosu:


open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-spring/pom.xml
```diff
@@ -41,4 +46,26 @@
...
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <goals>
+                            <goal>analyze-only</goal>
+                        </goals>
+                        <configuration>
+                            <ignoredUnusedDeclaredDependencies combine.children="append">
+                                <ignoredUnusedDeclaredDependency>org.springframework:spring-context</ignoredUnusedDeclaredDependency>
```
planetf1: I agree with this.

I think technically the use of spring-context should be defined as 'runtime' scope, as it's only needed then for injection, and there's no compile time dependency in this case (may apply to other uses of spring injection)

That being said, the validation done on poms doesn't distinguish for runtime only dependencies without any explicit exclusion - which you've added :-)

PR #1605, opened 2019-09-20
Relocates data-engine-proxy-connector module to fix #1599

### LINKED ISSUE #1599: Data Engine Proxy Connector module should be under consuming component
The data engine proxy connector interface and base classes are currently under the `adapters` module.  The adapters module is for specific implementations of connectors.  The connector API and base classes should belong to the module that is defining the interface of the connector(s) it supports (`data-engine-proxy-services` in this case).

cmgrote:


mandy-chessell:
Thank you :)

PR #1608, opened 2019-09-20
Azure Pipelines CI

bramwelt:
Adds Azure Pipeline builds for PRs, Merges, and Docker.

The workflows defined the Azure Pipelines currently match the Jenkins
workflows:

 * PRs are build and tested
 * Merges are build, tested, and published to the snapshot repo
 * Docker container are build every time changes are merged under the
   docker directory, and rebuilt every day

With one main difference being artifacts are published to Artifactory
(odpi.jfrog.io) instead of Nexus.

### DIFF
--- a/open-metadata-distribution/open-metadata-assemblies/pom.xml
+++ b/open-metadata-distribution/open-metadata-assemblies/pom.xml
@@ -26,6 +26,9 @@
 
     <build>
         <plugins>
+            <!-- This runs the assembly code to build our distribution archives 
+                 Run in an earlier phase than normal (prepare package) so it
+                 gets done before the build of the jar - which will contain the .gz -->
             <plugin>
                 <artifactId>maven-assembly-plugin</artifactId>
                 <executions>
@@ -54,6 +57,28 @@
                     <skip>false</skip>
                 </configuration>
             </plugin>
+            <!-- override the default files included in source archive
+                 we do not want the .gz files we build to appear -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>attach-sources</id>
+                        <phase>verify</phase>
+                        <goals>
+                            <goal>jar-no-fork</goal>
+                        </goals>
+                        <configuration>
+                            <excludes>
+                                <exclude>*.gz</exclude>
+                            </excludes>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <!-- run the jar archiving to build our distribution module 
+                 for maven repo -->
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-jar-plugin</artifactId>
@@ -64,21 +89,17 @@
                         <goals>
                             <goal>jar</goal>
                         </goals>
+                        <configuration>
+                            <!-- this isn't a real jar - we want an archive from other output files -->
+                            <classesDirectory>target</classesDirectory>
+                            <forceCreation>true</forceCreation>
+                            <includes>
+                                <include>*.gz</include>
+                            </includes>
+                        </configuration>
                     </execution>
                 </executions>
             </plugin>
         </plugins>
-        <resources>
-            <resource>
-                <directory>src/main/resources</directory>
-            </resource>
-            <resource>
-                 <filtering>false</filtering>
-                 <directory>target</directory>
-                 <includes>
-                     <include>*.gz</include>
-                 </includes>
-            </resource>
-        </resources>
     </build>
 </project>
--- a/open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
+++ b/open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
@@ -15,6 +15,7 @@ ENV repo ${repo:-"https://nexus.odpi.org/content/groups/public/"}
 ADD https://nexus.odpi.org/content/groups/public/org/odpi/egeria/egeria/${version}/ index.html
 
 WORKDIR /opt/egeria
+RUN mkdir /opt/egeria/binaries
 
 RUN mvn org.apache.maven.plugins:maven-dependency-plugin:2.5.1:get -DremoteRepositories=$repo \
     -Dartifact=org.odpi.egeria:open-metadata-assemblies:${version}:jar -Ddest=egeria.jar && \
@@ -46,7 +47,6 @@ RUN apk --no-cache add bash shadow && \
 
 COPY --from=build --chown=egeria:egeria /opt/egeria/binaries/server-chassis-spring-*.jar /opt/egeria/server-chassis-spring.jar
 COPY --from=build --chown=egeria:egeria /opt/egeria/binaries/access-services-user-interface-*.jar /opt/egeria/access-services-user-interface.jar
-COPY --from=build --chown=egeria:egeria /opt/egeria/git.* /opt/egeria/details/
 COPY --chown=egeria:egeria dist/entrypoint.sh /entrypoint.sh
 
 # Expose port 8080 (default) for client access, and allow for 5005 being used for remote java debug


.azure-pipelines/docker.yml
```diff
@@ -0,0 +1,63 @@
...
+        - master
+
+stages:
+  - stage: Docker
+    jobs:
+      - job: Maven
+        pool:
+          vmImage: 'ubuntu-latest'
+        strategy:
+          matrix:
```
planetf1: In order to improve performance we added some base images under 'build-images'. These are building code from other repos (atlas and ranger) without much customization (that's in the top level images). I'd expect us to only build infrequently as we're using <released> code. I'm just wondering if we can have a manual trigger for those - though an alternative is direct posting to the docker repo (which I've done before)

.azure-pipelines/docker.yml
```diff
@@ -0,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the ODPi Egeria project.
+---
+trigger:
+  paths:
+    include:
+      - open-metadata-resources/open-metadata-deployment/docker
```
planetf1: so I presume we could break down the pipeline and trigger each independently? The egeria one will change the most, the rest are less frequent? This could be a future step? - doesn't inhibit making this first step

.azure-pipelines/docker.yml
```diff
@@ -0,0 +1,63 @@
...
+              mavenPomFile: $(POM_FILE)
+              javaHomeOption: 'JDKVersion'
+              jdkVersionOption: '1.8'
+              jdkArchitectureOption: 'x64'
+              goals: 'clean install'
+              options: >-
+                  --batch-mode
+                  -e
+                  -Ddocker
+                  -Ddocker.repo=odpipoc
```
planetf1: so we'll flip this once we commit to azure as the master

.azure-pipelines/docker.yml
```diff
@@ -0,0 +1,63 @@
...
+    jobs:
+      - job: Maven
+        pool:
+          vmImage: 'ubuntu-latest'
+        strategy:
+          matrix:
+            apache-atlas:
+              POM_FILE: 'open-metadata-resources/open-metadata-deployment/docker/apache-atlas/pom.xml'
+            egeria:
+              POM_FILE: 'open-metadata-resources/open-metadata-deployment/docker/egeria/pom.xml'
```
planetf1: The current egeria docker build is broken. I'm fixing currently, but it's strategy is to make use of our maven artifacts. currently the Dockerfile refers to the odpi registry as the default - we'll need it to point to wherever the right url is for snapshots in jfrog. The repo is a parameter to the dockerfile, but I've not embraced that in the pom yet... I'll raise a seperate issue to correct this after I've fixed the egeria docker build. (The reason for the change was to avoid re-extracting from git and running a 40-60 min process when we could just use maven?)

.azure-pipelines/merge.yml
```diff
@@ -0,0 +1,55 @@
...
+              targetResolveSnapshotRepo: 'egeria-snapshot'
+              artifactoryDeployService: 'ODPiArtifactory'
+              targetDeployReleaseRepo: 'egeria-staging'
+              targetDeploySnapshotRepo: 'egeria-snapshot'
+              collectBuildInfo: true
+              buildName: 'Egeria'
+              buildNumber: '$(Build.BuildNumber)'
+              includeEnvVars: true
+              goals: 'clean install'
+              options: >-
```
planetf1: These days I'm using something like
```
mvn clean install -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss,SSS
```
to get timestamps, though perhaps we see that via pipelines anyway (need to check)

pom.xml
```diff
@@ -2225,6 +2226,21 @@
...
+            <distributionManagement>
+                <repository>
+                    <id>central</id>
+                    <name>odpi-artifactory-primary-0-staging</name>
+                    <url>https://odpi.jfrog.io/odpi/egeria-staging</url>
+                </repository>
+                <snapshotRepository>
+                    <id>snapshots</id>
+                    <name>odpi-artifactory-primary-0-snapshots</name>
+                    <url>https://odpi.jfrog.io/odpi/egeria-snapshot</url>
```
planetf1: Ah got it - that's the snapshot URL to use

.azure-pipelines/verify.yml
```diff
@@ -0,0 +1,44 @@
...
+              options: >-
+                  --batch-mode
+                  -e
+                  -DskipFVT
+                  -Dstream=$(System.PullRequest.TargetBranch)
+                  -DaltDeploymentRepository=staging::default::file:$(Agent.BuildDirectory)/local-repo
+                  -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+                  -Dmaven.source.skip=true
+          - publish: target
+            artifact: target-$(Build.BuildId)_Java-$(JAVA_VERSION)
```
planetf1: with us running java 8 and 11, does this mean they'll go to a different location in jfrog (never used) - will the maven artifacts still be named the same? For now whilst we'll build with 8 and 11, the official/final version we put on maven central etc needs to be from 8.

.azure-pipelines/merge.yml
```diff
@@ -0,0 +1,55 @@
...
+                  -e
+                  -Dmerge
+                  -s $(Agent.HomeDirectory)/.m2/settings.xml
+                  -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+                  -P artifactory
+          - task: ArtifactoryPublishBuildInfo@1
+            inputs:
+              artifactoryService: ODPiArtifactory
+              buildName: 'Egeria'
+          - publish: target
```
planetf1: So far we have only published maven artifacts as part of our build process.
The build process is generating some .gz archives which represent an archive of the runtime code or source - but hasn't yet packaged them up into maven (something I'm adding).

The reason for this As I mentioned above is to use within the docker build for deployment.

However an alternative could be to post other artifacts to jfrog but I'm unsure what standards might apply there? Open to any new suggestions though we should take on as a new issue, the PR is fine for now!

.azure-pipelines/docker.yml
bramwelt: yep!

.azure-pipelines/docker.yml
bramwelt: Yeah it wouldn't be much work to pull those out to a separate pipeline that doesn't have the 'scheduled' trigger as well. Then they can be triggered manually from the UI.

.azure-pipelines/docker.yml
bramwelt: We could - The builds could also be configured to depend on the others being created first. Ex: publish of base container triggers building of 3 dependent containers, etc.

.azure-pipelines/docker.yml
bramwelt: Right there's no need to rebuild the world when all you want to do is package things into a Docker container. It sounds like what you'd like is to have these builds trigger when the new snapshot (and release) jars are created, so that they're hopefully always up to date with the current builds and it's easy for users to pull down and run them. I'm pretty sure that can be done through AZP. It's probably best to meet and write down these future improvements so we're all clear on how the containers are setup and relate to the code/artifacts, as I'm still trying to figure the big picture. :)

.azure-pipelines/merge.yml
bramwelt: The pipelines already give timestamps for each log line, and how long each step takes - I think they may be hidden by default though. If you click the three dots on a build and 'Toggle timestamps' they should show up.

.azure-pipelines/verify.yml
bramwelt: Currently only the Java 8 builds are pushed to Artifactory - I've don't know yet how to manage multiple JDK version builds of projects there. These 'publish' steps will actually store them in Azure Pipelines though, the 'ArtifactoryMaven' task in the merge.yml file is what publishes them to Aritfactory.

As an example, you'd see them as target-1234_Java-1.8 under a location like: https://dev.azure.com/ODPi/Egeria/_build/results?buildId=186&view=artifacts

.azure-pipelines/merge.yml
bramwelt: Artifactory seems pretty flexible, and you can publish any binary to a 'generic' repo - basically an S3 bucket or site repo in Nexus. I think I saw something about skiping source jars? Are these gzips different then those? Or maybe there's some drawback to generating source jars I don't know.

.azure-pipelines/docker.yml
planetf1: So for the current 'phase' the proposal is we leave those running in jenkins? Do we need to open a specific issue to track implementing them in Azure pipelines?

.azure-pipelines/docker.yml
planetf1: ok I'll close this part of the thread, and will open a new issue if/when we want to explore improving docker build once we have the basics in place

.azure-pipelines/docker.yml
planetf1: ok

.azure-pipelines/docker.yml
planetf1: ok thanks for the info. will close the thread here and open up a new discussion - probably once we have the basics in place

.azure-pipelines/merge.yml
planetf1: ok

.azure-pipelines/verify.yml
planetf1: ok that's fine for now. will open up new issue when we're ready to commit to 11 only

.azure-pipelines/merge.yml
planetf1: the .gz files represent everything you might need to copy to a new machine to run the code for example
 - the core server runtime
 - the runtime & sources
 - client only libraries.

They are a little distinct from the maven artifacts which are more geared to other java applications using our code (client or more)

As a twist we also have an assembly maven artifact which itself includes the .gz for distribution . (well it would if it worked). I'm not sure how useful it is, but it's been used in lieu of having anywhere else we can post them.

So I think the question here for the project will be to consider where we wish/can publish too, and at that point we'll have to come back to refine the pipelines.

I'm not sure of the source jar reference?

planetf1:
Any last comments on this? I plan to merge tomorrow if not & we can track improvements via new issues as for now this will run in parallel

.azure-pipelines/docker.yml
bramwelt: Yeah it'd be helpful to have the issues written out for how you envision the docker builds working given it seems to be a bit more involved then just running maven.

.azure-pipelines/merge.yml
bramwelt: > generating some .gz archives which represent an archive of the runtime code or source
I thought these might be the same as `org.odpi.egeria-...-source.jar` files, but they're obviously a bit more than that.

Assuming those files (the .gz) wouldn't be used outside the container, we could generate and store them in the pipeline artifacts as a step before the docker builds. Or even if they were used outside, we could setup a 'generic' Artifactory repo to store them in - whichever makes the most sense.

planetf1: ok thanks. will see how the first stage goes and then open up the docker refinement discussion.

Issue #1116, opened 2019-06-11
Can't build, missing com.sleepycat.je-7.4.5 jar in Maven Central


patduin:
Error on `mvn install` (After clean git clone of the project).

> Failed to execute goal on project graph-repository-connector: Could not resolve dependencies for project org.odpi.egeria:graph-repository-connector:jar:1.1-SNAPSHOT: Could not find artifact com.sleepycat:je:jar:7.4.5 in central

Maven central has a couple of other versions: https://search.maven.org/search?q=g:com.sleepycat%20AND%20a:je

planetf1:
Intriguing. I see only 18.3.12, 18.1.11, 18.3.1, 5.0.73, 4.0.92 too 

I've checked our code dependency analyzer and it reports we are using 18.1.11

However if I look at the output of mvn dependency:tree I see that org.janusgraph:janusgraph-berkeleyje:jar is referring to 7.4.5 . That is a component from the janusgraph community and the dependency is expressed at https://repo1.maven.org/maven2/org/janusgraph/janusgraph-berkeleyje/0.3.1/janusgraph-berkeleyje-0.3.1.pom

This is also the latest version of janusgraph

So I'm not quite sure why our 'clean' builds work - it looks like we may resolve to a much later version

Are you building from full egeria source or making use of the maven artifacts from central
We just pushed some egeria 1.0 artifacts. but they'd been held up for a while on build related issues. 

In case it helps, As an alternative https://nexus.odpi.org/content/groups/public/ is the repository we are pushing snapshots too

planetf1:
As you mentioned a clean clone I'm trying that here again with a cleared out ~/.m2 to see if I can repro the issue.

patduin:
Yeah I worked around it by adding:
```
<dependency>
  <groupId>com.sleepycat</groupId>
  <artifactId>je</artifactId>
  <version>18.3.12</version>
</dependency>
```

to the dependency management, that made the build succeed was just a bit worried that it might break something when running it all. I'm completely new to Egeria.
If I see any issues I'll downgrade it to 18.1.11. Thanks for quick reply.

planetf1:
Hi again,
 I was able to build 'master' from a new clone & with an entirely clean .m2 (cache). I did this under linux (ubuntu 19.04 x86_64). I saw no issues with dependencies

In my log I saw:
```
[...]

PR #1609, opened 2019-09-20
Add Dependency on later version of Sleepycat

bramwelt:
The current version of janusgraph-berkeleyje (a dependency of
JanusGraph) relies on an outdated version of com.sleepycat.je that while
available from Oracle's Maven repository[1], it is no longer replicated
to Maven Central.

Thus when building Egeria without a cached version of sleepycat (or
pulling from a Maven repository without the cached version), some users
will continue to experience failures when building Egeria. Specifing a
greater version of sleepycat appears to bypass this failure and allow
the build to succeed.

Resolves: #1116

[1] https://download.oracle.com/maven/com/sleepycat/je/7.4.5/je-7.4.5.pom


bramwelt:
As an aside: I pulled sleepycat 7.4.5 was from Oracle and uploaded it to Artifactory in order for the builds to pass -  I don't think I understand enough of how maven dependency resolution works to explain why builds work someplaces and not others.

planetf1:
This causes us problems due to our use of JanusGraph. See also https://github.com/JanusGraph/janusgraph/issues/1621

At the time the earlier issue was raised we hadn't fully embraced the use of dependencyManagement to aid with transitive dependencies, but we have now - meaning we can  make this kind of change.

The only question is whether there's any concern/impact on the graph. I would expect not as berkeleydb is pretty stable now - and only used for a simple environment. 

Any comments @grahamwallis @mandy-chessell @GuyAckermans

bramwelt:
> The only question is whether there's any concern/impact on the graph. I would expect not as berkeleydb is pretty stable now - and only used for a simple environment.

I think the correct approach here is really to bump the JanusGraph dependency assuming it pulls in the newer version of sleepycat. But I'm not well versed in Egeria to know how much that version change might affect things. :thinking:

planetf1:
recheck

PR #1614, opened 2019-09-23
Data Engine OMAS - remove external mappings between processes

popa-raluca:


cmgrote:
We won't need these anymore for the DataStage proxy, but I'm wondering whether there will not be any use for them...

Just to confirm: when we have lineage that crosses system / tool boundaries, the assumption is that they will write to some consistent layer (like a database, a file, an event bus, etc), and it is based on one tool writing to this and another reading from it that we get the end-to-end lineage across all the tools, right?  So in that case we actually don't need a `LineageMapping` object (that exists outside the tools' own domains)?

popa-raluca:
This was my assumption, but the external lineage mappings section is just removed from the endpoint for creating processes.
There is still an endpoint for adding lineage mappings in Data Engine OMAS, that could be used for lineage across tools.

```
POST {{omas-url}}/servers/{{server-id-omas}}/open-metadata/access-services/data-engine/users/{{user-id}}/lineage-mappings
```

PR #1623, opened 2019-09-25
New labs, Asset Consumer API call and OMRS Clients

mandy-chessell:
This PR includes some new python notebooks and readmes in answer to the extremely helpful feedback we had last week from the community call and Chris N.

Also there are 2 new clients for OMRS Repository services for REX/TEX: one for local repository one for enterprise repository. @grahamwallis 

Finally, Asset Consumer's API now clearly distinguishes between exact match queries and fuzzy match searches.


planetf1:
Nte this will only work in the containerized environments once egeria images  are rebuilt/refresh. For now this is https://jenkins.odpi.org/view/egeria-docker/job/egeria-maven-docker-stage-master/ but see issue above. Should be ready by about 1200 UTC

PR #1627, opened 2019-09-25
egeria#863 - Fix issue with egeria consumer properties not being used

jnhagelberg:
This commit was supposed to be included in the previous PR I created, but due to an oversight it did not make it into that PR.  The issue is that in `KafkaOpenMetadataTopicConnector`, `consumerEgeriaProperties` is currently not being populated, so the default value consumer egeria properties are always being used.  There are handful of options that can be included in there.  These are enumerated in `KafkaOpenMetadataEventConsumerProperty`


jnhagelberg:
Thanks Nigel!

PR #1634, opened 2019-09-26
Tutorial updates

cmgrote:
Have tried to simplify the documentation around the entry point for running the tutorials, whilst retaining all of the additional flexible options for doing so.  Probably simplest to "preview" from here in my own fork (before merging): https://github.com/cmgrote/egeria/tree/tutorials/open-metadata-resources/open-metadata-tutorials

Mainly:
- Created a new "Beginner's Guide" that goes through all the technologies involved in 1-2 quick sentences
- (First breaks these into categories to give an initial filter to folks who may not be interested in the dev / build side of things, for example)
- Has the first link for the overall tutorials be about running the labs (not necessary to build and install Egeria just to run the labs with Docker Compose -- likely the simplest way to do it in general as it's just 1 tool to download and install)
- Tried to reduce the verbosity of most of the pages to just get people to the outcome (not discuss all of the options / what's happening behind-the-scenes -- for me we should leave that to the tutorial content of the Notebooks themselves)

I also updated the initial `read-me-first.ipynb` to directly link to follow-on labs, and to simplify the code (mainly introducing a function to do the repetitive task of status checking the servers rather than copy / pasted code blocks that, aside from taking lots of vertical real estate, might give people the wrong impression about Egeria's code itself 😉)

Happy to go through the other `.ipynb`s and make similar cleanups to compress their vertical sprawl a bit, but wanted to see what initial reactions were to these changes first...

One other thought, which I didn't change:

- Should we have so much space given to the simple icons for the users?  I'm all for diagrams, and perhaps we need more of these within the Notebooks to illustrate some of the ideas / concepts, but I'm not sure that these massive user icons are adding much to the understand-ability of the tutorials (and may be distracting given how much space on the screen they take up 😉)


cong78:
Very nice work!

cmgrote:
One other thought, to make it even easier to consume the Docker Compose material: it might be worth packaging the YAML and `notebook-start.d` contents into a very simple `.zip` archive that we could place under our GitHub releases (or GitHub's new "package" mechanism, if we enable that on the repo)?  That way we can provide a direct link to it and people don't need to be familiar with cloning the repo or downloading files individually in order to get started with Docker Compose...

(I'm assuming the YAML and such will be relatively static, since they pull the more dynamic stuff -- like the code-containing containers, Notebooks themselves, etc -- as part of their startup.)

cmgrote:
And actually the Notebooks do show a rendered preview in GitHub (see: https://help.github.com/en/articles/working-with-jupyter-notebook-files-on-github), just that the preview renders much more slowly than other file types...

Based on this, it raises the question: should we actually commit the Notebooks with a "clean run" of output (rather than in an un-run state as we do currently)?  This would mean anyone previewing the Notebooks on GitHub itself -- and in fact when loading them up to Jupyter for the first time -- should be able to see what a "clean run" looks like (ie. what output to expect).  The downside is that it means seeing the Notebook for the first time in Jupyter I think you'd also see the output, making it harder to distinguish as a user whether you've actually run it yourself or are just seeing the pre-defined "clean run" outputs...

mandy-chessell:
I disagree about removing the icons of the users - they help people remember the characters and their images appear in other diagrams.

planetf1:
* I would prefer not to have any prior run info in the notebooks when a user opens them in jupyter - it could easily cause confusion
* I do get the point about seeing if we could make a read-only copy available - and it's nice to see github can render. 
* To do both we'd need to check both copies in, strip the responses before adding to notebook ser er (but I think it remains confusing for git source users), or run the notebook on a test server. The latter I'd like to do anyway... but it depends on how we then publish any results. Maybe could be a side effect
* There are also other public servers for hosting notebooks, but they depend on back-ends in our case to run
* Maybe we could do something better for a 'release' rather than master.

That brings me to your other point - the zip file for the 'bits' needed for docker compose. I did wonder about this, but it's extra admin for us whilst working on master. It I think does have merit for 'released' versions of the notebook - ie we create a distribution package that we 'publish' somewhere (possibly github releases) that indeed just contains the files you reference. I wouldn't do that for master though.

creating a PR/investigating the jupyter team's images and proposing a new feature to 'prime' the notebook server with a git project/part thereof could be a neat feature .. 

Finally I've been working with a few other teams/people on using the notebook and have made some changes to the build process to assist(ie speeding it up immensely by reusing the main build artifact -- albeit a large clunky one... refactoring still needed) .. I'm considering 'baking in' the tutorials as part of the docker image build, since we've had some issues where the code (from the docker build)  is out of sync with the notebook (from git) but that has side effects too. I'll raise an issue on that one.

PR #1637, opened 2019-09-27
Further tutorial tweaks

cmgrote:
- Retains images, but wraps text around to reduce blank whitespace
- Changes "catalogues" to "catalogs" in various spots
- Adds functions to server startup lab


cmgrote:
Agreed — intention is precisely to make the Notebooks clearer by making the code more concise and self-explanatory rather than trying to make it more efficient.  Hence any functions are only 3-4 lines long and given meaningful names that allow those that may not follow the guts of things like http requests to be able to “abstract” those away in their minds via the functions.

Issue #1630, opened 2019-09-26
Trigger update of Egeria docker container after master merge


planetf1:
We are continually improving our jupyter notebooks & using them to demonstrate how to use our egeria API. In some cases those notebooks are dependent on new APIs such as https://github.com/odpi/egeria/pull/1623 . In other cases changes can break the notebook/server interaction.

In order to narrow the gap it would be useful to kick off the Egeria docker build after a master merge - though it's a little more subtle.

The docker build requires an artifact from maven central (old) and artifactory (new - #1619 ), so:
 - the merge master build will create this artifact
 - then we run the egeria (only) docker build to consume that artifact and produce the container

Recently the time to build the container has gone down from 30-60 mins to 3-5 minutes as per #821 so this becomes viable

One point of note - when notebooks start they pull the latest notebooks from git. This could still need to a discrepancy, so it's not perfect, though if using 'Always' as pull policy (or using docker-compose pull) that will refresh the image and bring things back in sync ish

The proposal here is we look at this automatic triggering. I suggest focussing on Azure pipelines only 
cc: @mandy-chessell @bramwelt

planetf1:
Note if we wanted to keep the docker images and notebooks consistent, we could bake-in the notebooks to the images... (and maybe provide the old behaviour as an option ...)

cmgrote:
This could be a good area to make use of the Artifactory API that we've used in eg. the connector image initialization for k8s, instead of pulling from Git directly -- assuming that we can package up the notebooks into their own assembly / artifact as part of the Maven build process?

That way we could use a "version" environment variable in the Dockerfile / compose / k8s scripts (with a default value of the latest snapshot), which if overridden would ensure that the package pulled from Artifactory is a released package rather than the latest snapshot.

(That is: all of the Dockerfile / compose / k8s scripts would replace their `git clone ...` with the `curl ...` from the Artifactory API using the version-based location for the artifact.)

planetf1:
#2579 has addressed this by adding the build of our core containers (mostly egeria itself) into the standard merge build

Issue #1619, opened 2019-09-24
Update egeria docker build to use jfrog for maven once switching 100% to Azure pipelines


planetf1:
The current jenkins builds publish maven artifacts to nexus.odpi.org
This will change to jfrog artifactory when we commit our official builds to come from Azure pipelines

So we should not make this change yet but need to be prepared to do at the same time as the build switchover

PR #1638, opened 2019-09-27
#1619 Point to jfrog for docker builds

planetf1:
FYI @bramwelt - these are changes for our docker builds
#1630 should be dependent on this
@cmgrote - may be associated changes in the connector repos

Am opening this for review and for build team, but suggest not merging until they are ready - from my perspective the azure builds are functioning very well now (and quickly) 


planetf1:
Note that I suspect the atlas/ranger builds do not require this as they are base images with no egeria dependencies, but since these builds are rarely built, nor automated so far I have made the url change only

cmgrote:
> from my perspective the azure builds are functioning very well now (and quickly)

Agreed!  Their output is also far nicer (and more read-able) 😄

planetf1:
Since we now have both builds running in parallel, nothing should break if either your changes or the one here are merged. It's just us making that call about the switch.

planetf1:
Super - I can confirm the docker build went fine after the change as expected - https://dev.azure.com/odpi/Egeria/_build/results?buildId=257

PR #1653, opened 2019-10-02
Added badges :-)

jmertic:


jmertic:
@cmgrote I do agree, though you could argue it's a good forcing function for keeping these tools up to date :-)

README.md
```diff
@@ -3,6 +3,11 @@
 
 ![Egeria Logo](assets/img/ODPi_Egeria_Logo_color.png)
 
+[![GitHub](https://img.shields.io/github/license/odpi/egeria)](LICENSE.md)
+[![CII Best Practices](https://bestpractices.coreinfrastructure.org/projects/3044/badge)](https://bestpractices.coreinfrastructure.org/projects/3044)
+[![Azure](https://dev.azure.com/odpi/egeria/_apis/build/status/odpi.egeria)](https://dev.azure.com/odpi/Egeria/_build)
```
planetf1: I wonder if rather than azure we say 'Build' or 'Build pipelines'

We could also add a link to maven artifacts - 'snapshots' https://odpi.jfrog.io/odpi/egeria-snapshot/ & releases at maven central https://mvnrepository.com/artifact/org.odpi.egeria

README.md
```diff
@@ -3,6 +3,11 @@
...
+[![Quality Gate Status](https://sonarcloud.io/api/project_badges/measure?project=odpi_egeria&metric=alert_status)](https://sonarcloud.io/dashboard?id=odpi_egeria)
+
```
planetf1: I see chris commented on this already. Obviously we define the gate, I think we all know we have more work to do here. I would be ok about including it though - we should be transparent, and welcome help!

mandy-chessell:
recheck

Issue #1568, opened 2019-09-16
Ui tenancy introduce UI config file and admin call for UI  


davidradl:
initial content to only contain the mappings to the back end. this initial piece of code will not effect current operations. The file config will be made active in a later issue

davidradl:
This includes adding the config store connector

mandy-chessell:
It would be better to use the term "configuration document".   This is partly to be consistent with the same concept in the OMAG Server Platform and partly because the storage destination of the configuration document is determined by the configuration store connector.  Using a file is merely the behaviour of the default connector.  Other connectors may use a database or another type of data platform.

PR #1662, opened 2019-10-06
#1568 initial check in of ui admin calls

davidradl:

This check in creates a file with the mappings to the backend server serverNames. It is not active. The platform services checks are not present. There is a working config request in the readme associated with the top level ui-admin project. Parameter validation has been added with messages and junits.

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/pom.xml
+++ b/open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/pom.xml
@@ -25,6 +25,7 @@
     <packaging>pom</packaging>
     <modules>
         <module>configuration-file-store-connector</module>
+        <module>configuration-encrypted-file-store-connector</module>
     </modules>
 
     <dependencies>
--- a/.gitignore
+++ b/.gitignore
@@ -75,6 +75,7 @@ ui.server.*.config
 
 *.registrystore
 *.testlab.results
+keystore_*/**
 
 # other files
 .DS_Store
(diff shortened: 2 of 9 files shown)


open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/pom.xml
```diff
@@ -38,6 +38,11 @@
 	<dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
+    </dependency>
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>ui-admin-services-api</artifactId>
+            <version>1.1-SNAPSHOT</version>
```
planetf1: ROgue version here (should be in top level pom)
 - though looks like PR is closed, so just FYI...

open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
```diff
@@ -249,8 +249,11 @@
             <artifactId>open-lineage-services-api</artifactId>
             <version>${open-metadata.version}</version>
         </dependency>
-
-
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>ui-admin-services-spring</artifactId>
+            <version>${open-metadata.version}</version>
```
planetf1: another rogue version - use top level pom dependencyManagement

PR #1672, opened 2019-10-07
Updates to schemas and models

mandy-chessell:
This pull request includes the updates to the schema and model open metadata types that were agreed in last week's meeting.

The common OCF asset handles and schema handlers have been updated.

It also restructures the open metadata archive capability and fixes a couple of bugs I found in the notebooks when testing the above.

### DIFF
--- a/.azure-pipelines/docker.yml
+++ b/.azure-pipelines/docker.yml
@@ -40,13 +40,13 @@ stages:
           - task: DownloadSecureFile@1
             name: dockerConfig
             inputs:
-              secureFile: egeria-docker.json
+              secureFile: odpi-docker.json
           - script: mv $(dockerConfig.secureFilePath) ${HOME}/.docker/config.json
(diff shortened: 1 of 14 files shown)


open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test/pom.xml
```diff
@@ -42,7 +42,7 @@
         </dependency>
         <dependency>
                <groupId>org.odpi.egeria</groupId>
-               <artifactId>repository-services-implementation</artifactId>
+               <artifactId>open-metadata-types</artifactId>
                <version>1.1-SNAPSHOT</version>
```
planetf1: technically this version shouldn't be here (though it wasn't a change in this PR). I will do a search/scan to fix up any hanging over or re-introduced and slipped through the net

PR #1677, opened 2019-10-07
making config store connectors generics

bogdan-sava:


bogdan-sava:
This make the both connectors generic so they can be reused for the  UI 
They are now independent of the OMAGServerConfig
@davidradl please use this connectors for UI

bogdan-sava:
@cmgrote Thanks! Yes, I did testing for it.
Functionality is the same, meaning can saves and retrieves any serializable including OMAGServerConfig. 
Usage is bit different, meaning that being generic need to specify the class as parameter on the retrieve.

mandy-chessell:
This is not a suitable approach because we need to ensure bakward compatiblity with existing connector.

Issue #1602, opened 2019-09-19
Open Lineage Services calling JanusGraph and pals directly


mandy-chessell:
The `open-lineage-services-server` module has dependencies on:
- janusgraph-core
- gremlin-core
- tinkergraph-gremlin
- lucene-analyzers-common

All of these external technologies should be called from connectors that have generic APIs to allow other technologies to be swapped in by changing the connector implementation.

The connector based architecture we have in Egeria is critical to ensuring our main libraries only have the minimal dependencies.  It then loads only the connectors it needs from the Connection objects found in the configuration.

DimitriosMaimaris:
There is an open pull request for the fix of the issue @mandy-chessell  can you please review it.

PR #1678, opened 2019-10-07
#1602 [Priority] Connector for Open Lineage Services

DimitriosMaimaris:
Issue: Open Lineage Services calling JanusGraph and pals directly #1602

Moving all the dependencies from Open Lineage Server to the Open Lineage Connectors.


open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
```diff
@@ -0,0 +1,107 @@
...
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <goals>
+                            <goal>analyze-only</goal>
+                        </goals>
+                        <configuration>
+                            <ignoredUnusedDeclaredDependencies combine.children="append">
+                                <!-- Runtime dependencies for janusgraph-->
+                                <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-cql:*</ignoredUnusedDeclaredDependency>
```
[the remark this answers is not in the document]
planetf1: I may need some help understanding JanusGraph here.

I see that at compile time we are dependent on janusgraph-core and gremlin-core (as per https://docs.janusgraph.org/connecting/java/ ). They are fine (and not excluded). 

At what point are the additional runtime dependencies needed? These are the ones in the exclusion (not marked as 'runtime').  I presume janusgraph-cql, es, cassandra come from the required backend that's been chosen for a *specific* configuration of janus - this is dependent on what the org wants for their janusgraph config -- not a requirement of the connector.

As such I'm wondering if any implementation of janusgraph that we provide should itself be in a different maven artifact - not the connector, and so those dependencies should not be expressed here. Additionally as they are custom it's likely down to the org to configure the libraries appropriately, not us.

Unfortunately I can't easily test my theories without more knowledge, and access to cassandra, but I think we are not adequately seperating the concerns here, and are pulling in more dependencies than required.

I assume the commons logging/codec, snappy, antlr are then needed by the above

Can you clarify any further?

@grahamwallis

planetf1:
See some review comments above - I have some concerns prior to merging

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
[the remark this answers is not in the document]
DimitriosMaimaris: I updated the dependencies.

pom.xml
```diff
@@ -163,6 +163,11 @@
         <cassandra.version>3.5.0</cassandra.version>
         <guava.version>20.0</guava.version>
         <tink.version>1.3.0-rc1</tink.version>
+        <commons-logging.version>1.1.3</commons-logging.version>
+        <commons-codec.version>1.10</commons-codec.version>
+        <snappy-java.version>1.1.7.2</snappy-java.version>
+        <antlr-runtime.version>3.5.2</antlr-runtime.version>
+        <joda-time.version>2.10.1</joda-time.version>
```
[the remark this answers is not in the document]
planetf1: I expect dependabot will update this once active in any case....

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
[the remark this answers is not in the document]
planetf1: I see we still have janusgraph es/cql, along with commons-logging & commons-codec

At build time the connector is not dependent on these, so it seems as if these additional dependencies are used at runtime by being found on the classpath.

I am trying to understand (not knowing janus well!) whether this is
 * Essential for anyone using Janus graph in egeria 
OR
 * Is a particular choice of backends (be it for logging or database) for Janus that is specific to your individual deployment option.

If it is the latter I wonder if these do not belong in the connector, but should instead be contributed by the environment, or through an additional packaging construct (like the way that for the server chassis we pop connectors into lib - but do not declare them as dependencies/include their jars)

@grahamwallis can you add anything on this?

I'm just trying to ensure we keep a clear separation

pom.xml
```diff
@@ -163,6 +163,11 @@
         <cassandra.version>3.5.0</cassandra.version>
         <guava.version>20.0</guava.version>
         <tink.version>1.3.0-rc1</tink.version>
+        <commons-logging.version>1.1.3</commons-logging.version>
+        <commons-codec.version>1.10</commons-codec.version>
```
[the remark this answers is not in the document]
planetf1: done

pom.xml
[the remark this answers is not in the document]
planetf1: Having looked at the janusgraph docs, I see we go with a default of berkeley (which makes sense), so is it correct that by adding the janusgraph components here we are adding in the required runtime jars so that we can do config such as 
```
storage.backend=cql
storage.hostname=localhost

index.search.backend=elasticsearch
index.search.hostname=100.100.101.1, 100.100.101.2
index.search.elasticsearch.client-only=true
```

I'm presuming We run Janusgraph locally , so the connector is more than a 'connector' .. its effectively the janus implementation too.

So on that basis I can see an argument for including these jars in as dependencies as they are tightly associated with janus and open up some additional configuration options - in this case for cassandra. Does this make sense to those reading? Anyone? ;-)

How about the apache components - can you explain how they are needed at runtime?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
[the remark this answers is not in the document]
planetf1: Having looked at the janusgraph docs, I see we go with a default of berkeley (which makes sense), so is it correct that by adding the janusgraph components here we are adding in the required runtime jars so that we can do config such as

storage.backend=cql
storage.hostname=localhost

index.search.backend=elasticsearch
index.search.hostname=100.100.101.1, 100.100.101.2
index.search.elasticsearch.client-only=true
I'm presuming We run Janusgraph locally , so the connector is more than a 'connector' .. its effectively the janus implementation too.

So on that basis I can see an argument for including these jars in as dependencies as they are tightly associated with janus and open up some additional configuration options - in this case for cassandra. Does this make sense to those reading? Anyone? ;-)

How about the apache components - can you explain how they are needed at runtime?

grahamwallis:
I don't seem to be able to reply above - but can do so here....we could *either* include the jars in the lib and only pick them up if someone configures a connector to use a graoh repo with cassandra/es as backends - in fact I prefer this if it is possible - or we could include them in our packages - does it not equate to the same thing really - i.e. that we don't load the classes unless they are used?

open-metadata-implementation/adapters/open-connectors/connector-configuration-factory/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/ConnectorConfigurationFactory.java
```diff
@@ -888,7 +888,34 @@ public Connection getDataPlatformConnection (String              serverName,
         return connection;
     }
 
+    /**
+     * Return the connection.  This is using the OpenLineageConnector.
```
mandy-chessell: What is this a connector for?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>open-lineage-connectors</artifactId>
+        <groupId>org.odpi.egeria</groupId>
+        <version>1.1-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>janus-connector</artifactId>
```
mandy-chessell: Add name and description for the artifact

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerOperationalServices.java
```diff
@@ -468,14 +468,13 @@ public SuccessMessageResponse activateWithSuppliedConfig(String           userId
...
                         configuration.getLocalServerUserId(),
-                        configuration.getLocalServerURL(),
-                        configuration.getMaxPageSize());
+                        configuration.getLocalServerURL());
                 instance.setOpenLineageOperationalServices(openLineageOperationalServices);
                 openLineageOperationalServices.initialize(openLineageConfig,
                         operationalRepositoryServices.getAuditLog(GovernanceServicesDescription.OPEN_LINEAGE_SERVICES.getServiceCode(),
-                                                                  GovernanceServicesDescription.OPEN_LINEAGE_SERVICES.getServiceName(),
-                                                                  GovernanceServicesDescription.OPEN_LINEAGE_SERVICES.getServiceDescription(),
-                                                                  GovernanceServicesDescription.OPEN_LINEAGE_SERVICES.getServiceWiki()));
```
mandy-chessell: Correct changes to layout

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
```diff
@@ -0,0 +1,83 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>open-connector-framework</artifactId>
+        </dependency>
+
+        <dependency>
```
mandy-chessell: Where is the dependency on the open lineage services api module where the interface for this connector is located?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/GraphStore.java
```diff
@@ -0,0 +1,8 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector;
+
+public interface GraphStore {
```
mandy-chessell: This interface should be located in the open lineage services api module.  It is an interface that other graph stores need to implement if JanusGraph is to be replaced.

It would help if it had a more meaningful name - such as HistoricalOpenLineageGraphStore.

In addition, to allow you to make changes to the API over time, you should have a base connector class and a base connector provider class in the same package.  Then the JanusGraph implementation would extend the base class.

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/GraphStore.java
```diff
@@ -0,0 +1,8 @@
...
+
+    void addEntity();
```
mandy-chessell: Javadoc?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/IndexingFactory.java
```diff
@@ -13,7 +13,7 @@
 
 import java.time.temporal.ChronoUnit;
 
-import static org.odpi.openmetadata.governanceservers.openlineage.util.GraphConstants.corePropertyTypes;
+import static org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.utils.GraphConstants.corePropertyTypes;
 
 public class IndexingFactory {
```
mandy-chessell: Javadoc?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/JanusConnector.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource;
+import org.janusgraph.core.JanusGraph;
+import org.odpi.openmetadata.frameworks.connectors.ConnectorBase;
+import org.odpi.openmetadata.frameworks.connectors.properties.ConnectionProperties;
+import org.odpi.openmetadata.frameworks.connectors.properties.EndpointProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.JanusFactory.openBufferGraph;
+
```
mandy-chessell: Javadoc

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/JanusConnectorProvider.java
```diff
@@ -0,0 +1,33 @@
+package org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector;
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+import org.odpi.openmetadata.frameworks.connectors.ConnectorProviderBase;
+import org.odpi.openmetadata.frameworks.connectors.properties.beans.ConnectorType;
+
+public class JanusConnectorProvider extends ConnectorProviderBase {
```
mandy-chessell: This should extend the graph store base provider class

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/JanusConnector.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Initialize the connector.
+     *
+     * @param connectorInstanceId  - unique id for the connector instance - useful for messages etc
+     * @param connectionProperties - POJO for the configuration used to create the connector.
+     */
+    @Override
+    public void initialize(String connectorInstanceId, ConnectionProperties connectionProperties) {
+
```
mandy-chessell: call super();

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/JanusConnectorProvider.java
```diff
@@ -0,0 +1,33 @@
...
+    static final String connectorTypeGUID = "e2f657d6-e5bd-11e9-81b4-2a2ae2dbcce4";
+    static final String connectorTypeName = "Janus Graph Connector";
+    static final String connectorTypeDescription = "Connector supports storing and retrieving entities for lineage from Janus Graph.";
+
+    /**
+     * Constructor used to initialize the ConnectorProviderBase with the Java class name of the specific
+     * registry store implementation.
+     */
+    public JanusConnectorProvider() {
+        Class connectorClass = JanusConnector.class;
```
mandy-chessell: call super();

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/GraphStore.java
DimitriosMaimaris: Ok I am trying to understand this so I want to ask. Does the BaseConnector class in Open lineage will have any implementation or just overriding the methods , I assume it will not so the methods will be empty right?. So then why can't I build a generic connector under connectors that will fetch the connector that is needed based on the configuration.

mandy-chessell: The connector interface and base class belong to the "framework module" - that is the module that the connector implementation needs to plug in to.  In this case the open lineage services is the "framework module" and so it specifies the API that it needs connectors to implement in order to support its lineage store.  The base class allows changes to be made to this API without needing to change any existing connector implementations (the base class provides the default implementation of any new methods).

The base class typcally begins reasonably empty - it may have useful utilities defined that are useful to all connector implementations.

If the API and the base class are in the open lineage services API module then the open lineage services have no dependencies on any connector implementation.  The connector implementations can be swapped in and out without any impact on the open lineage services module.

Each connector implement is then dependent on the open lineage services API module.  It extends the connector base class.

The open lineage services does not load the connector implementation.  It uses the **ConnectorBroker** to load the connector implementation requested in the Connection object for the connector.  This Connection object is stored in the configuration document for the open lineage services.

If the API and base class of the connector is in the same module as the janus graph connector then the open lineage services are dependent on a connector implementation.  This is the wrong way around and defeats the purpose of using connectors.

We need to use connectors because different vendors may need to change the implementation of our stores, event busses etc so that Eggeria will run on their platform.

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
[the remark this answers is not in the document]
DimitriosMaimaris: Based on the backend and the indexing you want to use the dependencies should be loaded dynamically correct. By default we can have Berkley for example. So it is needed to pass the dependencies based on configurations but for now I did not do that so I can speed up the development process since a lot of work is based on the code changes. But you are correct this should be taken into consideration. The Cassandra configuration in my case will be past when you configure the open lineage services. The Apache libraries are needed for janusgraph-es or janusgraph-cql I do not remember which one otherwise the build fails.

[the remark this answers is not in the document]
mandy-chessell: These properties need to be part of the configurationProperties located in the Connection object for the graph store connector.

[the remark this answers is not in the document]
DimitriosMaimaris: What I did now is added the properties on the config like this.

`{
	"class": "OpenLineageConfig",
	"openLineageDescription": "Open Lineage Service is used for the storage and querying of lineage",
	"inTopicName": "omas.omas.assetlineage.outTopic",
	"openLineageWiki": "wiki URL",
	"openLineageProvider": "org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.JanusConnectorProvider",
	"bufferGraphConfig":{
		"storageBackend": "cql",
		"storageHostname": "127.0.0.1",
		"storageCqlKeyspace": "janusgraph",
		"indexSearchBackend": "elasticsearch",
		"indexSearchHostname": "127.0.0.1:9200"
	}
}`

but I did not do anything for how to load the dependencies based on the implementation someone need to use. How should I do this? S. Now I hardcoded the Cassandra and elastic dependency which is not correct.

Issue #1574, opened 2019-09-17
Enable 'dependabot' for Egeria repos


grahamwallis:
Please see https://dependabot.com/#how-it-works

planetf1:
It certainly looks quite interesting. Also been acquired by github.

We do our scans independently with Nexus IQ - but then have the work to make necessary changes. This could make those changes easier, though the implication will be constantly pushing us to the latest (I presume). Are we happy with that? Probably in master we should be generally, though components with interdependencies like spring/spring-boot/spring-data may be more consternation

grahamwallis:
I agree that using dependabot might introduce pressure to constantly update but we would be at liberty to reject a suggested update. It could create more monitoring work for us - but overall the workload should be less as the PRs would have been prepared automatically.

There is also the issue that we've seen elsewhere, that an 'update' to a dependency can actually introduce a security regression. My guess we would need to continue with our current  build scanning, and certainly should not instantly switch to dependabot - but maybe introduce it experimentally to see what it comes up with?

planetf1:
Its a github provided option now - so it just needs someone with admin permission to enable if we're happy.

I'd like to discuss some security related topics at the next face to face egeria session, so suggest we talk through this as one of the tools to aid compliance

planetf1:
Agreed at face2face 2 Oct that we will deploy dependabot 

I will create an initial config file as per https://dependabot.com/docs/config-file/ and try it against a private repo, and then contact the build team to enable dependabot against the main egeria repo.

planetf1:
FYI: https://github.com/dependabot/feedback/issues/705

planetf1:
For a preview, I've enabled docker/javascript/java against my fork -> https://github.com/planetf1/egeria/pulls

planetf1:
Initial Observations
 * Config is fairly easy for Java/maven
 * For Javascript/.docker EVERY package.json/Dockerfile needs adding manually (once) - so config needs updating if new ones added (see issue raised above)
 * the proposed PRs look decent - and respects our use of properties->dependencymanagement-dependency
 * Docker dependencies on java err to Java 13 - I think not, but there are controls to decline the PR/stop checking
 * The 'LATEST' version classification isn't supported - not really an issue - I just happened to use for security checks. we can let dependabot do it itself
 * We def should do manual review
 * Default limits will restrict dependabot to a handful of changes at once - probably a good idea though I reconfigured my repo so I could get a better idea of the scope of its proposals
[...]

PR #1691, opened 2019-10-09
#1574 Add dependabot config to Egeria

planetf1:
See #1574 for info

Adds dependabot config for
 - Docker
 - UI (2 x package.json)
 - Docker (multiple) 


.dependabot/config.yml
```diff
@@ -0,0 +1,80 @@
...
+  # Explicit entries also needed for dockerfile
+  - package_manager: "docker"
+    directory: "/open-metadata-resources/open-metadata-deployment/docker/omrs-monitor"
+    update_schedule: "daily"
+    default_reviewers:
+      - "grahamwallis"
+      - "planetf1"
+  - package_manager: "docker"
+    directory: "/open-metadata-resources/open-metadata-deployment/docker/build-images/apache/ranger"
+    update_schedule: "daily"
```
cmgrote: I think for these build images we could probably move to weekly rather than daily updates -- the other docker containers you could leave at daily (as they're the only ones people should actually be "running" anyway, I think any within-a-week vulnerabilities on the build images should be negligible as we don't really run on top of them but only copy jar files out of them from the build process)...

planetf1:
The support team have enabled github dependency checker - it's based on dependabot so should have the same effect. Will now merge this change

.dependabot/config.yml
planetf1: Apologies. I meant to reply prior to merge. My bad. I've gone with daily for now as I need to get rapid feedback on the process working as I don't have easy access to dependabot itself. 

I'll suggest we go with a new issue to refine once we have some basic reports coming in.

PR #1700, opened 2019-10-09
Updated samples

cmgrote:
- Updated OMRS cohort payloads from loading Coco Pharma samples into IGC (including revised GUIDs and simplified types)
- (Pending) Revised PostgreSQL loads to be case-insensitive


cmgrote:
Strange -- worked that time... 🤷‍♂

PR #1706, opened 2019-10-10
Switch Artifact Names to use BuildNumber

### LINKED ISSUE #1703: Azure pipeline verify build failure - artifact already exists
@bramwelt  We have seen a build failure on the Java 11 build at https://dev.azure.com/odpi/Egeria/_build/results?buildId=358&view=results

This was for PR #1700 (@cmgrote ) but the error doesn't appear to be in the Java code, rather we get this in 'Publish Pipeline Artifact'
```
#[error]Artifact target-358_Java-1.11 already exists for build 358.
```

bramwelt:
It's still unclear what caused this issue: a build failed with the error
"Artifact .. already exists for build $(Build.BuildId)".

The documentation for AZP is not clear how BuildId is set or calculated,
but we can have more confidence that the artifact names will be unique
when using BuildName which might stop this from occuring in the future.

Resolves #1703

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,80 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Egeria project.
+
+version: 1
+
+update_configs:
+
+  # Main java project - will cover the whole tree
+  - package_manager: "java:maven"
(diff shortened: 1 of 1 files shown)


bramwelt:
Yep; Thanks!

PR #1707, opened 2019-10-10
Run Nexus IQ Scans Weekly and on Merges

bramwelt:
Adds a job to run Nexus IQ scans each time code is merged, so that users
can investigate the reports and take action.


planetf1:
@bramwelt I note that the CLM pipelines are failing with this merge

PR #1736, opened 2019-10-14
Fix dependabot configuration for UI components

planetf1:
See #1724 - I had made an error including package.json in the definition (dependabot only needs the pathname, and assumes this filename)


planetf1:
Merging - intent already agreed, fix is typo, and impact is only to dependabot checks

planetf1:
recheck

Issue #1606, opened 2019-09-20
ViewGeneratorConnector should be under consuming component


mandy-chessell:
The Interface and base classes for the ViewGeneratorConnector and Provider should be under the consuming component `virtualization-services-api`.  This is so that the virtualizer governance server does not have dependencies on Derby.

For clarity, the DerbyConnector should probably be called the ViewGeneratorDerbyConnector. This is to distinguish it from a derby connector that would be used by an application to acccess data in Apache Derby.

mandy-chessell:
In addition, the `getVirtualizationSolutionConnection()` method in the `ConnectorConfigurationFactory` has hard-coded package names for the derby connector.
It should be independent of the connector implementation with the specifics passed in on the parameters.

innerTide:
Hi Mandy, I'm working on this issue. I've created a new package viewgenerator at [virtualization-services-api](https://github.com/odpi/egeria/tree/master/open-metadata-implementation/governance-servers/virtualization-services/virtualization-services-api/src/main/java) and moved derby connector the the upper level of directory as ViewGeneratorDerbyConnector in view-generator-connectors. Is this what you suggest in the first reply? Thanks in advance for your clarification!

PR #1741, opened 2019-10-14
Virtualizer connector refactor

wyf-1116:
Refactoring based on the comments in [Issue 1606](https://github.com/odpi/egeria/issues/1606).


planetf1:
Noticed this has failed. This check should not be mandatory but has crept in.

In actuality thinking a suitable fix would  be to set the threshold about where we are currently.

Will investigate pm and if not get build team (USA) to revert

planetf1:
Having looked at the sonar log - this isn't a case of the full scan filing the gate (which we know) but rather sonar reporting that code quality has got worse.

Note though that it is not a blocking check - so it will not prevent merge, but may be commented on by reviewers (we need much more work before we can insist due to our backlog)

The changes are in the asset lineage area - and as you say, I can't see any change in this PR that would cause sonar to identify them specific for this PR. 

It's possible sonar is 'catching up' after not running for a while whilst the build changed over - unlucky.

You can opt to fix them or leave them IMO. They do look reasonable minor issues to address.

Issue #1742, opened 2019-10-14
A localMetadataCollectionId is always expected to be present


danielaotelea:
I have the following setup:
1. An instance of OMAG platform running on my local environment - server1. This is configured to support OMASes
2. An instance of OMAG platform running on cloud- server 2. This is configured to support local graph repository. 
Those two instances are running in the same cohort. 
In this case, the local in instance - server1 should use the repo from server2. 
(Also, this issue is present when the configuration involves an external repo - Atlas).

PR #1743, opened 2019-10-14
#1742 - A localMetadataCollectionId is always expected to be present

danielaotelea:


danielaotelea:
For example, I started an instance of OMAG platform on my local with no repository configured. This platform was configured to start the OMASes - server 1. In the same cohort, I had configured an instance of Atlas that was available in the cloud - server 2.
On my local configuration, in registrystore the localRegistration was not present, but the remoteRegistrations was present and properly configured. 
When I have tried to retrieve an entity using the GUID, I've got an exception.
The entity was present in Atlas, after the fix that is present in this commit, I was able to retrieve the entity.

@lcpopa please describe the second test that we have done.

lcpopa:
Hello,
My test needed to create a simple process via DataEngine OMAS.
Setup was made of 2 platforms, one having enabled only the OMAS, the other only the OMRS with localGraphDb. Creation failed (more details below) in the OMAS enabled Platform
After apllying the changes from this issue's pull request, creation was successful.

Details about failure: NPE was thrown as variable localMetadataCollectionId was null. This shows us it was always expected to be present, which is not the case in this platform configuration. Checking it for null allows for remote lookup.
local_config
remote_config

PR #1764, opened 2019-10-15
egeria#1763 - Fix issue with  InternalOMRSEventProcessingContext

jnhagelberg:
Signed-off-by: Jeff Hagelberg <jnhagelberg@us.ibm.com>

This only affects consumers using the async api (so probably only OMRS).  The processing result should only have futures related to the current event being processed.  That is achieved by making it local to the thread, and clearing it before processing a new event.  However, the actual list of futures was accidentally made static.


jnhagelberg:
Just rebased.

jnhagelberg:
See egeria#1763

PR #1777, opened 2019-10-17
Use json sanitizer to prevent json injection

ruxandraRosu:
Signed-off-by: ruxandraRosu <ruxandra-gabriela.rosu@ing.com>

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/docker/omrs-monitor/Dockerfile
+++ b/open-metadata-resources/open-metadata-deployment/docker/omrs-monitor/Dockerfile
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Copyright Contributors to the Egeria project
 
-FROM node:12-alpine
+FROM node:13-alpine
 
 LABEL org.label-schema.schema-version = "1.0"
 LABEL org.label-schema.vendor = "ODPi"
--- a/pom.xml
+++ b/pom.xml
@@ -125,7 +125,7 @@
 
         <!-- Versions of dependent libraries -->
         <junit.platform.version>1.4.2</junit.platform.version>
-        <junit.jupiter.version>5.4.2</junit.jupiter.version>
+        <junit.jupiter.version>5.5.2</junit.jupiter.version>
         <junit.version>4.12</junit.version>
         <surefire.plugin.version>3.0.0-M3</surefire.plugin.version>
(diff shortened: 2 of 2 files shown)


ruxandraRosu:
I've added json sanitizer dependency as fortify scan reported JSONinjection vulnerability.
https://github.com/OWASP/json-sanitizer
If you agree with this, I can add the sanitize step on the other  classes reported for this issue.

cmgrote:
I'm not sure this resolves the potential security problem, if it's only ensuring that JSON is well-formed?

I believe JSON injection is the risk that you still receive what appears to be perfectly well-formed JSON (because you're just receiving a String that you deserialize into a JSON object), but actually it has JSON data within it that you do not actually want to consider valid for your particular use.  See: https://vulncat.fortify.com/en/detail?id=desc.dataflow.java.json_injection

I think the only way to confirm that such injection cannot occur is either to validate the specific JSON data in the object you receive (after deserialization, but before using any of it), or make use of strict Jackson types for deserialization (such that any unexpected or undefined JSON data is discarded rather than kept during the deserialization process) (?)

ruxandraRosu:
Hi @danielaotelea @davidradl 
I've added the json-sanitizer to subject area and security officer as well, would you please have a look?
As stated above, this is about formatting the json, additional verification maybe are required.

ruxandraRosu:
Hi Mandy,
I've added the sanitize step in 3 components - where method mapper.readValue(resultString, className) was used. Should it be added in some other places as well? This library only sanitizes the string before creating the object using jackson mapper, it cannot replace jackson dependency. Please let me know if I need to do some other changes

davidradl:
@ruxandraRosu I was wondering if we had non well formed json coming in, it would be good to log it - or even reject it - so the caller could be corrected ?

planetf1:
I'm in two minds about this -- I somewhat agree that the input should be correct, and we should be precise in how we treat it, yet at the same time I defer to those with more experience in projects such as OWASP who deal with security/injection issues regularly, and the extra checks such libraries offer I feel gives a better level of protection against such injection attacks.

https://www.owasp.org/index.php/Deserialization_of_untrusted_data
https://www.nccgroup.trust/globalassets/our-research/us/whitepapers/2018/jackson_deserialization.pdf

ruxandraRosu:
Hi,
Do you want to discuss this in next functionality call?

Issue #1757, opened 2019-10-15
Add jupyter notebook status file dir to excludes


planetf1:
Add .ipynb_checkpoints to .gitignore

If a notebook is used locally these files represent the state, but should not be checked into the build

mandy-chessell:
This is working for me

PR #1797, opened 2019-10-21
#1757 add jupyter notebook checkpoints to .gitignore

planetf1:
Adds .ipynb_checkpoint files to .gitgnore

These files are created if notebooks as auto-save files if notebooks are saved within the source tree.

Explicit notebooks will be added to git as usual, this prevents accidental auto-saves from being checked in.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -465,29 +465,8 @@
             </dependency>
 
             <dependency>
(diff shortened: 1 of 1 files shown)


mandy-chessell:
I fully support this proposal - thank you @planetf1

Issue #1795, opened 2019-10-18
Update egeria master to 1.2-SNAPSHOT


planetf1:

planetf1:
As we have branched for the 1.1 release we now need to switch our version to 1.2

planetf1:
done

PR #1801, opened 2019-10-21
move master to 1.2-SNAPSHOT #1795

planetf1:
Updates all references to 1.1-SNAPSHOT to 1.2-SNAPSHOT

@cmgrote - you will need to change dependencies in the connector repos if you want to update

Draft: checking additional changes


planetf1:
@cmgrote this change includes pointing to the 1.2 release of the igc connector.
Are you able to branch the connector repos accordingly or would you prefer I left that change out and you can then update at your leisure?

cmgrote:
I've put in draft PRs to merge the needed changes to move to `1.2-SNAPSHOT` on the connectors (and have created a `1.1-release` branch for each of them as well prior to doing so).

However, I wasn't clever enough to foresee the ordering issue:

- Really I should just have updated my own `pom`s to be `1.2-SNAPSHOT` but left them referencing `1.1-SNAPSHOT` of Egeria core (for now, otherwise they won't build until this PR is merged)
- Only after this PR is merged (and the built artifacts available in Artifactory) should I then have updated my `pom`s to use `1.2-SNAPSHOT` of Egeria core as well

For this release, are we OK accepting that there might be a few hours where any Helm chart deployments relying on the connectors will probably fail (until all PRs across all the repos are merged and the Artifactory artifacts for all of them are available)? 🙏

planetf1:
@chris - I think we should go ahead. A few hours out won't cause an issue. We should try and improve for the next release (there is already an issue open on image versioning which I am working on for the 1.1 release in any case)

Issue #1610, opened 2019-09-22
Create a THIRD_PARTY.md description of the licenses associated with embedded software


mandy-chessell:
To make it easy for someone to understand the licenses associated with our complete stack, we need to include a THIRD_PARTY.md file at the top of our git repositories that list the licenses in use for each technology and a link to the software's full license.

planetf1:
Agree, This is essential

In addition to our git repository, we may also need to ensure we have the correct licenses in each maven artifact, as well as in any 'archives' we build for distribution 

See also #898 

I've noticed a plugin that may assist in some of this - https://www.mojohaus.org/license-maven-plugin/

cmgrote:
I'm willing to take a first stab at the markdown file...

> I've noticed a plugin that may assist in some of this - https://www.mojohaus.org/license-maven-plugin/

Or was the intention to use this plugin to generate the markdown file itself?  (Or to use this plugin to ensure that appropriate license information is bundled with any assemblies?)

planetf1:
As much automation as possible as this is a continually moving target -- for example just when investigating spring I needed to try with replacing some java.* with jakarta.* - not to say that change will stick, but it's going to continue changing ....

We also need to figure out what needs documenting.  For example we might not need to document the build plugins we use - or other code that is used at build time only.

We also have a variety of UI components that are references in another issue which need covering too.

the doc in the PR is a super start.. and we do have a review process for PRs where we could remind devs to update the license, but it is prone to error, so if we can do anything better that would be good.

cmgrote:
Indeed automation would be good, but might be tricky for the top-level...  A few things I noticed when pulling together the initial list:

- Some of the dependencies are dual-licensed, presumably meaning a person needs to decide which license to we're using the dependency under
- The licenses are not always linked as you might expect: sometimes the GitHub repo lists them, sometimes it doesn't; sometimes they're listed in the Maven artifact metadata in Maven Central, sometimes not; etc
- The way licenses are referred to is not always the same: eg. some might say "EPL", others "Eclipse Public License", some "version 1.0" vs. "1.0" vs "1" vs "v1.0"
- It's worth having a sanity-check on the dependencies that we are including across the project this way as well: for example, I see the DataStax Cassandra driver and an Oracle NoSQL database driver -- not sure what these are being used for, but seems like they should be pulled-in at runtime and / or potential considerations for other (connector) repositories rather than included in the "core" (?)

[...]

PR #1802, opened 2019-10-21
Initial list of third party dependencies for #1610

cmgrote:


planetf1:
Thanks Chris - nice work.
Were you able to generate this with a script? 
Also the UI dependencies need adding
(more later)

cmgrote:
> Were you able to generate this with a script?

See comments on #1610 

> Also the UI dependencies need adding

I've simply taken everything that was in the root-level `pom.xml`, given our push to use dependency management...  Where / how are the UI dependencies defined?

planetf1:
I'm still waiting for feedback.

My suggestion is we merge this as a first-pass then make incremental improvements. I'll defer copying to 1.1 until more complete

Issue #1684, opened 2019-10-08
Data Engine OMAS - update functionality


popa-raluca:
This issue covers the update functionality for the Data Engine OMAS. 
The payload for updating processes will be the same as for the creation, Data Engine OMAS will add following checks for each process:
	* verify if the process exists based on the provided qualified name, if it does it will update the process properties.
	* verify if the portImplementations/portAliases are different from the new ones, based on the qualified names. Create the new ports, update properties for the existing ones and delete the old ones.  
	* verify if the schemType exists, based on the qualifiedName. If if does, update the proerties and verify the shcemaArributes. Update existing schemaAttributes, create new ones and delele the ond ones.
	* verify if a relationship exist before creating it(for all relationships - ProcessPort, PortDelegation, PortSchema, LineageMapping)

popa-raluca:
@cmgrote , @mandy-chessell can you confirm that this is the desired behavior for Data Engie OMAS on update? I'm not sure about removing old ports and schema types(that are no longer present in the payload).
Can I assume that the payload for a process will always include all the ports & schemaTypes needed? And should the old ones be deleted on update?

cmgrote:
In the case of DataStage, I expect to send you a complete payload each time even for the update (so that you can delete anything that is not included in the payload I send in the update that might already exist on your end).

However, I'm not sure whether we should assume this to be the case for _all_ data engines...  As such, it might be worth adding an "update semantic" enumeration to the update payload -- something along these lines:

- `REPLACE` would mean take only what is in the payload sent to the update, and remove anything that already exists that is not included in the payload
- `APPEND` would mean add whatever is in the payload sent to the update to whatever already exists (ie. don't delete anything)

popa-raluca:
Thanks @cmgrote , I will add the new updateSemantic enum to the payload. It will only be taken into account at update time, and it will be optional. The default behavior if the enum is not present will be REPLACE.
Do we need it need at process level, or is it ok to have something like:
{
	"processes": [
    	{...},
    	{...},
  	],
    updateSemantic: "REPLACE"
}

cmgrote:
> Do we need it need at process level, or is it ok to have something like:

I think either way is reasonable -- up to you.

popa-raluca:
Resolved with Data Engine OMAS - implement update functionality #1808

PR #1808, opened 2019-10-22
Data Engine OMAS - implement update functionality

popa-raluca:
https://github.com/odpi/egeria/issues/1684

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -144,7 +144,7 @@
         <spring-security.version>5.1.5.RELEASE</spring-security.version>
         <tomcat.version>9.0.21</tomcat.version>
         <commons-io.version>2.4</commons-io.version>
-        <commons-collections.version>4.3</commons-collections.version>
+        <commons-collections.version>4.4</commons-collections.version>
         <commons-lang.version>3.8.1</commons-lang.version>
         <springfox.version>2.9.2</springfox.version>
         <h2-database.version>1.4.196</h2-database.version>


open-metadata-implementation/access-services/data-engine/data-engine-api/src/main/java/org/odpi/openmetadata/accessservices/dataengine/model/Process.java
```diff
@@ -32,14 +32,15 @@
...
+    private UpdateSemantic updateSemantic;
 
     public Process() {
     }
 
     public Process(String qualifiedName, String name, String description, String latestChange,
                    List<String> zoneMembership, String displayName, String formula, String owner, OwnerType ownerType,
                    List<PortImplementation> portImplementations, List<PortAlias> portAliases,
-                   List<LineageMapping> lineageMappings) {
+                   List<LineageMapping> lineageMappings, UpdateSemantic updateSemantic) {
```
cmgrote: In addition to this new constructor with the update semantic, should we have another constructor with a "default" update semantic (which simply passes through to this one)?  I'd suggest the default should be `REPLACE`, as it appears to be further below; but without an additional constructor that defaults this then any implementation using these methods would need to explicitly send `null` to this constructor (or it won't even compile) (?)

open-metadata-implementation/access-services/data-engine/data-engine-client/src/main/java/org/odpi/openmetadata/accessservices/dataengine/client/DataEngineClient.java
```diff
@@ -24,7 +25,7 @@
 public interface DataEngineClient {
 
     /**
-     * Creates the process, with all the ports, schema types and corresponding relationships
+     * Create or update the process, with all the ports, schema types and corresponding relationships
```
cmgrote: Would it make sense to have update semantics on all of these revised methods as well (perhaps again 2 methods for each: one that defaults to `REPLACE` and another that takes the update semantic explicitly)?

open-metadata-implementation/access-services/data-engine/data-engine-client/src/main/java/org/odpi/openmetadata/accessservices/dataengine/client/DataEngineImpl.java
```diff
@@ -95,31 +97,32 @@ public DataEngineImpl(String serverName, String serverPlatformRootURL, String us
...
     @Override
-    public String createProcess(String userId, String qualifiedName, String processName, String description,
-                                String latestChange, List<String> zoneMembership, String displayName, String formula,
-                                String owner, OwnerType ownerType, List<PortImplementation> portImplementations,
-                                List<PortAlias> portAliases, List<LineageMapping> lineageMappings) throws
-                                                                                                   PropertyServerException,
-                                                                                                   InvalidParameterException,
-                                                                                                   UserNotAuthorizedException {
-        final String methodName = "createProcess";
+    public String createOrUpdateProcess(String userId, String qualifiedName, String processName, String description,
```
cmgrote: As commented on DataEngineClient, maybe worth update semantics on all of these methods (?)

open-metadata-implementation/access-services/data-engine/data-engine-client/src/main/java/org/odpi/openmetadata/accessservices/dataengine/client/DataEngineClient.java
popa-raluca: I added the update semnatic just for port implementations and processes for now. It doesn't make sense to add it at SchemaType or PortAlias level, since I can not determine what is the old SchemaType/PortAlias without having the PortImplementation/Process in the payload.
For a process with UpdateSemnatic.REPLACE, it will remove all old Port Implementations & PortAliases and create all new ones based on the process payload.
For a PortImplementation with UpdateSemnatic.REPLACE, it will remove the old SchemaType(&attributes) and create the new one based on the port implementation payload.

cmgrote: Just to check, when we say "replace" do we mean it literally does a full delete / purge of any existing pieces and a create (new) of everything?  Or is there some sort of determination of what's no longer in the new payload that was there previously, and only deleting those pieces + adding net-new ones that appear in the payload and are not found already in the environment?

Just thinking about the potential impact if we do a complete delete & complete new creation each time: that would create all new GUIDs for all of the underlying pieces, and potentially that would mean we'd lose any existing relationships against those underlying pieces?  My hunch is that while such an approach would be easier overall, it has greater potential for information being lost (?)

For example: DataStage sends through information about a new process and its assets. This all gets created via the Data Engine OMAS.  Then, somewhere else (even in Egeria itself) someone creates a new SemanticAssignment relationship to one of those pieces (either a process, one of its underlying SchemaTypes, or one of the assets that is read from / written to).  If we then have an update from DataStage with "replace" semantic, but actually the only piece that is different in the update is that there is one more asset that is written (say a log file), but we delete and purge _everything_ and then create _everything_ new, we will lose that SemanticAssignment relationship that was set in-between on one of the pre-existing assets (which actually hasn't changed)...

popa-raluca: Not everything will be deleted and recreated - only the entitites that are no longer in the payload will be removed. I'm determining what was changed bassed on the existing relationships. 
For example, for PortImlementations - DE is updating the ones from the payload, then retrieving all the PortImplementations for that process based on the ProcessPort relationship, then removing every port that has a relationship and is not in the payload. Only after removing the old ports I'm creating the new ProcessPort relationships.

If I take your example, then in the payload sent from DataStage proxy I will have all the old ports and schemas, plus an extra port implenentation with a schema type. In this case, nothing will be removed, the new entities will be added and the existing ones will be updated. 

I'm not sure about the performance though, haven't done enough tests yet. I tried to paralelize everything as much as possible, but for each entity from the payload now there is an extra call to retrieve it from the repository based on qualifiedName. If it is retrieved, then there is another call for updating that entity.

cmgrote: Excellent!

I've tried to emphasise whenever possible the criticality of `find...` methods working as fast as possible on `qualifiedName` for precisely these sorts of reasons -- that they will be called MANY times, so even what appears to be a tiny optimisation will have a significant impact 😉

PR #1814, opened 2019-10-22
[Asset Catalog OMAS] Unit Tests added 

danielaotelea:
1. unit tests added for client and spring module
2. fix the dependencies 
3. correct the URL and the responses 
4. change the long method  name for search

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-client/pom.xml
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-client/pom.xml
@@ -49,6 +49,11 @@
             <groupId>org.springframework</groupId>
             <artifactId>spring-web</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>com.mikesamuel</groupId>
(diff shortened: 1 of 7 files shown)


open-metadata-implementation/access-services/asset-catalog/asset-catalog-client/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/AssetCatalog.java
```diff
@@ -0,0 +1,309 @@
...
+        detectExceptions(methodName, assetResponse);
+
+        return assetResponse;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public RelationshipResponse getRelationshipBetweenEntities(String userId,
```
ruxandraRosu: This method should probably return directly the relationships(similar for the other methods). This can be done in a separate PR.

Issue #1791, opened 2019-10-18
DE OMAS - Update external data engine registration methods from software server capability 


cong78:
For Data Engine OMAS, there are methods and REST endpoints have been defined from clients side to create software server capability from external data engines. However, there are few things need to be added or changed to make the full logic working. For this issue there are following listed changes will be implemented: 
1. Update DE OMAS Client method from `createSoftwareServerCapability` to `createExternalDataEngine` to make it more intuitive and clear from other OMASes with similar capabilities. 
2. Add Common Services methods in `repositoryHandler` that has `createExternalEntity` and `createExternalRelationship` that can pass the external data engine's GUID. 
3. Update DE OMAS Server entity and relationship creation methods(handlers) in DE OMAS with parameters of external data engine's GUID. 
4. Update DE OMAS Client & Spring modules with updated methods that can pass external data engine's GUID.

@popa-raluca @cmgrote 
Please let me know if you have any questions. Thanks. :)

popa-raluca:
Looks good, just one more thing - common handler SchemaTypeHandler needs to be updated as well to make use of the newly added methods from RepoistoryHandler. 
DataEgineSchemaTypeHandler uses SchemaTypeHandler when creating and updating SchemaTypes & Schema Attributes

cong78:
From the conclusion of discussion in PR #1831 :
1. We will only keep the `externalSourceName` (qualified name from external data engine) to be included in the payload. And DE OMAS Server will be responsible to search the linked GUID of the external source. 
2. We will keep the value of `externalSourceName` as a property in DE OMAS client (cache). And when the client is dead the governance engine (like Data Engine Proxy) should be able to reset the value according to the payload.

PR #1831, opened 2019-10-23
DE OMAS - Including methods for external source GUID & Name

cong78:
Issue #1791

### DIFF
--- a/open-metadata-implementation/access-services/data-engine/pom.xml
+++ b/open-metadata-implementation/access-services/data-engine/pom.xml
@@ -23,12 +23,6 @@
 
     <artifactId>data-engine</artifactId>
 
-    <properties>
-        <junit.platform.version>1.2.0</junit.platform.version>
-        <junit.jupiter.version>5.2.0</junit.jupiter.version>
-        <junit.surefire.version>1.2.0</junit.surefire.version>
-    </properties>
-
     <dependencies>
 
     </dependencies>
--- a/open-metadata-implementation/access-services/governance-engine/pom.xml
+++ b/open-metadata-implementation/access-services/governance-engine/pom.xml
@@ -31,11 +31,6 @@
         <module>governance-engine-spring</module>
     </modules>
 
-    <properties>
-        <junit.platform.version>1.2.0</junit.platform.version>
-        <junit.jupiter.version>5.2.0</junit.jupiter.version>
-        <junit.surefire.version>1.2.0</junit.surefire.version>
-    </properties>
 
     <dependencies>
 
--- a/open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
+++ b/open-metadata-implementation/user-interfaces/access-services-user-interface/pom.xml
@@ -33,10 +33,6 @@
         <exec.plugin.version>1.6.0</exec.plugin.version>
         <jjwt.version>0.6.0</jjwt.version>
         <h2.version>1.4.196</h2.version>
-        <junit.platform.version>1.2.0</junit.platform.version>
-        <junit.jupiter.version>5.2.0</junit.jupiter.version>
-        <junit.surefire.version>1.2.0</junit.surefire.version>
-        <junit.version>4.12</junit.version>
     </properties>
 
     <dependencies>
--- a/open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test/pom.xml
+++ b/open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test/pom.xml
@@ -53,7 +53,6 @@
          <dependency>
                 <groupId>org.slf4j</groupId>
                 <artifactId>slf4j-api</artifactId>
-                <version>1.7.26</version>
         </dependency>
 
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -128,7 +128,7 @@
         <junit.jupiter.version>5.5.2</junit.jupiter.version>
         <junit.version>4.12</junit.version>
         <surefire.plugin.version>3.0.0-M3</surefire.plugin.version>
-        <mockito.version>2.28.2</mockito.version>
+        <mockito.version>3.1.0</mockito.version>
         <testng.version>6.14.3</testng.version>
         <log4j.version>1.2.17</log4j.version>
         <slf4j.version>1.7.28</slf4j.version>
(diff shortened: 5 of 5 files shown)


open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/mappers/SoftwareServerPropertiesMapper.java
```diff
@@ -3,8 +3,8 @@
 package org.odpi.openmetadata.accessservices.dataengine.server.mappers;
 
 /**
- * SoftwareServerPropertiesMapper provides property name mapping for Software Servers. Specifically, a single engine
- * is represented as an SoftwareServerCapabilityRequestBody entity.
+ * SoftwareServerPropertiesMapper provides property name mapping for external data engines. Specifically, a single data
+ * engine is represented as an DataEngineRegistrationRequestBody entity.
  */
 public class SoftwareServerPropertiesMapper {
```
cmgrote: Should we rename this class to be consistent with the other renamings (from `SoftwareServerPropertiesMapper` to `DataEnginePropertiesMapper`) (?)

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/handlers/ProcessHandler.java
```diff
@@ -77,17 +79,16 @@ public ProcessHandler(String serviceName, String serverName, InvalidParameterHan
...
+            throws InvalidParameterException, UserNotAuthorizedException, PropertyServerException {
         final String methodName = "createProcess";
 
         InstanceProperties properties = buildProcessInstanceProperties(userId, qualifiedName, processName, description,
                 latestChange, zoneMembership, displayName, formula, owner, ownerType, methodName);
 
         TypeDef entityTypeDef = repositoryHelper.getTypeDefByName(userId, ProcessPropertiesMapper.PROCESS_TYPE_NAME);
-        return repositoryHandler.createEntity(userId, entityTypeDef.getGUID(), entityTypeDef.getName(), properties,
-                InstanceStatus.DRAFT, methodName);
+        return repositoryHandler.createExternalEntity(userId, entityTypeDef.getGUID(), entityTypeDef.getName(), externalSourceGUID,
```
cmgrote: I notice in most (all?) places we're sending both the external source GUID and the external source name -- do we need both?  If the OMAS is handling this for us in all cases, then no problem; but otherwise it might be easier to ask external engines to provide one or the other (rather than both) (?)

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/mappers/SoftwareServerPropertiesMapper.java
cong78: True. I will rename this part. :)

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/handlers/ProcessHandler.java
cong78: I was also a bit unsure about whether I should include both of them in all methods. 
In the original methods from OMRSMetadataCollection it requires external GUID and Name (unique name) when metadata being created from external sources. From OMAS I think it would be good to keep the consistency from that. That's the reason that I did in this way. 

The implementation actually is being done slightly different from DE OMAS spring module and client module: 
If you want to call the DE OMAS REST APIs directly you need to include the external GUID and NAME in every payload. 
If you call it from client side, you need to the method of createExternalDataEngine to create the external engine GUID and Name, which will be save as DE Engine OMAS client side properties. So the following REST calls has been set to use the external source GUID and name from that client instance. In this case the payloads from external DEs can exclude the information with source GUID and name.

Not sure it is the ideal approach? I just did in this way when I was coding it yesterday.

cong78:
@mandy-chessell 
Hi Mandy, I have also added few methods of create entity/relationship in RepositoryHandler and SchemaTypeHandler. It would be nice if you can have a quick look as I might not aware some rules of adding new methods on them. Thanks.

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/handlers/ProcessHandler.java
cmgrote: Fair enough — let’s discuss briefly in today’s call and see what everyone thinks

popa-raluca: After today's call we concluded that we should only send the external source Name as a parameter for the methods, and retrieve the GUID of the source from the reporitory based on that name. This should be done internally in DE OMAS server.
The client should also follow the same approach as the spring module, and not save the GUID and Name as properties

cmgrote: One minor revision: I think the idea is that the client should expect to receive the name and use this to lookup the GUID, and can "cache" this internally so that it does not need to lookup the GUID on every single request -- but we should expect that the client at some point will be restarted (losing the cache) and that it therefore needs to lookup again based on the name (not create a new external source with the same name / different GUID each time it restarts, but also not expect to always have the GUID and name locally from the very start).

mandy-chessell: My understanding was  from the discussion that the REST API took the name and the server did the look up.  This way the client is effectively stateless.  This is good for an API where the external Id is an occasional use.  However for the OMASs that typically support an engine where the external Id is the same on every request, then I can see the value of caching in the client.

mandy-chessell:
Thanks @cong for doing that, I will have a look - but it is ok to merge before I get to see it.

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/handlers/ProcessHandler.java
cong78: Thanks a lot. I will write a conclusion in the issue and redo the logic of this part.

PR #1888, opened 2019-10-29
Unit tests for exceptions of Asset Catalog OMAS

wyf-1116:


innerTide:
@danielaotelea There are three methods that don't throw the expected exceptions:
- repositoryHandler.getEntityByGUID()
- metadataCollection.getLinkingEntities()
- metadataCollection.getEntityNeighborhood()

Could you please help me with this issue? Thanks! :)

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/test/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/AssetCatalogHandlerTest.java
```diff
@@ -29,12 +29,7 @@
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.TypeDef;
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.TypeDefLink;
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.repositoryconnector.OMRSRepositoryHelper;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.EntityNotKnownException;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.FunctionNotSupportedException;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.PagingErrorException;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.PropertyErrorException;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.RepositoryErrorException;
-import org.odpi.openmetadata.repositoryservices.ffdc.exception.TypeErrorException;
+import org.odpi.openmetadata.repositoryservices.ffdc.exception.*;
```
danielaotelea: It is a good practice to import only the classes necessary  from the package. If the number if the imported classes is greater that 20, then use import *.
In this case, please set up the IDE to import only the classes that you use.

PR #1896, opened 2019-10-29
Merged dependency updates from dependabot recommendations

planetf1:
Remaining dependabot proposals batched for efficiency

Tested notebooks, UI


cmgrote:
Thought we agreed in this morning's call that we'd only merge these in the mornings? 👀

PR #1911, opened 2019-10-29
Design models and glossaries

mandy-chessell:


open-metadata-resources/open-metadata-labs/egeria-server-start.ipynb
```diff
@@ -98,9 +98,89 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 11,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
planetf1: Output should be removed

Issue #1818, opened 2019-10-22
Investigate why Java 11 builds (Azure Dev Pipelines) take 30-50% more time than Java 8


planetf1:
If one looks at recent builds such as:
https://dev.azure.com/odpi/Egeria/_build/results?buildId=1095&view=results

it's noted that the Java 8 build is consistently faster - noticeably so. Between 30-50%, and typically at least 5 mins more. This slows down merging of PRs when processing a backlog

Need to investigate cc: @bramwelt

planetf1:
A few observations
 a) We are still creating the 'large' assembly artifact containing all egeria code which is no longer needed given our change to using finer grained artifacts from artifactory instead. - we could likely save 45s+ on both builds by removing this 
 b) On that particular build the repo upload took several minutes on java 11 side, less than 1 on 8 - but this is probably rather variable given upload times
 c) Whilst the overall maven elapsed time is indeed where we see a difference, the individual modules don't seem to differ that much
 d) The config for the jobs seems the same from the pipeline definition, differing only in java version. The same task version is used. The java compiler used is from azul systems but is at current levels (8/222, 11.0.4). These builds are openjdk based

planetf1:
A few local tests (MBP 2016)
 - OpenJDK/hotspot 11.0.5 : 13 minutes -- and this is repeated
 - OpenJDK/hostspot 1.8.0.222 : 8 minutes
 - OpenJDK/openJ9 11.0.5 : 8 minutes

I'd like to do many more iterations to be sure - and this test was on MacOS. The build environment uses ubuntu, with a openjdk based java (Azul - default in Azure)

planetf1:
Suggestion from openjdk team - 'Default GC implementation changes to G1 GC. ' wrt jdk 11
-> will try using same GC as 8

planetf1:
Changing GC had no noticeable difference.

Some research including https://medium.com/@john_freeman/faster-maven-builds-with-maven-opts-822cdc82fa85 suggests a few other options.

"-XX:TieredStopAtLevel=1" has had a noticeable affect, but still not as fast as 8 (13-> 11 vs 8 minutes). There is significant use of this option by various projects (google search), and the impact is to limit hit optimization, which can be valuable for short running tasks (like each compile).

Proposal is to add this to the build environment to speed up PR validation.

Keeping this open to identify other improvements

bramwelt:
[...]

PR #1916, opened 2019-10-30
#1818 limit java optimization to improve compile perf in verification…

planetf1:
… builds on Azure pipelines

See #1818 for background

Should improve Java 11 build performance


planetf1:
Note I've just changed verify jobs at this point - the others are less time critical and taking a conservative approach in changing minimum possible.

planetf1:
There areThe only question is how it affects performance time, but generally +10% or so. No negative effects noted.  numerous projects using this, and it's referred to in articles at nvidia, redhat etc in this context. As such I will attempt the merge, monitor, and back off if we see any regression

PR #1925, opened 2019-10-31
IGC samples updates

cmgrote:
- reworks the sample classifications and glossary metadata in IGC for clearer separation (see below)
- updated set of payload examples from IGC event mapper (pending: will include latest GUIDs)

The IGC glossary samples now look like the following:
Screenshot 2019-10-31 at 13 16 54

Specific changes:
- There is now a root-level category called `Classifications`: everything under it (categories and terms) is treated with a special meaning, and thereby ignored everywhere else as "normal" categories or terms
   - implemented classifications (like those seen to the right) should now all be under this IGC category
- All other root-level categories are now treated as OMRS `Glossary` entities
   - any terms that appear anywhere under the root-level (whether directly under `Coco Pharmaceuticals` in our example, or nested further under eg. `Organization/Hospital`) will be given a `TermAnchor` relationship to the root-level (Glossary)
   - any categories that appear anywhere under the root-level (as above) will be given a `CategoryAnchor` relationship to the root-level (Glossary)
- Classifications will be determined by relationships to those objects under the `Classifications` IGC category

Rationale:
- This more clearly separates the classifications from glossary-related content
- Enables the re-use of the classifications _across_ glossaries -- without needing to repeat them inside each glossary (in scenarios where there are multiple glossaries)

Changes to the IGC connector that rely on this new sample setup are part of PR https://github.com/odpi/egeria-connector-ibm-information-server/pull/121

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-client/pom.xml
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-client/pom.xml
@@ -67,7 +67,7 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-assembly-plugin</artifactId>
-                <version>3.1.0</version>
+                <version>3.2.0</version>
                 <configuration>
                     <descriptorRefs>
                         <descriptorRef>jar-with-dependencies</descriptorRef>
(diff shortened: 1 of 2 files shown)


cmgrote:
Discussed on 05/11 call and agreed to proceed with this change -- will merge once checks pass...

Issue #1826, opened 2019-10-23
SchemaType optimizations for Data Engine OMAS


popa-raluca:
Some changes need to be done in DE OMAS for including the schema types optimizations defined in https://github.com/odpi/egeria/issues/1669 & https://github.com/odpi/egeria/issues/1317

 * DataEngine OMAS creates/updates TabularSchemaTypes with the associated SchemaAttributes. Each SchemaAttribute will have a TypeEmbeddedAttribute classification for providing the type information. The SchemaAttributes will no longer have SchemaTypes associated to them. 
* Review and update the SchemaAttribute propeties - based on the changes done in https://github.com/odpi/egeria/pull/1545/files
* Replace the deprecated property 'name' with 'displayName' for SchemaAttribute

popa-raluca:
@cmgrote @mandy-chessell can you confirm that this is the desired behavior for DataEngine OMAS? At this moment, it only creates TabbularSchemaTypes with associated SchemaAttributes, so I assume that it shouldn't created the SchemaTypes for the attributes anymore?

cmgrote:
Sounds correct to me:

- Create / update a `SchemaType` for the `TabularSchemaType`
- Create / update `TabularColumn`s (including a `TypeEmbeddedAttribute` classification on each `TabularColumn` and an `AttributeForSchema` relationship between `TabularSchemaType` and each `TabularColumn`)

And indeed do not create any `TabularColumnType`s for the `TabularColumn`s.

What I can't recall is whether we're treating everything as Tabular, as this would imply, or whether we're using the more abstract (super) types: eg. `SchemaType` instead of `TabularSchemaType` and `SchemaAttribute` instead of `TabularColumn` (?)  Either way I think it will be important to be consistent and either use only the abstract super types, or only the more granular tabular-specific types...

popa-raluca:
It creates TabularSchemaTypes with SchemaAttributes right now, I'll modify it to be consistent

cong78:
@popa-raluca @mandy-chessell 
Can we have this feature being done in the 1.2 release instead of 1.3?

mandy-chessell:
There are only 3 days left to complete the release.  Can it be done in time?

cong78:
> There are only 3 days left to complete the release. Can it be done in time?

hmm in this case probably not then.

popa-raluca:
The implementation is done, I have a PR in draft for the changes. I will do some tests today and move it into ready status if everything is ok with the tests

[...]

PR #1937, opened 2019-11-01
Data Engine OMAS - Schema Type optimizations

popa-raluca:
Implements issue https://github.com/odpi/egeria/issues/1826.

SchemaAttributes no longer have SchemaTypes associtated with them, they have the associated TypeEmbeddedAttribute classifications.
LineageMapping relationships are now created between SchemaAttributes instead of SchemaTypes.

The TypeEmbeddedAttribute classifications are added in DataEningeSchemaTypeHandler for now, until the SchemaTypeHandler is updated for the SchemaType optimizations.
SchemaAttribute new properties(minCardinbality, maxCardinality, allowsDuplicateValues, orderedValues) are added as additionalProperties for the TabularColumns ritght now, SchemaTypeHandler will be refactored to add these new properties to a SchemaAttribute.

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/type-explorer/diagram-manager.js
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/type-explorer/diagram-manager.js
@@ -203,7 +203,7 @@ class DiagramManager extends PolymerElement {
         var diagSelector = this.$.diagramSelector;
         var opt = document.createElement('option');
         opt.value = "Inheritance";
-        opt.innerHTML = "Inheritance";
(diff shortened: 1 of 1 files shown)


popa-raluca:
I assumed that the new classification is mandatory, every schema attribute will have a TypeEmbeddedAttribute classification linked to it. 
Each classification will have a dataType property added, with the value being taken from the payload. 
Is this the desired behavior?

cong78:
> Each classification will have a dataType property added

In this case it is true each schemaAttribute will attach a new classification.

mandy-chessell:
@popa-raluca the TypeEmbeddedAttribute is **not** mandatory.  There are complex schema types that can not be represented by the classification.  It is more of a choice - either the classification is used or there is a relationship to the SchemaType.

PR #1938, opened 2019-11-01
UI Multi tenancy, adding view services and operational start stop capability 

davidradl:
After this change the ui config contains a list of view services, similar to the way the OMAG Server config contains a list of access services.  The view services register themselves and are configured using the same pattern as the access services.

The configuration and start and stop of the server does not currently effect the runtime. This pr is required prerequisite work, so that the runtime can can the view service instance handlers to obtain their configuration.


open-metadata-implementation/user-interfaces/access-services-user-interface/src/main/java/org/odpi/openmetadata/userinterface/accessservices/server/DiscoveryServerInstanceHandler.java
```diff
@@ -0,0 +1,56 @@
+/* SPDX-License-Identifier: Apache 2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.discoveryserver.server;
```
planetf1: I was surprised to see the discovery server content in the UI - is it needed?

davidradl: @planetf1  I agree this does not look right. I think it might have crept in with the merge. I will take a look

davidradl:
sorting out the merge issues - I will submit a new PR when ready

PR #1943, opened 2019-11-04
Fix NPE in GovernanceEngineTopicListener

mandy-chessell:


mandy-chessell:
The notebooks were showing as updated - I probably corrected layout or something minor - git would not let me push until these were committed.

planetf1:
SHouldn't be a problem with leaving notebooks out of the 'git add' - you can add specific files or subdirectories. Happens to me regularly as I use notebooks for test. The notebooks tend to show as updated whenever they are run locally (a recent change removed the checkpoints, but I think they auto save too). Not an issue though.

Issue #1770, opened 2019-10-17
Add an InTopic to DE OMAS


cmgrote:
I think we should have an InTopic for the Data Engine OMAS, so that things like the data engine proxy can hand-over asynchronous messages to the Data Engine OMAS rather than relying on synchronous API calls.

At the moment the data engine proxy services are using the Data Engine OMAS client to make API calls, and there is a long wait until all of these calls are processed by the OMAS before state is persisted back into the underlying data engine. I think it would be more robust to generate messages, pass these asynchronously, and then immediately store the data engine's processing state while the Data Engine OMAS is still processing?  Any subsequent state changes within the data engine can then be picked up and asynchronously passed across again (queuing up behind any others already in the topic) while the Data Engine OMAS continues to process the topic.

To give a sense of timings (and why I think this addition / change is important), with our simple 3-4 job Coco Pharma example:

- The data engine proxy services startup and processing of the samples takes ~15 seconds (~3 seconds of which is actual job lookup and transformation, the rest is service, connection, etc startup (one-off) overhead)
- The data engine OMAS processing (and onwards) takes more than 2 minutes from there
- Only once that is complete is the data engine (DataStage) given the go-ahead to store when it last sync'd changes

Should anything go wrong in the second bullet, the "last sync'd changes" marker in the underlying data engine is not updated -- despite it's actions (the first bullet) happening fairly quickly.

This relates to https://github.com/odpi/egeria-connector-ibm-information-server/issues/64, which is only possible if there is an InTopic on the OMAS to write into. This change would ensure there is minimal risk of a service disruption requiring changes to be re-processed from the data engine. (Instead they'd be state-persisted in the InTopic, ready for processing whenever service is restored.)

popa-raluca:
I assume this have to be done before the linege related release(https://egeria.odpi.org/release-notes/release-notes-1-3.html) ?

cong78:
Hi Chris I am starting working on this user story and from what I understood over the description, after the inTopic of DE OMAS, the IGC Connector will publish events directly without any intermediate steps to DE Proxy right?

cmgrote:
No, the IGC connector is not involved here...  The steps would be (after OMASes and proxies configured and an `InTopic` available):

1. DE Proxy polls for changes via its configured connection (in our example, the DataStage Connector)
1. DataStage Connector queries (via IGC, but not the IGC connector) for any changed objects since the last sync (or all, if no previous sync found)
[...]

PR #1945, opened 2019-11-04
Data Engine OMAS - Implementation of  InTopic

cong78:
See issue: #1770 

Why Draft:
Currently implementation I am reusing some methods from `DataEnineRestService.java `to prevent duplicating lots of codes and future refactoring.
And it also has a dependency on #1937 that is pending as a Draft PR. 
@popa-raluca 
We need to rework some logic parts to make some of the methods more reusable for both events and REST APIs when you are back from your holiday.

### DIFF
--- a/open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/pom.xml
+++ b/open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/pom.xml
@@ -90,6 +90,12 @@
             <artifactId>spring-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
 </project>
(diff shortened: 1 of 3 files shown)


cong78:
> Looks reasonable to me -- I haven't implemented these InTopics before, though, so just a sanity-check question that we're following the same pattern seen in other OMASes with InTopics?

Hi Chris yes. I followed other OMASes as well as OMRS event listener. I will let you know after talk to Raluca about some implementation details. :)

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/intopic/DataEngineInTopicProcessor.java
```diff
@@ -0,0 +1,335 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
+import java.util.*;
+import java.util.function.Predicate;
+
+import static java.util.stream.Collectors.partitioningBy;
+
+/**
+ * The Data Engine in topic processor is listening events from external data engines about
+ * metadata changes. It will handle different types of events defined in Data Engine OMAS API module.
```
cong78: typo

cong78: actually not. ....

PR #1972, opened 2019-11-07
Enforces UTF-8 encoding for REST calls to resolve #1872

### LINKED ISSUE #1872: Be wary of character encoding in RestTemplate
It seems that by default `RestTemplate` (at least `.exchange()`) will communicate any strings using ISO-8859-1 encoding rather than (as you might expect) UTF-8...

If you're making use of RestTemplate to send any information to REST endpoints, it might be worth explicitly setting the string message converter in use to UTF-8 (assuming you expect to handle anything other than basic English text) -- see https://github.com/odpi/egeria-connector-ibm-information-server/pull/117 for an example of how to do this.

cmgrote:


cmgrote:
This change fixed the problem where I was using RestTemplate in my connectors, so I believe it will do the same for the core (so long as everyone is re-using this rest connector) -- however, I suspect it will be difficult to confirm without actually merging the change for people to make active use of it.

Therefore, going to go ahead and merge, but please shout at me if it causes any problems!

Issue #1712, opened 2019-10-11
java.lang deprecations (Int, Double constructors etc)


planetf1:
Java 11 build reports deprecations in OCF:
```
15:14:57,223 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/ConnectorBroker.java:[169,75] newInstance() in java.lang.Class has been deprecated
15:14:57,224 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/ConnectorProviderBase.java:[222,59] newInstance() in java.lang.Class has been deprecated
15:15:00,022 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/test/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/TestCommentType.java:[30,36] Integer(int) in java.lang.Integer has been deprecated
15:15:00,023 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/test/java/org/odpi/openmetadata/frameworks/connectors/properties/TestAssetSummary.java:[208,38] Integer(int) in java.lang.Integer has been deprecated
15:15:00,024 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/test/java/org/odpi/openmetadata/frameworks/connectors/properties/TestAssetSummary.java:[225,61] Integer(int) in java.lang.Integer has been deprecated
15:15:00,024 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/test/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/TestStarRating.java:[29,36] Integer(int) in java.lang.Integer has been deprecated
15:15:00,025 [WARNING] /Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/frameworks/open-connector-framework/src/test/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/TestRelatedMediaUsage.java:[30,36] Integer(int) in java.lang.Integer has been deprecated
[...]

PR #1990, opened 2019-11-09
remove deprecation warnings for "new Integer(int)" cases as part of issue #1712

wbittles:
This is the change that removes the "deprecated" warnings issued by Sonar for "new Integer(int) cases.


planetf1:
* When committing you need to make sure you 'sign' the change. See https://gist.github.com/kwk/d70f20d17b18c4f3296d for resigning a change you've already made. Normally just use the '-s' option (assuming you have configured your email - see https://help.github.com/en/github/setting-up-and-managing-your-github-user-account/setting-your-commit-email-address & also have set user.name)
* Its a good idea to 'git pull upstream master' before opening a PR just to keep ahead of any conflicts.

Issue #1417, opened 2019-08-15
[warnings] enable compiler warning flags and resolve any warnings


cmgrote:
(This is required for lowest level **passing** for CII.)

The project MUST enable one or more compiler warning flags, a "safe" language mode, or use a separate "linter" tool to look for code quality errors or common simple mistakes, if there is at least one FLOSS tool that can implement this criterion in the selected language.

Examples of compiler warning flags include gcc/clang "-Wall". Examples of a "safe" language mode include JavaScript "use strict" and perl5's "use warnings". A separate "linter" tool is simply a tool that examines the source code to look for code quality errors or common simple mistakes. These are typically enabled within the source code or build instructions.

"We routinely set compiler warning levels as high as possible. It doesn't make sense to waste time trying to find a problem that the compiler could find for you! We need to concentrate on the harder problems at hand." ("The Pragmatic Programmer" by Andrew Hunt and David Thomas, p. 91-92) "Tip 23: Always use Source Code Control. Always. Even if you are a single-person team on a one-week project." ("The Pragmatic Programmer" by Andrew Hunt and David Thomas, p. 88)

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#warnings)

Furthermore:
The project MUST address warnings. [warnings_fixed]

These are the warnings identified by the implementation of the warnings criterion. The project should fix warnings or mark them in the source code as false positives. Ideally there would be no warnings, but a project MAY accept some warnings (typically less than 1 warning per 100 lines or less than 10 warnings).

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#warnings_fixed)

cmgrote:
I would assume this also applies to our build process (Maven), where we have a number of warnings along the lines of:

```
[INFO] --- maven-antrun-plugin:1.8:run (...) @ ... ---
[WARNING] Parameter tasks is deprecated, use target instead
```

```
[INFO] --- maven-jar-plugin:3.1.2:jar (...) @ ... ---
[WARNING] JAR will be empty - no content was marked for inclusion!
```

```
[INFO] --- maven-assembly-plugin:3.1.1:single (...) @ ... ---
[WARNING] Assembly file: /.../.../... is not a regular file (it may be a directory). It cannot be attached to the project build for installation or deployment.
```

and a few others related to deprecated method use, `npm install` (for Polymer?), missing JavaDoc bits, etc.

planetf1:
[...]

PR #1992, opened 2019-11-11
#1417 Enable additional lint options for javac

planetf1:
As per #1417 enables additional compile-time check for java

The java options have been changed so that
`-X lint:deprecated`
which was showing deprecation warnings only

Is now changed to 
`-X lint:all`

The compiler will show current options by using `javac -X help`

As an example, my current local macOS compiler (11.0.5, openJ9) will offer:

```
          auxiliaryclass       Warn about an auxiliary class that is hidden in a source file, and is used from other files.
          cast                 Warn about use of unnecessary casts.
          classfile            Warn about issues related to classfile contents.
          deprecation          Warn about use of deprecated items.
          dep-ann              Warn about items marked as deprecated in JavaDoc but not using the @Deprecated annotation.
          divzero              Warn about division by constant integer 0.
          empty                Warn about empty statement after if.
          exports              Warn about issues regarding module exports.
          fallthrough          Warn about falling through from one case of a switch statement to the next.
          finally              Warn about finally clauses that do not terminate normally.
          module               Warn about module system related issues.
          opens                Warn about issues regarding module opens.
          options              Warn about issues relating to use of command line options.
          overloads            Warn about issues regarding method overloads.
          overrides            Warn about issues regarding method overrides.
          path                 Warn about invalid path elements on the command line.
          processing           Warn about issues regarding annotation processing.
          rawtypes             Warn about use of raw types.
          removal              Warn about use of API that has been marked for removal.
          requires-automatic   Warn about use of automatic modules in the requires clauses.
          requires-transitive-automatic Warn about automatic modules in requires transitive.
          serial               Warn about Serializable classes that do not provide a serial version ID.
                             Also warn about access to non-public members from a serializable element.
          static               Warn about accessing a static member using an instance.
          try                  Warn about issues relating to use of try blocks (i.e. try-with-resources).
          unchecked            Warn about unchecked operations.
          varargs              Warn about potentially unsafe vararg methods
          preview              Warn about use of preview language features
```

This changes DOES NOT fail the build if any of these checks fail, nor does it fix any. It is purely adding information for the developer who should be encouraged to view these reports when building and testing their code

As such in itself it doesn't completely address #1417, but only enables.


planetf1:
For reference a quick check on my current build shows the following additional warnings - manageable:
```
 102 unchecked call 
  24 unchecked cast
 134 unchecked conversion
   1 unchecked generic array
  47 unchecked method invocation:
```

Issue #1633, opened 2019-09-26
Add reference copy tests to CTS


grahamwallis:
Need to add Ref Copy tests and associate them with the REFERENCE_COPY_STORAGE, REFERENCE_COPY_LOCKING and REFERENCE_COPY_DELETE requirements of the optional REFERENCE_COPIES profile

grahamwallis:
Fixed by PR #1994

PR #1994, opened 2019-11-11
Extensions to CTS for release 1.2 (issue #1633)

grahamwallis:


cmgrote:
Not sure what's happening on the build here -- is this the CTS update that we are targeting the connectors to be conformed against for release 1.2?

grahamwallis:
Hi Chris - I ran into a problem with external instances - it is now fixed - just re-running the CTS against both in-mem and graph repositories to make sure and will then update the PR. Should be very soon.

PR #1995, opened 2019-11-11
Development ui - add filters to asset search view, fix breadcrumb bugs and styling and add swagger to UIPlatform

ruxandraRosu:


ruxandraRosu:
> I tried out the changes to the breadcrumb UI
> 
> * On 'type explorer' the text to enter the address/servername now seems partially cut - see https://pasteboard.co/IHfhmww.png
> * Why are the titles 'Open Metadata - asset-search' instead of 'Open Metadata - Asset Search' https://pasteboard.co/IHfiDIL.png
> * The trailing > is gone - thanks
> * I still find the mental model is confusing, with 'Home' something that doesn't exist, since on opening the UI one is placed straight into Asset Search, so it's superfluous in the navigation - though perhaps actually having a 'Home' is more clear
> * The rogue asset lineage '3rd element' navigation entry issue has been addressed
> * NEW PROBLEM. If I go to Asset lineage, click 'ultimate destination' , then go to type explorer, and back to asset lineage, the breadcrumb shows 'Home -> Asset Lineage ultimateSource, yet the highlighted tab is ultimate destination https://pasteboard.co/IHfjYWg.png
> * similar to the first point, a navigation trail of 'Home > Asset Lineage sourceAndDestination' still seems to be using 'internal' labels rather than a normal name'. Why don't we just use 'Source and Destination'  It's also confusing it's asset Lineage with no separator, so perhaps 'Asset Lineage: Source And Destination' but I think it's another manifestation of the confusion between horizontal and vertical navigation. A mixed model. The page already has a left 'nav bar' & the lineage page already has a tab with underlines, so does the breadcrumb trail actually add anything? I'm wondering if it causes more confusion
> 
> There is a lot of inconsistency in the UI -- we know of course it is work in progress. Perhaps we should focus on this in our next face to face and see what general opinion in the team is, including of the navigation.

Hi @planetf1 ,
I fixed the issue 1&6 in this PR. For issue#1, I've changed the absolute location, but probably the proper solution is to replace all absolute locations with relative locations. I think this can be done in a separate PR along with other inconsistencies between UIs. For #2, I think this was already like this before. For #4 and #7, we've briefly discussed about in the last functionality call, we can re-discuss tomorrow.

planetf1:
thanks - makes sense. I would suggest we are ok with merging, but discuss the layout & bread crumbing in general on a call/meeting ie for the more general issues

davidradl:
It seems that the small number of types you have hard coded makes it easy to use, I am not sure if we have a large number of types that this ui component is appropriate. I wonder whether you should expose the types in the user search. Maybe start by choosing asset type, so in your case choose database, the user is then allowed to check the table and column, maybe with a checkbox to include glossary terms.  So if the user chooses an Avro assert type then further Avro specific filters would be displayed. What do you think? The other concern I had was can we not expose the type guid in the UI / API. I think we should be able to cope with a type name being supplied.
Also I notice that there are error boxes displayed if I click around the open lineage - it does not seem to be checking that it has not got a guid. I am not sure if this is to do with this check in or not.

I am also not sure what value the breadcrumb gives, as all its information is in the url (and Home does not mean anything).


 

@ruxandraRosu I would like you thoughts on these issues.

ruxandraRosu:
> It seems that the small number of types you have hard coded makes it easy to use, I am not sure if we have a large number of types that this ui component is appropriate. I wonder whether you should expose the types in the user search. Maybe start by choosing asset type, so in your case choose database, the user is then allowed to check the table and column, maybe with a checkbox to include glossary terms. So if the user chooses an Avro assert type then further Avro specific filters would be displayed. What do you think? The other concern I had was can we not expose the type guid in the UI / API. I think we should be able to cope with a type name being supplied.
> Also I notice that there are error boxes displayed if I click around the open lineage - it does not seem to be checking that it has not got a guid. I am not sure if this is to do with this check in or not.
> 
> I am also not sure what value the breadcrumb gives, as all its information is in the url (and Home does not mean anything).
> 
> @ruxandraRosu I would like you thoughts on these issues.

Hi David,
Regarding the filters for types, we discussed to have it in a previous meeting related to asset search ui together with Mandy. What we have in mind is to display all supported types for searching, the multi-select component allows for search while typing. The type guid is not visible in the ui, why is it a concern to have it in the API? We can discuss further if we should use the type names in the APIs, this might also require changes on asset catalog endpoint. The guid validation should be handled separately, it is not related to the multi-select. 
As Nigel suggested, I think we can discuss in another meeting about the layout issues: breadcrumb, multi-select component and any concerns regarding the types guids

ruxandraRosu:
Hi,
I would like to merge it as it also fixes some bugs that broke search functionality completely in master caused by oversights in refactoring in backend components. Regarding the concerns related to this PR: breadcrumb and type guids used in APIs, is it ok to address them separately?

davidradl:
raised #2037 to address the exposing of the type guids in the omas

Issue #1593, opened 2019-09-19
Dependency issue of Data Platform Service server side with generic API implementation


cong78:
Currently the Data Platform Service under Governance Server is adding the dependency from Cassandra Connector, which is not architecturally correct and may introduce potential security and build issues. One of them is an existing issue #1519 that Cassandra maven dependency version is old with an insecure version of Guava.

As the next step, there should be a generic interface in API module from Data Platform Service side for any potential data platforms who want to plug in. Which is showed in the following drawing:
![data-platform-services copy](https://user-images.githubusercontent.com/11947525/65242462-54b51e80-dae6-11e9-88fc-862d627a5055.png)

For this issue, there are following tasks should be implemented: 
1. Refactoring of Data Platform Service Server module and remove the Cassandra Dependency. 
2. Implementation of generic interface in Data Platform Service API module.
3. Implementation of Cassandra Connector to adapt this new generic interface.

mandy-chessell:
Thank you @cong78

mandy-chessell:
One other point to look at is the ConnectorConfigurationFactory.  This should not include references to specific connector implementations (except for the core connecctors for OMRS).

The method `getDataPlatformConnection` has the connector provider class name is passed in as a string - but then towards the end, it is ignored and cassandra provider is used to create the connector type.  This means the connection returned is always for cassandra.

cong78:
For Apache Cassandra, we also need to build a data-store-connector to retrieve the Connection Object of Cassandra Assets.

mandy-chessell:
The Cassandra data-store-connector instance is for accessing data on cassandra.  An instance of this connector is created by the ConnectorBroker using the Connection object attached to the Cassandra Asset.  This is done in the Asset Connumer's program.

The Cassandra data-store-connector is different from the Cassandra metadata extractor connector that runs in the Data Platform server.

cong78:
> This is done in the Asset Connumer's program.

Is that means the Asset Consumer OMAS will create the connection object for Cassandra data-store-connector instance ?

mandy-chessell:
The Asset Consumer OMAS is responsible for creating the **connector** instance.  It uses the Connection object attached to the Asset when the Asset is created.   The Data Platform server needs to create the Connection for the Cassandra data-store-connector for each Cassandra Asset it creates.  This Connection object is linked to the Asset

cong78:
[...]

PR #2005, opened 2019-11-13
Data Platform Services and Cassandra Connectors - fix some methods and update documentation

cong78:
See issue: #1593

Update documentation and fix some connector bugs.
@mandy-chessell 
For now I am adding the JAR of cassandra-metadata-extractor-connector to the server-chassis-spring as dependency when configuring the data-platform-service through POST call. I asked Dimitrios he is also doing it way so I hope it is architecturally correct.


cong78:
@planetf1 
Hi Nigel, do you know why DCO is not responding? I don't think I have any of commits not signed.

planetf1:
Hi. - no I don't know about DCO, but it's very standard GitHub function, so my guess is GitHub themselves have an issue. I can't see anything in status, so I'd recommend redoing your last push (you can just force push to your branch) to re-trigger the check

PR #2021, opened 2019-11-18
Extend kafka docs to cover security

### LINKED ISSUE #838: Kafka: add notes on setup with secured kafka, required topics, debugging
When trying to get Egeria working with IBM Event streams (a secure implementation of Kafka on public cloud) I noticed that after starting up egeria and configuring into a cohort, a stream of messages such as:

```
2019-03-13 09:22:57.309  WARN 18542 --- [ionView.inTopic] org.apache.kafka.clients.NetworkClient   : [Consumer clientId=consumer-3, groupId=0d1dc807-681b-4b6f-a430-d3cec52001ad] Error while fetching metadata with correlation id 374 : {null.server.myserver.open-metadata.access-services.InformationView.inTopic=UNKNOWN_TOPIC_OR_PARTITION}
```

were logged at INFO level 

However nothing was visible if logging was disablled. There are no audit messages logged.

The CAUSE of this issue (which has functional impact) was that the secure implementation is not configured to auto create topics. So the topics needed by many egeria components do not exist.

A few issues become clear
 * Where does a user go to ensure they can tell the kafka team (or configure manually themselves) the FULL list of kafka topics that Egeria needs
 * The log (INFO) message should contain the topic name & any other salient information
* There needs to be at least one audit message (or a low frequency periodic) reporting the fact that the topic does not exist.
 * the attempt to launch the server instance reported 200, with a result including:
```
{
    "class": "VoidResponse",
    "relatedHTTPCode": 200,
    "successMessage": "Wed Mar 13 09:22:04 GMT 2019 myserver is running the following services: [Repository Services, Data Engine OMAS, Subject Area OMAS, Governance Engine OMAS, Governance Program OMAS, Information View OMAS, Asset Consumer OMAS, Connected Asset OMAS, Asset Catalog OMAS]"
}
```
Which may be deemed slightly misleading since those services have been launched, but are not in any way running correctly.

Since we may have a situation where kafka comes up later we might want to consider how else to handle this - perhaps there should be an additional status check for each individual OMAS to report it's fine grained status for example.. Needs consideration.

planetf1:
* Corrected incorrect docs in topic connector - out of date default values
* Added docs on configuring kafka connector in a secured environment
  -> Specifically for IBM Event streams so far (only example tested)
* Added docs for VDC helm chart to cover EventStreams configuration (kafka secured)
* Typos, crosslinks

Fixes #838

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,33 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the ODPi Egeria project.
+---
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 30 days if no further activity occurs. Thank you
+  for your contributions.
+        stale-pr-message: >
+  This PR has been automatically marked as stale because it has not had
+  recent activity. It will be closed in 30 days if no further activity occurs. Thank you
+  for your contributions.
+        stale-issue-label: 'no-issue-activity'
+        stale-pr-label: 'no-pr-activity'
+        days-before-stale: 180
+        days-before-close: 30
+        exempt-issue-label: 'pinned'
+        exempt-pr-label: 'pinned'
+        operations-per-run: 30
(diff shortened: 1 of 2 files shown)


cong78:
Hi Nigel, looking good. 
Just a question the VDC helm chart is also specifically IBM Event Stream right?

planetf1:
> Hi Nigel, looking good.
> Just a question the VDC helm chart is also specifically IBM Event Stream right?

It is, though that's mostly because it was the only cloud based/secure kafka service I tried/had access to. It may work with other providers, and if not the pattern is good to follow, and we can make custom logic for another provider.

Do you have another in mind?
Would you like to contribute anything?

Happy to help ... 

Also the VDC chart itself will likely need more work to match the requirements of what you might wish to deploy - again that's something else we can work on as needed.

Issue #469, opened 2018-12-17
gaiandb url as property in pom.xml


bogdan-sava:
As in some environments there is no internet connection and build fails as cannot access github for download it is useful to have such urls as property to be able to override them.

mandy-chessell:
Is this fixed?

planetf1:
I'm not aware of any reports of issues in this area recently - @bogdan-sava if this is still a problem re-open and I will queue to address

planetf1:
Reopened after discussion in 1843.. cc: @cong78 @mstrelchuk 

We download many artifacts during the build
 - maven dependencies
 - 'node'/npm especially for UI
 - polymer/web components
in addition to gaian, so can the same approach be taken with caching locally?

I note that recently https://github.com/odpi/egeria/pull/1910 added parms for the node/npm download, so would the same approach work for gaian? Then in your environment you could cache on artifactory and set the build properties externally (just as happens in azure pipelines for environment specific settings)

Also related are a variety of issues considering removing gaian/ranger from the main repo such as #958, and moving into a specific repo - but again that is longer term, and I think your need is shorter term?

planetf1:
The top level pom now has a new property 'gaiandb.url' which can be set to the location of the GAIAN distro .
If this isn't sufficient please can you reopen @bogdan-sava

PR #2023, opened 2019-11-18
Add maven dependency license files to build

planetf1:
See #1610 - auto generation of license files (maven)

### DIFF
--- a/open-metadata-implementation/adapters/pom.xml
+++ b/open-metadata-implementation/adapters/pom.xml
@@ -21,10 +21,6 @@
         plug open metadata capability into another technology.
     </description>
 
-    <properties>
-        <gaiandb.url>https://github.com/gaiandb/gaiandb/blob/master/build/GAIANDB_V2.1.8_20160523.zip?raw=true</gaiandb.url>
-    </properties>
-
     <artifactId>adapters</artifactId>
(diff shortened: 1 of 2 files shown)


planetf1:
Build has failed on gaian/ranger.

This is due to the fact that gaian is downloaded and added into a local repo during the build.
However the license tool runs during the 'process-resources; phase, which is earlier.
Didn't cause. problem locally as once downloaded the gaian repo is left in-situ

Will need to adjust the sequencing - probably ensuring gaian plugin is downloaded earlier (process-resources or before in the lifecycle)

planetf1:
Migrated the download of gaian into top level pom & moved to initialise phase so that it is done very early on before on dependency, license, security checks -- otherwise other maven goals can fail/warn due to these not being present.

Additionally contained the download within a profile, activated by default, so that if an org wishes NOT to download additional files direct from the internet during the build, they can instead host on their own local artifactory (since gaian is not available publicly except in zip format) which should fix #469. While not explicitly offering a URL, the net effect is the same since additional maven repos can be configured when executing the build, and these will also be searched for artifacts - in our case gaiandb & derby

planetf1:
To avoid download use 
mvn clean install -DskipThirdPartyLocalRepo

Issue #1385, opened 2019-08-15
[report_responses] issue cleanup of anything > 6 months old


cmgrote:
(This is required for lowest level **passing** for CII.)

The project MUST acknowledge a majority of bug reports submitted in the last 2-12 months (inclusive); the response need not include a fix. 

The response MAY be 'no' or a discussion about its merits. The goal is simply that there be some response to some requests, which indicates that the project is still alive. For purposes of this criterion, projects need not count fake requests (e.g., from spammers or automated systems). If a project is no longer making enhancements, please select "unmet" and include the URL that makes this situation clear to users. If a project tends to be overwhelmed by the number of enhancement requests, please select "unmet" and explain.

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#report_responses)

cmgrote:
Hi @mandy-chessell -- should we consider installing something like https://probot.github.io to automatically mark and manage stale issues?  I've seen such mechanisms used in various other projects (and we are reaching the point where some issues are at nearly 12 months of no activity).

(@jmertic not sure if there is a "preferred" such bot already? cc: @planetf1)

jmertic:
Good idea @cmgrote! I don't know of a preferred tool yet, but saw this probot script which I think would be perfect for what you are trying to do...

https://github.com/apps/stale

cmgrote:
Seems like https://github.com/marketplace/stale is now owned and operated by GitHub itself (like Dependabot)...

Suggesting we consider adding it to keep us diligent on issue hygiene 😬  TBD on our next functionality call?

cmgrote:
@jmertic we discussed this on our community call 22/10 and agreed that we'd like to set up this bot (https://github.com/marketplace/stale) on all of our Egeria repos.  Can you help us enable / install this on odpi/egeria, odpi/egeria-connector-ibm-information-server, and odpi/egeria-connector-apache-atlas ?

In particular, we'd like to go with these parameters to start with (though we might be able to configure this bit ourselves once the bot is installed / enabled on the repos):
- `daysUntilStale: 180`
- `daysUntilClose: 30`
- `exemptLabels: [ "pinned", "security" ]`

jmertic:
I've enabled this app now. I also added a stale.yml as a PR for the team to review -> https://github.com/odpi/egeria/pull/1817

planetf1:
@cmgrote assume you were driving this. Not seen any action yet. Let me know if you need me to help

[...]

PR #2024, opened 2019-11-18
Create stale.yaml for GitHub 'stale' action

planetf1:
Alternative implementing for #1385 using GitHub action's native 'stale' rather than probot

Note that only one label is permitted to be exempt. Other settings copied from the probot config - that config can be cleaned up if this works

cc: @cmgrote


planetf1:
Interestingly this PR was made using the native GitHub UI - I guess it doesn't handle DCO. 
Will create a manual PR.

PR #2025, opened 2019-11-18
#1385 Configure github actions stale action

planetf1:
Alternative implementing for #1385 using GitHub action's native 'stale' rather than probot

Note that only one label is permitted to be exempt. Other settings copied from the probot config - that config can be cleaned up if this works

cc: @cmgrote


planetf1:
see docs at  https://github.com/marketplace/actions/close-stale-issues

PR #2041, opened 2019-11-20
#1385 correct errors in stale.yml for autoclosing of issues

planetf1:
See #1385 for info.

The current stale action was failing due to a yaml error (missing ':').
In fact this was caused by incorrect indentation (typical yaml!) of the 
multi line issue/pr text

Now runs nearly clean through yamllint (only complaint is 'on' but that is documented syntax for the GitHub actions).


planetf1:
Also deleted old file from prior attempt using probot

PR #2043, opened 2019-11-20
Replace H2 with HSQLDB

bogdan-sava:


pom.xml
```diff
@@ -144,7 +144,7 @@
         <commons-collections.version>4.4</commons-collections.version>
         <commons-lang.version>3.9</commons-lang.version>
         <springfox.version>2.9.2</springfox.version>
-        <h2-database.version>1.4.196</h2-database.version>
+        <hsqldb-database.version>2.4.0</hsqldb-database.version>
```
planetf1: 2.5.0 is current

planetf1:
I don't see any explanation around this change. It would be good to have an issue proposing.
But let me ask here

HSQL seems an older/less well maintained database than H2, and elicits FAR less hits on google. Given that, what are your reasons for recommending we switch to hsqldb? What are the other options?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/pom.xml
```diff
@@ -65,8 +65,8 @@
         </dependency>
 
         <dependency>
-            <groupId>com.h2database</groupId>
-            <artifactId>h2</artifactId>
+            <groupId>org.hsqldb</groupId>
```
planetf1: Why hsqldb?

bogdan-sava: H2 new version has compatibility issues

bogdan-sava: and it's a toy database anyway used only for demo purpose

bogdan-sava: is hsqldb not acceptable ?

bogdan-sava:
> What are the other options?

any database can be used as it is using JPA that makes it plugable

bogdan-sava:
> HSQL seems an older/less well maintained database than H2

how do you end-up to this conclusion ? any documentation? any benchmark?

planetf1:
I don't have a specific problem with moving to another database, I'm just keen we agree with the broader team when we're changing a dependency & explain the reasons why. The scope of where we use H2 is of course small, which does reduce the concern. So I think it should be proposed for review, with some considerations such as 
 - how vibrant is the project - will issues be fixed, will CVEs being addressed
 - How is it licensed
 - What functional impact does it have (UI login only in test useage)
 - what problems does it address
 - what other options are there

The popularity comment I made  was very simply based off google hits, and vibrancy from looking at the repo, but in balance hsqldb appears to get good reviews, and is certainly more scalable (probably not a factor here). We do need to address the h2 issue and agree this is one way

So it's just about documenting & explaining the reasons for the change, not that my opinion is a  specific no to switching

davidradl:
On observation: we had discussed creating our own custom in memory version- this would be more likely to be portable to non-Spring environments.

grahamwallis:
The license on HSQLDB is not one of the regular (approved) licenses. It looks OK to me but I am not a lawyer. We should get it screened before committing to using HSQLDB.

I don't have an opinion either way between H2 and HSQLDB, but on a general point of principle we should be proposing (in an issue) a dependency change like this so it can be discussed - in case it affects other components or we as a community have concerns. Maybe this particular change is something we could discuss on the Thursday call (today) to expedite this.

mandy-chessell:
Do we need this component at all ?

planetf1:
@grahamwallis good point on the license - the web page does show a modified license which technically means another legal approval, specific exclusions/acceptance in the build & nexus - though it seems ok

@davidradl - I think both H2 and HSQLDB are usable outside spring

@bogdan-sava In terms of scalability I think it's unlikely an issue given our useage. A few users for login in a test environment only.

Another observation is h2is developed in the open on GitHub, whilst hsqldb though BSD2 licensed is not open in terms of development (as best I can see) - so less ability to understand, less aligned with our open goals, harder to get support [but again, due to scope this doesn't mean we can't.. just some observations]

Issue #2030, opened 2019-11-19
Create CVE lists for release


planetf1:
Currently CVE lists are built by
 - Nexus IQ scans
 - Running mvn clean install -Pspotbugs

In the latter case we make use of the 'owasp' plugin

This issue is being opened to
 a) Run the scan by default (if quick)
 b) Aim to build an aggregate list 
 c) Link to results (or post) as part of release

Targeting 1.2 to aid in CI badge requirements

planetf1:
Using the dependency-check-maven plugin I have added an OWASP scan to the top level module only. This does add around 60s to the build time. It is enabled by default but can be disabled with `-D skipCVE` 

This will generate a file in target/dependency-check-report.html that is an aggregate across all subproject within maven.  This file could be posted alongside the release to summarise the current vulnerability status for egeria.

Also doing this for individual submodules adds an unreasonable extra processing time to the build so this has been removed

Example report: https://gist.github.com/b19714bf5d301054cb9405106de9815f
(Download and open as html)

PR #2045, opened 2019-11-20
#2030 generate CVE report under target/dependency-check-report.html

planetf1:
As per #2030 this will generate a CVE vulnerability report as a standard part of the build.

Currently the build will not fail (though this is an optional configuration for the future)

@mandy-chessell a few thoughts
 a) is the report what you expect [I think it is!]
 b) Ok to run by default? If not I could run on merge or release only
 c) Ok with output filename? I can name differently if needed.

I propose getting this into 1.2 so that I can take a copy and post alongside the release.

We can do a better job of collating all the report info in a future release.

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
@@ -263,17 +263,19 @@ public class TestSupportedRelationshipSharingPropertySearch extends RepositoryCo
 
                 for (Relationship relationship : result) {
                     InstanceProperties relationshipProperties = relationship.getProperties();
-                    InstancePropertyValue ipValue = relationshipProperties.getPropertyValue(attrName);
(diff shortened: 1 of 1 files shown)


planetf1:
The Java 8 build has failed with
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.2.0:single (assemble-all) on project janus-connector: Execution assemble-all of goal org.apache.maven.plugins:maven-assembly-plugin:3.2.0:single failed: Plugin org.apache.maven.plugins:maven-assembly-plugin:3.2.0 or one of its dependencies could not be resolved: Failed to collect dependencies at org.apache.maven.plugins:maven-assembly-plugin:jar:3.2.0 -> org.apache.maven.shared:maven-common-artifact-filters:jar:3.1.0: Failed to read artifact descriptor for org.apache.maven.shared:maven-common-artifact-filters:jar:3.1.0: Could not transfer artifact org.apache.maven.shared:maven-common-artifact-filters:pom:3.1.0 from/to central (https://odpi.jfrog.io/odpi/egeria-release): Failed to transfer file https://odpi.jfrog.io/odpi/egeria-release/org/apache/maven/shared/maven-common-artifact-filters/3.1.0/maven-common-artifact-filters-3.1.0.pom with status code 503 -> [Help 1]
```
It's possible we are exceeding some API request limit (cc: @bramwelt) or it may be a transient error

planetf1:
agreed ok in team call

Issue #2048, opened 2019-11-21
Branch 1.2 release & update master to 1.3-SNAPSHOT


planetf1:
It's time to branch for release 1.2
As part of this our version identifiers will be updated
 master -> 1.3-SNAPSHOT
 1.2 branch (1.2-SNAPSHOT) -> 1.2

planetf1:
done as above

PR #2049, opened 2019-11-21
#2048 Move version up to 1.3-SNAPSHOT

planetf1:
Moves master up to 1.3-SNAPSHOT

### DIFF
--- a/developer-resources/License-Example-Files/License_for_POM_Files.xml
+++ b/developer-resources/License-Example-Files/License_for_POM_Files.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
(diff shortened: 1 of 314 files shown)


planetf1:
Unfortunately not that I'm aware of - it fails to resolve as it doesn't seem to do variable replacement.

We also have references to the version in other places - documents, other files. We should try and replace with variable substitution where we can - I agree (feel free to open an issue on that) but it will be tricky in some areas unless we either
 - use a build process for docs to resolve substitutions 
 - use some kind of check-in action to insert the correct value

planetf1:
There are some additional plugins that will pre-process the poms, but that adds more unknowns. Also the maven-release-plugin can change all the pom versions - but the updates still need checking in, and it doesn't handle the files elsewhere. So we're still mostly stuck with a global text change. At least it's quick

As above though we can open an issue to explore improvements

PR #2069, opened 2019-11-22
Egeria release 1.2 fix

mandy-chessell:
This fix should just change OMRSMetadataCollectionBase at lines 5667 and 5921


mandy-chessell:
This is not right - pulled content merged from master

PR #2085, opened 2019-11-25
Asset lineage updated with OpenLineage Services

DimitriosMaimaris:
Functionality for building Main Graph from Buffer Graph without SchemaType optimization.

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/buffergraph/BufferGraphConnector.java
+++ b/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/buffergraph/BufferGraphConnector.java
@@ -6,7 +6,6 @@ import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSo
 import org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.__;
 import org.apache.tinkerpop.gremlin.structure.Edge;
 import org.apache.tinkerpop.gremlin.structure.Vertex;
-import org.apache.tinkerpop.gremlin.structure.io.IoCore;
 import org.janusgraph.core.JanusGraph;
 import org.odpi.openmetadata.accessservices.assetlineage.model.GraphContext;
 import org.odpi.openmetadata.accessservices.assetlineage.model.LineageEntity;
@@ -14,15 +13,12 @@ import org.odpi.openmetadata.accessservices.assetlineage.event.LineageEvent;
 import org.odpi.openmetadata.frameworks.connectors.ffdc.ConnectorCheckedException;
 import org.odpi.openmetadata.frameworks.connectors.properties.ConnectionProperties;
 import org.odpi.openmetadata.governanceservers.openlineage.buffergraph.BufferGraphConnectorBase;
-import org.odpi.openmetadata.governanceservers.openlineage.ffdc.OpenLineageException;
-import org.odpi.openmetadata.governanceservers.openlineage.model.GraphName;
 import org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.factory.GraphFactory;
 import org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.model.JanusConnectorErrorCode;
 import org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.model.ffdc.JanusConnectorException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/pom.xml
```diff
@@ -75,6 +75,13 @@
             <scope>compile</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-core</artifactId>
+            <version>5.2.1.RELEASE</version>
```
planetf1: The version should be removed here (and don't use hardcoded values). The version should be picked up automatically from top level pom's dependencyManagement

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
```diff
@@ -67,6 +67,19 @@
             <artifactId>repository-services-apis</artifactId>
         </dependency>
 
+
+        <!-- berkeley & lucene aren't compile dependencies, so need exception for maven dependency plugin -->
+        <dependency>
```
planetf1: are these the only supported options, or can a user configure with a different db/indexer? If so, I wonder if they should be hard dependencies... or rather should be provided as part of the install/environment setup - in this case they would not be dependencies at all, but it would be the responsibility of the deployer to ensure janus is configured.

it somewhat depends if you have a 'hidden, embedded; graph, or one you expect to be configured... since an enterprise is probably use cassandra, I think perhaps they shouldn;'t be a dependency>?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
```diff
@@ -87,6 +100,8 @@
                                 <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-es:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>commons-logging:commons-logging:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>commons-codec:commons-codec:*</ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-berkeleyje:*</ignoredUnusedDeclaredDependency>
```
planetf1: which is related to this of course

open-metadata-implementation/server-chassis/server-chassis-spring/pom.xml
```diff
@@ -330,6 +335,8 @@
                                 <ignoredUnusedDeclaredDependency>org.springframework.boot:spring-boot-starter-web:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>org.springframework.boot:spring-boot-starter-security:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>org.springframework.boot:spring-boot-starter-data-jpa:*</ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>org.odpi.egeria:janus-connector:*</ignoredUnusedDeclaredDependency>
```
planetf1: as above

open-metadata-implementation/server-chassis/server-chassis-spring/pom.xml
```diff
@@ -292,6 +292,11 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
```
planetf1: we'll end up pulling in so much here with lucene - which already has some security exposures - and end up infecting all our components?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
GuyAckermans: Hi Nigel, 

These dependencies should indeed by managed dynamically, I think Mandy and Graham are working on the architectural design to make this possible?

open-metadata-implementation/server-chassis/server-chassis-spring/pom.xml
GuyAckermans: This dependency should be removed

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/utils/EdgeLabelsMainGraph.java
```diff
@@ -0,0 +1,13 @@
...
+package org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.utils;
+
+public enum EdgeLabelsMainGraph {
+    EDGE_LABEL_COLUMN_AND_PROCESS,
+    EDGE_LABEL_TABLE_AND_PROCESS,
+    EDGE_LABEL_SEMANTIC,
+    EDGE_LABEL_GLOSSARYTERM_TO_GLOSSARYTERM,
+    EDGE_LABEL_SUBPROCESS_TO_PROCESS,
+    EDGE_LABEL_CONDENSED,
+    EDGE_LABEL_INCLUDED_IN
```
GuyAckermans: Already exists in GraphConstants but unless I am mistaken they are not being removed there?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/utils/GraphConstants.java
```diff
@@ -34,6 +22,18 @@
     public static final String PROPERTY_NAME_PARENT_PROCESS_GUID = "parent.process.guid";
     public static final String PROPERTY_NAME_GLOSSARY = "glossary";
 
+    public static final String PROPERTY_NAME_GUID                             = "guid";
```
GuyAckermans: Why split the list?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/utils/VertexLabelsMainGraph.java
```diff
@@ -0,0 +1,11 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.utils;
+
+public enum VertexLabelsMainGraph {
+    NODE_LABEL_COLUMN,
+    NODE_LABEL_TABLE,
+    NODE_LABEL_GLOSSARYTERM,
+    NODE_LABEL_PROCESS,
+    NODE_LABEL_SUB_PROCESS
```
GuyAckermans: Already exists in GraphConstants but unless I am mistaken they are not being removed there?

pom.xml
```diff
@@ -1942,11 +1942,11 @@
...
-            <dependency>
-              <groupId>com.sleepycat</groupId>
-              <artifactId>je</artifactId>
-              <version>18.3.12</version>
-            </dependency>
+<!--            <dependency>-->
+<!--              <groupId>com.sleepycat</groupId>-->
+<!--              <artifactId>je</artifactId>-->
+<!--              <version>18.3.12</version>-->
+<!--            </dependency>-->
```
GuyAckermans: Has it been confirmed with @planetf1 @grahamwallis that this dependency can be removed from Egeria root POM?

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/pom.xml
planetf1: If we need to make it a compile dependency now, perhaps one approach would be to open up an issue to track removing that hard dependency. like 'make janus berkeley/lucene loading dynamic' or similar? At least then we'll have an issue to follow ?

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/util/Validator.java
```diff
@@ -2,10 +2,16 @@
 /* Copyright Contributors to the ODPi Egeria project. */
 package org.odpi.openmetadata.accessservices.assetlineage.util;
 
+import com.sun.corba.se.spi.activation.RepositoryHelper;
```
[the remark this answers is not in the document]
planetf1: Good spot!!!! ++ agreed. We should NOT use com.sun

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/buffergraph/BufferGraphConnector.java
```diff
@@ -95,47 +96,85 @@ public void addEntity(LineageEvent lineageEvent){
...
-        List<String> guidList = vertices.stream().map(v -> (String) v.property(PROPERTY_KEY_ENTITY_GUID).value()).collect(Collectors.toList());
+            List<String> guidList = vertices.stream().map(v -> (String) v.property(PROPERTY_KEY_ENTITY_GUID).value()).collect(Collectors.toList());
 
-        for (String guid : guidList) {
-            Iterator<Vertex> initial =  g.V().has(PROPERTY_KEY_ENTITY_GUID,guid).has("displayName","initial_load");
-            if(!initial.hasNext()) {
+            guidList.stream().forEach(process -> findInputColumns(g,process));
+            g.tx().commit();
+        }catch (Exception e){
+            log.debug(e.getMessage());
```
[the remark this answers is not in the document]
planetf1: I think it depends on the purpose of what we are doing here.... If it affects the user org and they need to do something, or know about something then 100% agree - and we need to clearly explain what to do. Only use debug if it's purely.. well.. debug to understand the code operation - and even then we need to take care with logging exceptions to ensure they don't leak priviliged information

pom.xml
planetf1: Good point. The entries under dependencyManagement get used not just to express a dependency for our code, but also they affect transitive dependencies (ie dependencies of dependencies of dependencies etc). Since we use Janus for our graph repo as well as for open lineage, I rather suspect we need them too. They allow us to control what version is used (ie in terms of security fixes)

You can run 'mvn dependency:tree' to see if we reference this artifact at all. If not, it's safe to remove. If it is (which I think it is), then it's not.

Thanks for spotting @GuyAckermans

planetf1: Looks like this may have caused the verify build failure

PR #2092, opened 2019-11-26
Fix layout issues with LF Projects header and broken favicon

jmertic:


jmertic:
@mandy-chessell Try this PR instead ;-)

mandy-chessell:
The build is failing because the RAT check is failing - this means there are new files without propertly licenses in them (or not excluded)

PR #2102, opened 2019-11-26
Changes to the way unsupported functions are reported

grahamwallis:


cmgrote:
Testing locally and will update as soon as I have results (probably ~1 hour)...

cmgrote:
Yes, this looks good -- now seeing that current property searches are `CONFORMANT_FULL_SUPPORT` and metadata maintenance profile (where entity lifecycle resides) is `CONFORMANT_NO_SUPPORT`

planetf1:
Is this also expected for 1.2?

cmgrote:
I would suggest yes, as without these changes read-only repositories can never by conformant with the mandatory profile...

grahamwallis:
Yes definitely want this in 1.2

Issue #2054, opened 2019-11-21
Extensive INFO/WARNINGS from dependency check


planetf1:
The build is reporting a lot of WARNINGS early on in the build along the lines of
```[main] INFO org.apache.maven.cli.event.ExecutionEventLogger - --- maven-dependency-plugin:3.1.1:analyze-dep-mgt (depmgmt) @ egeria ---
[main] INFO org.apache.maven.plugins.dependency.analyze.AnalyzeDepMgt - Found Resolved Dependency/DependencyManagement mismatches:
[main] INFO org.apache.maven.plugins.dependency.analyze.AnalyzeDepMgt - 	None
[main] WARNING org.codehaus.plexus.PlexusContainer - The following dependencies could not be resolved at this point of the build but seem to be part of the reactor:
[main] WARNING org.codehaus.plexus.PlexusContainer - o org.odpi.egeria:open-connector-framework:jar:1.3-SNAPSHOT (compile)
[main] WARNING org.codehaus.plexus.PlexusContainer - Try running the build up to the lifecycle phase "package"
[main] WARNING org.codehaus.plexus.PlexusContainer - The following dependencies could not be resolved at this point of the build but seem to be part of the reactor:
[main] WARNING org.codehaus.plexus.PlexusContainer - o org.odpi.egeria:open-connector-framework:jar:1.3-SNAPSHOT (compile)
[main] WARNING org.codehaus.plexus.PlexusContainer - Try running the build up to the lifecycle phase "package"
```
for every component
followed by many
```
[main] INFO org.owasp.dependencycheck.maven.AggregateMojo - Unable to resolve org.odpi.egeria:open-connector-framework:1.3-SNAPSHOT as it has not been built yet - creating a virtual dependency instead.
[main] INFO org.owasp.dependencycheck.maven.AggregateMojo - Unable to resolve org.odpi.egeria:repository-services-apis:1.3-SNAPSHOT as it has not been built yet - creating a virtual dependency instead.
[main] INFO org.owasp.dependencycheck.maven.AggregateMojo - Unable to resolve org.odpi.egeria:***-services-api:1.3-SNAPSHOT as it has not been built yet - creating a virtual dependency instead.
[main] INFO org.owasp.dependencycheck.maven.AggregateMojo - Unable to resolve org.odpi.egeria:***-services-registration:1.3-SNAPSHOT as it has not been built yet - creating a virtual dependency instead.
[...]

PR #2107, opened 2019-11-26
#2054 resolve some lifecycle/module errors in poms

planetf1:

A few issues found in poms with poor lifecycle management -- ie when a plugin is being invoked in the wrong phase, which means the required resources may not be in place.

This is only seen when running lifecycles other than 'mvn clean install' but the consequence is that adding new plugins leads to unpredictable behaviour - for example artifacts may not be found (as they are being resolved in too early a phase) or in the worst case, the wrong artifacts may be used - from a previous version

Tested by running a maven against each of these in turn:

pre-clean clean post-clean validate initialize generate-sources process-sources generate-resources process-resources compile process-classes generate-test-sources process-test-sources generate-test-resources test-compile process-test-classes test prepare-package package pre-integration-test integration-test post-integration-test verify install
Having cleaned up ~/.m2 between each iteration.

only tested from the top level module
only tested with default profiles

See #2054 for more info

### DIFF
--- a/open-metadata-conformance-suite/pom.xml
+++ b/open-metadata-conformance-suite/pom.xml
@@ -41,7 +41,7 @@
                 <executions>
                     <execution>
                         <id>copy-dependencies</id>
-                        <phase>prepare-package</phase>
+                        <phase>package</phase>
                         <goals>
                             <goal>copy-dependencies</goal>
                         </goals>


planetf1:
A summary of the code changes:

* Build Polymer UI in compile phase - same as other code (not process-resources)
* copy the UI resources as part of packaging (not generate-resources)
* Build the FVT open types code in compile - not generate-sources
* Create dummy source/javadoc in packaging phase (not process-resources)
* Move the jacoco setup from initalize to process-test resources - only needed for tes
ting.

The issue with the prior phase assignments is whilst they might have sounded ok,
they break down in a cross module project since we will build each module in
order only up to the phase given - maven does not build each module to phase a, then b
, then c - so as soon as there are cross-module dependencies things get out of sync

planetf1:
Identified one further issue when building in another environment - so leave as draft. commit to be added.

PR #2109, opened 2019-11-27
1.2 release - CTS fixes backport

planetf1:
Backport of recent CTS fixes by @cmgrote @grahamwallis

Commits:
546eb9d
19544e2
ab5f6f8
a4ac408
8446f6f
f200e16
2118133
b18ff2f
8248f6d
afb65f6
a228b38
2118133
(the last one was already merged in)

Please review

### DIFF
--- a/open-metadata-resources/open-metadata-labs/conformance-testing-labs/run-conformance-test-suite.ipynb
+++ b/open-metadata-resources/open-metadata-labs/conformance-testing-labs/run-conformance-test-suite.ipynb
@@ -72,7 +72,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 22,
    "metadata": {},
    "outputs": [
     {
@@ -127,7 +127,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 23,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -141,7 +141,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": 24,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -167,7 +167,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 4,
+   "execution_count": 25,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


planetf1:
Awaiting a further change from @grahamwallis

open-metadata-implementation/repository-services/repository-services-apis/src/main/java/org/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/OMRSMetadataCollectionBase.java
```diff
@@ -5868,7 +5869,7 @@ public  void deleteRelationshipReferenceCopy(String         userId,
         /*
          * Delete relationship
          */
-        reportUnsupportedOptionalFunction(methodName);
+        super.deleteRelationshipReferenceCopy(userId, relationship);
```
cmgrote: @grahamwallis these lines where we have replaced the `reportUnsupportedOptionalFunction()` method call with a super-class call to the method itself look suspect to me:

- will this mean repositories that have not overridden these methods will no longer automatically throw the FNS exception?
- why have we replaced these calls only for the reference copy operations, and none of the others?

grahamwallis:
Hi Chris - it looks like Mandy wanted the super class to delegate to the actual method (i.e save rather than delete) which if implemented will save to the repository and if not implemented will report unsupportedOptionalFunction. Maybe this is to ensure that the base class provides the switch (from delete to save).

cmgrote:
Ok, interesting (and a bit confusing without documentation in the class explaining why it's different)...  In any case, I've just kicked off a local run with this entire PR built as-is to do a final check that everything still operates as expected.

planetf1:
Added

f4fb9503456fa1197af6ba09ad1b77c0e509b19e
620a173edda7970ec2b9e10460cf557eddf7e626
e13be8018694a97bf9822c2448b8e1e363390902

This covers a reporting issue, and the notebook (as it makes it easier to test)

planetf1:
Leaving as draft - will be interesting to hear of your results @cmgrote. If good will look at merging - and I'll also rerun the cts against inmem/graph with the 1.2 branch

cmgrote:
Still looks fine from my perspective after testing overnight.

PR #2134, opened 2019-11-29
Issue 2100

grahamwallis:


grahamwallis:
This PR should boost performance for the read-only portion of the CTS property search tests.

grahamwallis:
The approach taken is to perform positive and negative searches with one of the discovered values for each property found in the discovery set. It then performs dual property tests using one pair of properties chosen from the discovery set. This is used to verify behaviour of match criteria ANY, ALL, NONE. It then performs propertyValue searches (using searchCriteria string) based on one string property from the discovery set, using all of its discovered values. This last test is repeated for EXACT, PREFIX, SUFFIX & CONTAINS matches.

grahamwallis:
Closed as wanted to address the case of instances matching on multiple properties' values - should only add to expectedGUIDs if not already present.

Issue #1936, opened 2019-11-01
Asset catalog get supported types


bogdan-sava:
create API rest endpoint that returns types supported by asset-catalog
create method in client

gueryParameter: name - used to filter all supported types by name or displayName, that contains the parameter. Parameter is not mandatory.

danielaotelea:
Asset Catalog OMAS configuration should provide a list of: 
SupportedTypes and DefaultTypes. These will be added on accessServiceOptions. 
The defaulted types for the AC OMAS are: GlossaryTerm, Asset, SchemaElement. 

The second task is to provide a REST endpoint that returns the supported types. 
@bogdan-sava is it enough for this service to return the typeName and typeGUID?

danielaotelea:
Hi, Bogdan! There is something that is missing from the Asset Catalog OMAS related to the supported types? 
If not, can we close this issue?

PR #2140, opened 2019-12-02
#1936 Asset catalog get supported types 

danielaotelea:
- service added to fetch the supported types
- configure supported types for the AC OMAS using SupportedTypesForSearch in the admin config
- client updated for the new service
- set required on false for types in REST endpoints


open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/CommonHandler.java
```diff
@@ -79,15 +92,113 @@ public String getTypeDefGUID(String userId, String typeDefName) {
...
+     * @return entity detail object
+     * @throws InvalidParameterException  one of the parameters is null or invalid.
+     * @throws UserNotAuthorizedException user not authorized to issue this request.
+     * @throws PropertyServerException    problem retrieving the entity.
+     */
+    public EntityDetail getEntityByGUID(String userId,
+                                        String guid,
+                                        String type) throws InvalidParameterException, UserNotAuthorizedException, PropertyServerException {
+
+        String entityTypeName = type == null ? "Unknown" : type;
```
davidradl: type is supplied but does not seem to be used. We should either use the type or remove it from the method signature.

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/CommonHandler.java
```diff
@@ -79,15 +92,113 @@ public String getTypeDefGUID(String userId, String typeDefName) {
...
+        String methodName = "getEntityByGUID";
+        try {
+            return repositoryHandler.getMetadataCollection().getEntityDetail(userId, guid);
+        } catch (org.odpi.openmetadata.repositoryservices.ffdc.exception.InvalidParameterException e) {
+            e.printStackTrace();
```
davidradl: we should not be using stack trace here. This invalid parameter should be passed back to the caller.

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/CommonHandler.java
```diff
@@ -137,7 +137,7 @@ public EntityDetail getEntityByGUID(String userId,
         try {
             return repositoryHandler.getMetadataCollection().getEntityDetail(userId, guid);
         } catch (org.odpi.openmetadata.repositoryservices.ffdc.exception.InvalidParameterException e) {
-            e.printStackTrace();
+            errorHandler.handleRepositoryError(e, methodName);
```
davidradl: I suggest we check to see if the invalid parameter is the guid and then call errorHandler.handleUnknownEntity - so we do not lose this insight.

Issue #2137, opened 2019-11-30
Release 1.2 - azure pipeline failure in creating release


planetf1:
Pipeline run to create release 1.2
->

```
2019-11-29T15:48:17.6282389Z ##[section]Starting: Artifactory Build Promotion - Dry Run
2019-11-29T15:48:17.6580726Z ==============================================================================
2019-11-29T15:48:17.6580982Z Task         : Artifactory Build Promotion
2019-11-29T15:48:17.6581132Z Description  : Promote a published build in Artifactory.
2019-11-29T15:48:17.6581234Z Version      : 1.8.1
2019-11-29T15:48:17.6581362Z Author       : JFrog
2019-11-29T15:48:17.6581487Z Help         : 
2019-11-29T15:48:17.6581588Z ==============================================================================
2019-11-29T15:48:19.3364601Z Downloading: https://api.bintray.com/content/jfrog/jfrog-cli-go/1.31.2/jfrog-cli-linux-amd64/jfrog?bt_package=jfrog-cli-linux-amd64
2019-11-29T15:48:20.5584579Z Caching tool: jfrog 1.31.2 x64
2019-11-29T15:48:20.6048791Z Running jfrog-cli from /opt/hostedtoolcache/jfrog/1.31.2/x64/jfrog.
2019-11-29T15:48:20.6137639Z JFrog CLI version: 1.31.2
2019-11-29T15:48:20.6791578Z [Info] [Dry run] Promoting build...
2019-11-29T15:48:21.2353055Z [Error] Artifactory response: 400 Bad Request
2019-11-29T15:48:21.2354659Z {
2019-11-29T15:48:21.2355019Z   "errors": [
2019-11-29T15:48:21.2355260Z     {
2019-11-29T15:48:21.2355355Z       "status": 400,
2019-11-29T15:48:21.2356202Z       "message": "Unable to find artifacts of build 'Egeria-Stage' #a1ce3b2bf: aborting promotion."
2019-11-29T15:48:21.2356378Z     }
2019-11-29T15:48:21.2356455Z   ]
2019-11-29T15:48:21.2356571Z }
2019-11-29T15:48:21.2437529Z ##[error]Error: Command failed: /opt/hostedtoolcache/jfrog/1.31.2/x64/jfrog rt bpr "Egeria-Stage" "a1ce3b2bf" "egeria-staging-local" --url="https://odpi.jfrog.io/odpi" --user=*** --password=*** --include-dependencies=false --copy=false --dry-run=true
[...]

PR #2153, opened 2019-12-03
Remove egeria-release from triggers

jwagantall:
This is in order to fix the pending release
being worked on in IT-18384

Signed-off-by: Jessica Wagantall <jwagantall@linuxfoundation.org>


planetf1:
See #2137

Issue #2081, opened 2019-11-25
SonarScans failing since 1 Nov (Out of memory)


planetf1:
Sonar scans at https://sonarcloud.io/organizations/odpi/projects last ran on 1 Nov
Checking an example build at https://dev.azure.com/odpi/Egeria/_build/results?buildId=2281&view=results shows that the build failed due to

Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "Report about progress of Java AST analyzer"

Additionally the PMD results are not found
[ERROR] Can't find PMD XML report: /home/vsts/work/1/s/open-metadata-implementation/common-services/ocf-metadata-management/ocf-metadata-client/target/pmd.xml

That's not a surprise, since this is a PR build.
Additionally whilst scanning a PR could be useful, the way sonarcloud tracks results isn't helpful - we just see a continuous stream of stats from build to build, and the more erratic nature of PRs (some fail for example) will just cause confusion

The build first failed at https://dev.azure.com/odpi/Egeria/_build/results?buildId=1688&view=logs&j=f38f3889-b668-5b4c-0263-c9548e3b1185&t=4606f1f8-98d1-5773-4e52-dcf0da0b4452&l=371 which was #1941 although #1846 made a change to the configuration data which perhaps was only picked up following the build.

So to look at
 * Why is PMD not running
 * Why are we running out of memory (maybe this causes the lack of pmd files!)

cong78:
Our ING SonarQube scan is always running out of memory for our project even the service team helped us to increase the memory. :P

bramwelt:
I'm not sure my change introduce the sonar failures, but from my limited search of [SO](https://stackoverflow.com/a/20481827) I'm wondering if tweaking the inclusion/exclusion rules for Sonar might help - As it seems Sonar falls over when it tries to pull in files larger than it can handle.

A couple of notes:
1. The Azure builds aren't setup to run against PRs - these are sonar scan failures on Master.
2. PMD doesn't run because we're only running `sonar:sonar` for the sonar step while passing `-DfindBugs` which looks to requires the `verify` phase. A simple fix would be to include `pmd:pmd` and/or move `-DfindBugs` to the install step just before sonar.
3. It could be a simple matter of tweaking the memory given to the JVM process: the Azure runners provided 7Gb which seems like it'd be more than enough for code scanning...

planetf1:
As a first step, I propose to fix the base sonar scan - without pmd/spotbugs

[...]

PR #2177, opened 2019-12-06
#2081 remove findbugs/pmd, switch scheduling to batch

planetf1:

2 changes

 a) findbugs/spotbugs is still reporting an error - due to a java 8 compatibility issue in the plugin (seemingly). As per the original intent of 2081 I want to get regular sonar working reliably, so have removed the -Dfindbugs and addition of the pmd/spotbugs xml into the scan - there is other refactoring to be done in this area in any case, as well as maven site reports with xfref code which will provide the same data

 b) The sonar scans take longer than a regular build. This was especially true with pmd etc. The impact was to potentially cause other builds to queue - and sonar results aren't constantly looked at. One option was to move to daily, but instead I've set batch=true in the job yaml which just means that if a job is already underway we won't overlap - see https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml -- this may also be useful in other pipelines

### DIFF
--- a/.azure-pipelines/sonar.yml
+++ b/.azure-pipelines/sonar.yml
@@ -2,6 +2,7 @@
 # Copyright Contributors to the ODPi Egeria project.
 ---
 trigger:
+  batch: true
   branches:
     include:
       - master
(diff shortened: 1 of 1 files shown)


planetf1:
As a side effect it also brings the reported results more closely aligned to when the sonar cloud plugin is used locally in intellij.

Issue #2090, opened 2019-11-26
CTS: move findRelationship... methods to optional profile


cmgrote:
Currently the `findRelationship...` methods are part of the mandatory profile.  As discussed and agreed on 26/11 call we will move these methods to an optional profile in order to reduce the burden on simple repositories that may not have first-class support for relationships (and therefore find it very difficult or impossible to support these methods).

grahamwallis:
Fixed by #2185

PR #2185, opened 2019-12-09
Makes findRelationship an optional CTS profile

grahamwallis:
Essentially what this PR does is to split up the search tests so that:

For the 'basic' searches (i.e. repo helper literalised expressions)
- Entity Property & Value Search are (still) part of METADATA_SHARING (mandatory)
- Relationship Property & Value Search moves into a new RELATIONSHIP_RELATIONSHIP_SEARCH (optional)

For the 'advanced' searches (arbitrary regex expressions) these are now split between:
- Entity Advanced Property & Value Searches are in ENTITY_ADVANCED_SEARCH (optional)
- Relationship Advanced Property & Value Searches are in RELATIONSHIP_ADVANCED_SEARCH (optional)

So it is possible to only support basic entity searches, or to do that and optionally add any of basic relationship search, advanced entity search and/or relationship advanced search.


grahamwallis:
Fixes #2090

PR #2191, opened 2019-12-10
Data Engine OMAS - fix issues

popa-raluca:
Bug fixing for DE OMAS:
- fixes for removing ports/schema types on update
- use org.apache.commons dependencies for collections and strings utility methods
- add debug logging
- update endpoint for adding ports to process based on qualified names
- update unit tests

### DIFF
--- a/open-metadata-implementation/access-services/asset-catalog/asset-catalog-api/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/model/Classification.java
+++ b/open-metadata-implementation/access-services/asset-catalog/asset-catalog-api/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/model/Classification.java
@@ -36,8 +36,7 @@ public class Classification implements Serializable {
     private Long version;
     private String status;
 
-    private String typeDefName;
-    private String typeDefDescription;
+    private Type type;
 
     private Map<String, Object> properties;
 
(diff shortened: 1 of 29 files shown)


open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/handlers/ProcessHandler.java
```diff
@@ -325,7 +328,8 @@ public void updateProcessStatus(String userId, String guid, InstanceStatus insta
             return new HashSet<>();
         }
 
-        return entities.parallelStream().map(InstanceHeader::getGUID).collect(Collectors.toSet());
+        return entities.parallelStream().filter(entityDetail -> entityDetail.getType().getTypeDefName().equalsIgnoreCase(portTypeName))
```
cong78: Not sure if I understand why we need portType as a parameter here from process for extra checking?

popa-raluca: Because I want to retrieve only the port implementations/port aliases for a certain process. repositoryHandler.getEntitiesForRelationshipType will retrieve all the entitites for a ProcessPort relationship that have the first end the processGUID, and I need the result filtered based on the actual type

PR #2197, opened 2019-12-11
Code generated deprecations

wbittles:
@planetf1 as you suggested I increased the size of my search buffer and this PR should remove around 230 warnings.


planetf1:
@wbittles I've just re-queued your java11 PR build. It failed very oddly on the user interface build - but when extracting npm related artifacts. No clear network error. Perhaps a resource problem. Never seen before, Hopefully a one off. If not I'll try to investigate further. I don't believe it's your change causing the issue

davidradl:
U updated master and the java 11 build failed. It looks like a piple line issue not a code issue @planetf1 @wbittles

planetf1:
> I updated master and the java 11 build failed. It looks like a pipeline issue not a code issue @planetf1 @wbittles

It failed before your change. See earlier post. I'll investigate but later today

planetf1:
The rebuild worked.... Something funny with the npm/UI build. Probably relating to the sources service being down. There may be further things we can cache in artifact to improve reliability. Will investigate

I've merged the change

Issue #2179, opened 2019-12-09
AssetElement in Asset search API


davidradl:
I notice that the Asset search OMAS exposes a concept called AssetElement, a list of these are returned to the UI after a search.

1) I am not sure what AssetElement is as it does not have any javadoc, it subclasses Element, but is referred to in the UI methods as asset header. I am not sure what context and connection means here.
2) Element has 
"    private static final long serialVersionUID = 1L;

    private String guid;
    private String typeDefName;
    private String typeDefGUID;
    private String metadataCollectionId;
    private String name;
    private String createdBy;
    private Date createTime;
    private String updatedBy;
    private Date updateTime;
    private Long version;
    private String status;
    private String url;
    private Map<String, Object> properties;
    private List<Element> parentElement;"

I am thinking that metadataCollectionId and typeDefGUID should not be exposed to the caller of the OMAS - these are internal things. I think display name and qualified name could be included - I am not sure which name is. I am not sure what the url is.
3) I assume that AssetElement is an Element associated with an asset typedef. Is this correct? So would not include glossary terms.
4) parentElement is singular and points to a List. I am not sure what parent this means here. does parent mean anything here in terms of ownership / containment?
5) I see there is Element with subclasses of AssetElement (which has context and connections) and AssetDescription (which has classifications and relationships). I assume that we do not need to have Asset connections ,context , relationships and classifications all together. I assume that AssetElement is the searchresult (rename to AssetSearchResult?) and AssetDescription is the details of the Asset (maybe rename to AssetDetails so as not to be confused with the text description associated with the asset).
@bogdan-sava

bogdan-sava:
I Agree

danielaotelea:
Hello, David! Thank you for your review!

1. I will update the JavaDoc 
2. In the last PR for the Asset Catalog OMAS I removed the typeDefGUID. 
3. Maybe the name of this POJO is not correct. The initial design of the Asset Catalog was to search for assets, since then we have added some other types for search as: Glossary Terms and Schema Elements. I will try to rename the class with something meaningful. The response of the asset catalog OMAS includes the Glossary terms currently.
4. The parentElement is used to build the context of the given element. The list can be removed, I will investigate.
5. The connection contains the connection identifier and qualified name. Those will be used by UI to get the connection (using Asset Consumer) in order to connect to the asset. This connection is empty for the Glossary Terms or other types returned by the Asset Catalog OMAS.
[...]

PR #2204, opened 2019-12-11
#2179 Asset Catalog OMAS search API

danielaotelea:
1. removed the metadata collection id from response
2. parent element is only one attribute instead of a list
3. javadoc added for Asset Element and Element


open-metadata-implementation/access-services/asset-catalog/asset-catalog-api/pom.xml
```diff
@@ -43,5 +43,11 @@
             <artifactId>validation-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>open-connector-framework</artifactId>
+            <version>1.3-SNAPSHOT</version>
```
planetf1: This version should be removed

danielaotelea: Indeed. I removed it. Thank you, Nigel!

planetf1: thanks

PR #2206, opened 2019-12-12
Fix spring updates - data, boot, spring, security have differing lifecycles

planetf1:
This addresses #2178 proposed by dependabot - for which the compile fails

The reason for this is primariy that spring-boot, spring-web, spring-security, spring-data are different projects (and have distinct groupids) and have slightly differerent lifecycles. They do express both a minimum and supported dependency, and may lag by around 1 . version.

This PR splits out the versions by group id so that they can be managed independently - which should also help dependabot get things correct in future. 


planetf1:
H2 update now works in conjunction with these latest changes (it didn't before), so merging together

tested usual notebooks, cohort, and login to UI, also /h2 web console is available. No errors on stdout.

PR #2216, opened 2019-12-12
Replace @RequestMapping\(method\s=\sRequestMethod.GET, with GetMapping\(

wbittles:
Signed-off-by: William Bittles <wbittles@uk.ibm.com>


open-metadata-conformance-suite/open-metadata-conformance-suite-spring/src/main/java/org/odpi/openmetadata/conformance/server/spring/ConformanceSuiteResource.java
```diff
@@ -12,6 +12,7 @@
 import org.springframework.web.bind.annotation.RequestMapping;
```
planetf1: I've not tested, but I wonder if we should remove the RequestMapping - ie clean up the imports, since we're changing to the get/put etc mappings?

wbittles: So this fix only removes the the Get's so the RequestMapping are needed for the other HTTP verbs. Post, Put etc. When I remove all those the last PR will remove the 
import org.springframework.web.bind.annotation.RequestMapping;
when it's no longer needed ?

planetf1: perfect! thanks for clarifying!

Issue #2214, opened 2019-12-12
Remove warning from Sonar Scans caused by Spring Annotations


wbittles:
The are currently 671 occurrences of @RequestMapping usage throughout the Egeria code base (regex @RequestMapping\(method\s=\sRequestMethod) that are generating warning from the Sonar Scans. This is issue is intended to replace @RequestMapping calls with the more up to date @GetMapping , @PostMapping , @DeleteMapping ,@PutMapping

planetf1:
It's an interesting example of some of the more minor issues we have. It's tagged as a code smell/minor. I agree it's clearer to use the new forms, but @requestmapping itself isn't deprecated.

We may have two other options to consider for all such issues
 a) We can triage and close the issues as 'wont fix' or 'false positive' - this is fine for the odd warning, but painful for many
 b) We may be able to change the profile used to not detect the error

That all being said, if it's simple to fix then the easiest course I agree is to fix (though there are more serious issues we also need to get sorted - bugs & security).

wbittles:
It looks like a pretty straight forward change, I've opened a PR with the changes for the @GetMapping case (220 instances) , please feel free to reject it if there there is a preference for another approach. Point taken on the more severe issues.

planetf1:
The main merges have gone in ... are you planning more changes  - if so we can keep open?

wbittles:
Hi Nigel I'm not planning on putting any more in until after the code freeze. Should I close the PR's ?

planetf1:
The PRs are closed.
If this issue is incomplete I suggest keeping it open (and summarising what area needs looking at)
Otherwise let's close it

mandy-chessell:
I think it is ok to close - I have updated the new REST APIs I added since @wbittles did the change.  I like it - the new annotations are easier to read.

PR #2224, opened 2019-12-15
Spring mappings

wbittles:
This PR converts all matches of @RequestMapping\(method\s=\sRequestMethod.POST,  with calls to @PostMapping Issue #2214

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <logback.version>1.2.3</logback.version>
         <jackson.version>2.10.1</jackson.version>
         <jackson.databind.version>2.10.0</jackson.databind.version>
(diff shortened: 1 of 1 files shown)


wbittles:
@mandy-chessell I've been removing the @RequestMapping calls in bitesize chunks, after this PR there are 67 occurrences still remaining in the code base, which will remove the remaining HTTP verbs and then the import statement that will be no longer needed ?

planetf1:
> Is it possible to remove the obsolete import of the ReguestMapping module?

I made the same comment on an earlier PR, I agree, but it's already planned ... Splitting the changes up does make sense, more manageable. 

Inspecting changes looks good. just doing a quick test

Issue #2135, opened 2019-11-29
Exception in AssetCatalog when searching for csv resources


planetf1:
If, having run the building-a-data-catalog notebook I try and search for the file `file://secured/research/clinical-trials/drop-foot/DropFootMeasurementsWeek1.csv` using it's fully qualified name (as written) in the Asset Search UI I get exceptions including:
```
2019-11-29 14:57:25.898 ERROR 33714 --- [nio-8443-exec-4] o.o.o.u.u.s.s.AssetCatalogOMASService    : Error searching the assets by criteria file://secured/research/clinical-trials/drop-foot/DropFootMeasurementsWeek1.csv
2019-11-29 14:57:25.920  WARN 33714 --- [nio-8443-exec-4] .m.m.a.ExceptionHandlerExceptionResolver : Failure in @ExceptionHandler org.odpi.openmetadata.userinterface.uichassis.springboot.api.ExceptionHandlerController#handleAssetCatalogException(AssetCatalogException, WebRequest)

java.lang.IllegalStateException: Could not resolve parameter [0] in protected org.springframework.http.ResponseEntity<java.lang.Object> org.odpi.openmetadata.userinterface.uichassis.springboot.api.ExceptionHandlerController.handleAssetCatalogException(org.odpi.openmetadata.accessservices.assetcatalog.exception.AssetCatalogException,org.springframework.web.context.request.WebRequest): No suitable resolver
	at org.springframework.web.method.support.InvocableHandlerMethod.getMethodArgumentValues(InvocableHandlerMethod.java:164) ~[spring-web-5.2.1.RELEASE.jar!/:5.2.1.RELEASE]
	at org.springframework.web.method.support.InvocableHandlerMethod.invokeForRequest(InvocableHandlerMethod.java:134) ~[spring-web-5.2.1.RELEASE.jar!/:5.2.1.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.ServletInvocableHandlerMethod.invokeAndHandle(ServletInvocableHandlerMethod.java:106) ~[spring-webmvc-5.2.1.RELEASE.jar!/:5.2.1.RELEASE]
	at org.springframework.web.servlet.mvc.method.annotation.ExceptionHandlerExceptionResolver.doResolveHandlerMethodException(ExceptionHandlerExceptionResolver.java:403) ~[spring-webmvc-5.2.1.RELEASE.jar!/:5.2.1.RELEASE]
```
[...]

PR #2228, opened 2019-12-16
#2135 Exception in AssetCatalog when searching for csv resources

danielaotelea:


danielaotelea:
Hi, Nigel!
In order to understand more about this asset, you use another service: /asset-context/{assetGUID}. This part is not integrated to the UI yet. Also, from this search, not all the properties are displayed in the UI. I will address this problems in another issue, if you agree.

planetf1:
Yes I agree - thanks. This was just the first step to clear up the exceptions... UI dev is still work in progress of course

PR #2241, opened 2019-12-17
Temp fix for freezing UI

GuyAckermans:


GuyAckermans:
I think the problem here is two-fold: 

1 The asset search is currently throwing an exception.
2 Any uncaught exceptions thrown by the asset search can cause the UI platform to become unavailable because while the spinner is there, the user is unable to even navigate to the UI views that do work. Ideally I think we should have a high level, general UI platform exception catcher which allows the user to still navigate to other UI tabs even though a specific tab may thow an unhandled exception.

I agree my workaround is a bad approach long-term, but right now it is the only way for me to be able to navigate to other views than Asset Search.

PR #2262, opened 2019-12-17
Fixes invalid type references for connector with subset of types

### LINKED ISSUE #2213: CTS NPEs on relationship endpoint EntityDef lookups
Came across a number of NullPointerExceptions in the (master) CTS as follows: under the `run()` method of the test classes below, the endpoint definitions of a relationship are retrieved but `repositoryHelper.getTypeDefByName(...)` fails with an NPE when the type definition is a supertype that is not implemented by the repository being tested. (For example, `ActorProfile` when the repository only implements `Person`; `Asset` when the repository only implements `Database`; `ComplexSchemaType` when the repository only implements `TabularSchemaType`, etc.)

Seems like another case where we should be retrieving the endpoint definitions of the relationship from the open type definitions themselves -- not from the repository under test?

Seen in all of the following:

- [ ] `TestSupportedRelationshipLifecycle`
- [ ] `TestSupportedRelationshipReidentify`
- [ ] `TestSupportedRelationshipPropertySearch`
- [ ] `TestSupportedRelationshipPropertyAdvancedSearch`

grahamwallis:
Fixes #2213

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
@@ -361,15 +361,14 @@ public class TestSupportedRelationshipSharingPropertySearch extends RepositoryCo
                     PrimitivePropertyValue ppv = new PrimitivePropertyValue();
                     ppv.setPrimitiveDefCategory(propertyCatMap.get(attributeName));
                     if (propertyCatMap.get(attributeName) == OM_PRIMITIVE_TYPE_STRING) {
-                        String literalisedValue = literaliseStringProperty((String)value);
+                        String literalisedValue = literaliseStringProperty((String) value);
                         ppv.setPrimitiveValue(literalisedValue);
-                    }
-                    else {
+                    } else {
                         ppv.setPrimitiveValue(value);
                     }
 
                     ppv.setPrimitiveValue(value);
-                    matchProperties.setProperty(attributeName,ppv);
+                    matchProperties.setProperty(attributeName, ppv);
 
                     /*
                      * Expected result size
@@ -381,15 +380,15 @@ public class TestSupportedRelationshipSharingPropertySearch extends RepositoryCo
                      */
 
                     result = metadataCollection.findRelationshipsByProperty(workPad.getLocalServerUserId(),
-                                                                           relationshipDef.getGUID(),
-                                                                           matchProperties,
-                                                                           MatchCriteria.ALL,
-                                                                           fromElement,
-                                                                          null,
-                                                                          null,
-                                                                          null,
-                                                                          null,
-                                                                           pageSize);
+                                                                            relationshipDef.getGUID(),
+                                                                            matchProperties,
+                                                                            MatchCriteria.ALL,
+                                                                            fromElement,
+                                                                            null,
+                                                                            null,
+                                                                            null,
+                                                                            null,
+                                                                            pageSize);
 
                     /*
                      * The approach to checking results match expectations is as follows:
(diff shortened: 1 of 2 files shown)


open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipPropertySearch.java
```diff
@@ -264,22 +264,22 @@ protected void run(TestPhase phase) throws Exception
                 RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getRequirementId());
 
         String end1TypeDefName = relationshipDef.getEndDef1().getEntityType().getName();
+        EntityDef knownEnd1EntityDef = (EntityDef) repositoryHelper.getTypeDefByName(workPad.getLocalServerUserId(), end1TypeDefName);
         EntityDef end1EntityDef = entityDefs.get(end1TypeDefName);
```
cmgrote: This line seems to be the problematic one -- if the `end1TypeDefName` is for example `Asset` then the line before it will work fine (drawing from the open type definitions themselves) but this line (drawing from the entities that are implemented in a connector that does not implement the more abstract supertypes) will simply return `null` and the assertion on the very next line will therefore always fail since `end1EntityDef == null`.

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipPropertySearch.java
```diff
@@ -264,22 +264,22 @@ protected void run(TestPhase phase) throws Exception
...
-                RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getRequirementId());
+        verifyCondition( (end1EntityDef != null) && end1EntityDef.equals(knownEnd1EntityDef),
+                         assertion0,
+                         testTypeName + assertionMsg0,
+                         RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getProfileId(),
+                         RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getRequirementId());
 
         String end2TypeDefName = relationshipDef.getEndDef2().getEntityType().getName();
+        EntityDef knownEnd2EntityDef = (EntityDef) repositoryHelper.getTypeDefByName(workPad.getLocalServerUserId(), end2TypeDefName);
         EntityDef end2EntityDef = entityDefs.get(end2TypeDefName);
```
cmgrote: As above -- if the connector does not implement supertypes, this will return `null` for many relationships and the assertion just below will always fail as a result.

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
```diff
@@ -194,28 +197,24 @@ protected void run() throws Exception
...
-        verifyCondition((endTwoEntityDef.equals(knownEnd2EntityDef)),
-                assertion0,
-                testTypeName + assertionMsg0,
-                RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getProfileId(),
-                RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getRequirementId());
-
 
+        String end1TypeDefName = relationshipDef.getEndDef1().getEntityType().getName();
+        EntityDef knownEnd1EntityDef = (EntityDef) repositoryHelper.getTypeDefByName(workPad.getLocalServerUserId(), end1TypeDefName);
+        EntityDef end1EntityDef = entityDefs.get(end1TypeDefName);
```
cmgrote: As above -- if the connector does not implement supertypes, this will return `null` for many relationships and the assertion just below will always fail as a result.

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSharingPropertySearch.java
```diff
@@ -194,28 +197,24 @@ protected void run() throws Exception
...
+        verifyCondition( (end1EntityDef != null) && end1EntityDef.equals(knownEnd1EntityDef),
+                         assertion0,
+                         testTypeName + assertionMsg0,
+                         RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getProfileId(),
+                         RepositoryConformanceProfileRequirement.CONSISTENT_TYPES.getRequirementId());
+
+        String end2TypeDefName = relationshipDef.getEndDef2().getEntityType().getName();
+        EntityDef knownEnd2EntityDef = (EntityDef) repositoryHelper.getTypeDefByName(workPad.getLocalServerUserId(), end2TypeDefName);
+        EntityDef end2EntityDef = entityDefs.get(end2TypeDefName);
```
cmgrote: As above -- if the connector does not implement supertypes, this will return `null` for many relationships and the assertion just below will always fail as a result.

grahamwallis: Please see additional commit "Refinements to relationship end type checks" which should resolve all the above. Sorry about that.

PR #2308, opened 2019-12-27
Open lineage schema optimization

DimitriosMaimaris:
Changes for NestedScehmaAttribute Relationship between Relational Columns and Relational Tables.


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/AssetContextHandler.java
```diff
@@ -134,20 +134,19 @@ private void buildAssetContext(String userId, EntityDetail entityDetail,String s
...
+            }
             for (EntityDetail schemaTypeEntity : tableTypeEntities) {
                 if (isComplexSchemaType(userId, schemaTypeEntity.getType().getTypeDefName())) {
                     setAssetDetails(userId, schemaTypeEntity);
                 } else {
-                    List<EntityDetail> tableEntities = buildGraphByRelationshipType(userId, schemaTypeEntity, SCHEMA_ATTRIBUTE_TYPE, typeDefName, false);
-                    if (!CollectionUtils.isEmpty(tableEntities)) {
-                        buildAssetContext(userId, tableEntities.stream().findFirst().get(), superType);
-                    }
+                        buildAssetContext(userId, tableTypeEntities.stream().findFirst().get(), superType);
```
popa-raluca: get() might return NoSuchElementException if the Optional is empty. 
Also, it doesn't look like the superType parameter is used anywhere in the method

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/ProcessContextHandler.java
```diff
@@ -78,6 +78,10 @@ public ProcessContextHandler(String serviceName,
             Optional<EntityDetail> entityDetail = commonHandler.getEntityDetails(userId, processGuid, PROCESS);
             if (!entityDetail.isPresent()) {
                 log.error("Entity with guid {} was not found in any metadata repository", processGuid);
+//                AssetLineageErrorCode errorCode = AssetLineageErrorCode.ENTITY_NOT_FOUND;
```
popa-raluca: You should remove the commented code if it doesn't make sense to have it anymore

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/ProcessContextHandler.java
```diff
@@ -103,8 +107,8 @@ public ProcessContextHandler(String serviceName,
                                                                                                                PropertyServerException,
                                                                                                                UserNotAuthorizedException {
 
-        List<EntityDetail> entitiesTillLastRelationship = getEntitiesLinkedWithProcessPort(userId, entityDetail);
-        if(!entitiesTillLastRelationship.isEmpty()){
+        boolean entitiesTillLastRelationshipExist = getEntitiesLinkedWithProcessPort(userId, entityDetail);
```
popa-raluca: Might be better if you rename getEntitiesLinkedWithProcessPort to hasEntititesLinkedWithProcessPort

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/ProcessContextHandler.java
```diff
@@ -180,68 +184,92 @@ public ProcessContextHandler(String serviceName,
...
-    
-    private List<EntityDetail> getTabularSchemaTypes(List<EntityDetail> entityDetails, String userId) throws InvalidParameterException,
+
+    /**
+     * Retrieves the TabularSchemaTypes that are related to a Port Implementation Entity.
+     * @param entityDetails      list of entities
+     * @param userId             String - userId of user making request.
+     * @return boolean true if relationships exist otherwise false.
+     */
+    private boolean getTabularSchemaTypes(List<EntityDetail> entityDetails, String userId) throws InvalidParameterException,
```
popa-raluca: Same comment for all the get.. methods that return a boolean. They should be renamed to follow the naming conventions. Also, the javadoc might be misleading

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/ProcessContextHandler.java
```diff
@@ -180,68 +184,92 @@ public ProcessContextHandler(String serviceName,
...
         List<EntityDetail> result = new ArrayList<>();
         for (EntityDetail entityDetail : entityDetails) {
-            result = getRelationshipsBetweenEntities(userId, entityDetail.getGUID(),
+            result.addAll(getRelationshipsBetweenEntities(userId, entityDetail.getGUID(),
                                            processRelationshipsTypes.get(entityDetail.getType().getTypeDefName()),
-                                           entityDetail.getType().getTypeDefName());
+                                           entityDetail.getType().getTypeDefName()));
         }
-        return result;
+        return result.isEmpty()? false: true;
```
popa-raluca: return statement might be simplified by using !result.isEmpty() instead

pom.xml
```diff
@@ -1950,11 +1950,11 @@
                 <scope>compile</scope>
             </dependency>
 
-            <dependency>
-              <groupId>com.sleepycat</groupId>
-              <artifactId>je</artifactId>
-              <version>${sleepycat.version}</version>
-            </dependency>
+<!--            <dependency>-->
```
popa-raluca: Why is this commented?

planetf1: Agreed.... this is likely causing your build failure as you are looking for an older version of sleepycat which is not in the public repos.

planetf1: This will affect other components too -- can you confirm what version of sleepycat are used after the pr using
'mvn dependency:tree' and look for all instances of sleepycat. I would expect them to be at ${sleepycat.version}.

We need this dependency to uplift the version of the dependency transitively, which also fixes a build problem seen outside of our official/local builds, due to missing artifacts on maven central.

planetf1: Ah looking at latest changes I see this is now corrected. thanks :-)

Issue #2324, opened 2019-12-31
connector type archive generated files need excluding from git


planetf1:
#2319 introduced the execution of a connector types archive builder within the build process.

After running a build the following new files were found in tracked git directories:
```
	DataStoreConnectorTypes.json
	DataStoreConnectorsGUIDMap.json
	open-metadata-resources/open-metadata-archives/open-connector-archives/DataStoreConnectorTypes.json
	open-metadata-resources/open-metadata-archives/open-connector-archives/DataStoreConnectorsGUIDMap.json
``` 
These should be removed from git tracking - either by exclusion, or by writing them into a target directory instead of source location.

planetf1:
Note that in the above I have the files in two locations.
I ran the build once from the root, but also ran it directly from 'open-connector-archives' - so it looks like it may be writing files into the current working directory.

In the build process (from within maven) we should ensure this is always written to a consistent location.

planetf1:
We can use
```
<workingDirectory>/tmp</workingDirectory>
```
to specify the cwd when executing the java tool. I am thinking ${project.build.directory} is most appropriate - it would be the 'target' directory for this module.

Currently this archive doesn't appear to be used anywhere -- couldn't find in the assemblies.  if/when it does we may need to revisit the appropriate location - (as well as the lifecycle phase the task is bound to)

mandy-chessell:
Where is this utility is run by the build?  I do not see the files appear in my build tree unless I run the utility directly.  I would like to run it as part of the build - just not sure of the best way to do it.

Once we are running it as part of the build (and the other archive utilities) then can add resulting archive files to the assemblies - and then use them in the labs :)

planetf1:
The utility was being run in open-metadata-resources/open-metadata-archives/open-connector-archives .. initially with maven exec.. now with antrun. However I don't think the results of the build are being used yet - as you say one requirement is that they are easily available to the notebook environment.

We start to get into #2215 here - I think we need to work through that in more detail -and of course ideally the tutorial environment should use the assemblies for all the egeria assets, that way we prove they are correct. 

This particular issue is resolved by #2325 so will close.

I did note that in #2329 entries were added to gitignore, but IMO my fix #2325 was a better option - we should aim to create files outside the core source tree - it's confusing, and in this case was kinda masking another issue, which was that the files were not always created in the same place depending on the users cwd, #2325 fixed that making the gitignore entries superfluous (and arguably having more than we need in there is in itself confusing ...)

PR #2325, opened 2019-12-31
#2324 correct connector archive type build proc to use consistent tar…

planetf1:
…get dir

Ensures current directory is ${project.build.dir} which is basically 'target' within the module, regardless of from where the executable was run - previously it would write to whatever current working directory was

As a side effect 'git status' no longer shows modified files after running the build, since target is a known exclusion

The implementation required a switch to use the maven antrun plugin -- just as we use for the UI -- since the exec plugin is not flexible enough to allow the working directory to be specified when running java

Final testing before out of draft..


planetf1:
See #2324

PR #2336, opened 2020-01-02
Resolves #2326 by removing version-checking of typedefs in CTS

### LINKED ISSUE #2326: CTS checking types against a hard-coded version number?
Currently the CTS type-checking tests are looking at a hard-coded type version number of `1`.  As best I can tell one of the recent merges updated the types to version `2`, which is resulting in pretty much every type now failing validation through the CTS.

(See assertion 7 and 8 in `TestSupportedTypeDef`.)

I'm not sure which one needs the corrective action:

- should the CTS skip checking a version number at all?
- should the CTS type checking test cases be updated to match the version number of the open types for the release?
- should the open metadata types not have their version increased unless they have actually been patched / created / somehow modified in the new release (ie. should unmodified types retain their original version number)?
- some other change / fix?

cmgrote:


cmgrote:
Ok, let me revert the changes locally -- it could be that I was testing against some interim state of the merged PRs that caused the original problem, and perhaps it is now fixed anyway.

mandy-chessell:
One of the bugs I fixed yesterday was in the clone method of the typeDefPatch and the updateTime and updatedBy fields were not being copied - so it is possible that you are a victim of that problem.

cmgrote:
Whatever the problem was, indeed it seems to be working now without this PR -- closing this and the related issue.

Issue #2348, opened 2020-01-03
IGC connector not loading - user/env error - improve audit msgs


planetf1:
Using the odpi-egeria-vdc helm chart with the following override
ie
```
helm install vdc odpi-egeria-vdc -f ~/etc/cloud_igc.yaml
```
where the yaml contains: (hostname/ip obfuscated)
```
ibmigc:
  enabled: true
  internal:
    enabled: false
  external:
    enabled: true
    ip: "111.222.333.444"
    hostname: "igc.openmetadata.mydomain.com"
    ports:
      https: "9446"
      broker: "59092"
```
results in all the pods etc starting up. However the IGC proxy fails to initialize correctly with:
```
Fri Jan 03 14:31:48 GMT 2020 ibmigc Exception OMRS-AUDIT-0035 The connector to the local repositorys event mapper failed with a {0} exception and the following error message: {1}
```

Two issues here
 a) The audit message is incorrectly formatted - preventing customer PD
 b) Why the actual failure

The same error was reported  on slack at https://odpi.slack.com/archives/CBPRDQPRC/p1578028854030200 -- this was running locally, not using containers. I tried the helm chart to get my own environment working for debugging and comparison.

cmgrote:
a) It looks like all of the messages in `OMRSAuditCode` use numerically-indexed formatters `{0}`, `{1}`, etc -- whereas our other classes all seem to just use `{}` with the order implied by the order the arguments are passed to the formatting method (?)

b) This is likely due to either not being able to resolve the address between the host running Egeria and the IGC hostname, or that the IGC host's Kafka bus believes it is a different hostname than the one it was initially called on.

I'll try to do some further testing related to (b) on my side and see if I can add more meaningful audit log messages on the IGC event mapper side...

planetf1:
Enabled debug (should not be needed)

Errors include:
```
2020-01-03 14:50:41.503 ERROR 1 --- [nio-8080-exec-1] o.o.o.a.r.ConnectorConfigurationFactory  : Bad connectorProviderClassName: org.odpi.egeria.connectors.ibm.igc.repositoryconnector.IGCOMRSRepositoryConnectorProvider
2020-01-03 14:50:41.869 ERROR 1 --- [nio-8080-exec-5] o.o.o.a.r.ConnectorConfigurationFactory  : Bad connectorProviderClassName: org.odpi.egeria.connectors.ibm.igc.eventmapper.IGCOMRSRepositoryEventMapperProvider
```

[...]

PR #2352, opened 2020-01-03
Corrects connector usage within Helm charts

cmgrote:
- uses Artifactory API to download latest available pre-built artifacts (snapshots or releases)
- corrects (again?) the configuration calls used to initialize the connectors
- updates the connector jar filenames based on new connector assembly mechanisms


cmgrote:
Should address the issue identified in #2348, but does not resolve the lack of audit logging that makes identifying this issue problematic.

PR #2353, opened 2020-01-03
Avoids swallowing exceptions for classes provided at runtime

### LINKED ISSUE #2348: IGC connector not loading - user/env error - improve audit msgs

cmgrote:
Resolves #2348 issue related to no audit logged message in cases of ClassNotFoundException, ie. when class details are provided at runtime (eg. external connectors).

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/connector-configuration-factory/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/ConnectorConfigurationFactory.java
+++ b/open-metadata-implementation/adapters/open-connectors/connector-configuration-factory/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/ConnectorConfigurationFactory.java
@@ -503,11 +503,16 @@ public class ConnectorConfigurationFactory
      * @param url  location of the repository proxy
      * @param configurationProperties name value pairs for the connection
      * @return Connection object
+     * @throws ClassNotFoundException when the provided class cannot be found
+     * @throws InstantiationException when the provided class cannot be instantiated
+     * @throws IllegalAccessException when there is insufficient access to instantiate the provided class
      */
     public Connection  getRepositoryProxyConnection(String              serverName,
                                                     String              connectorProviderClassName,
                                                     String              url,
-                                                    Map<String, Object> configurationProperties)
+                                                    Map<String, Object> configurationProperties) throws ClassNotFoundException,
+                                                                                                        InstantiationException,
+                                                                                                        IllegalAccessException
     {
         final String endpointGUID             = UUID.randomUUID().toString();
         final String connectionGUID           = UUID.randomUUID().toString();
@@ -536,7 +541,7 @@ public class ConnectorConfigurationFactory
         connection.setDisplayName(connectionName);
         connection.setDescription(connectionDescription);
         connection.setEndpoint(endpoint);
-        connection.setConnectorType(getConnectorType(connectorProviderClassName));
+        connection.setConnectorType(getDynamicConnectorType(connectorProviderClassName));
         connection.setConfigurationProperties(configurationProperties);
 
         return connection;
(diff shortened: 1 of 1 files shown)


cmgrote:
With this PR, now the following behavior is present:

- If trying to setup an event mapper connection via the configuration API using a class that is not available to the OMAG Server Platform, the user will immediately receive a response body along these lines (although confusingly an HTTP return code of `200`?):

```
{
    "relatedHTTPCode": 500,
    "exceptionClassName": "org.odpi.openmetadata.adminservices.ffdc.exception.OMAGConfigurationErrorException",
    "exceptionErrorMessage": "OMAG-ADMIN-500-001 Method setLocalRepositoryEventMapper for OMAG server myserver returned an unexpected exception of java.lang.ClassNotFoundException with message org.odpi.egeria.connectors.ibm.igc.eventmapper.IGCOMRSRepositoryEventMapperProvider1",
    "exceptionSystemAction": "The system is unable to configure the OMAG server.",
    "exceptionUserAction": "This is likely to be either an operational or logic error. Look for other errors.  Validate the request.  If you are stuck, raise an issue."
}
```

- if trying to setup a repository connector using a class that is not available to the OMAG Server Platform, on the `/instance` call the following audit log entries will be output:

```
Fri Jan 03 17:30:23 GMT 2020 myserver Information OMRS-AUDIT-0001 The Open Metadata Repository Services (OMRS) is initializing
Fri Jan 03 17:30:23 GMT 2020 myserver Information OMRS-AUDIT-0003 The local repository is initializing with metadata collection named myserver with an id of 0218dee3-b0a6-441f-920b-29a8ce558b66
Fri Jan 03 17:30:23 GMT 2020 myserver Information OMRS-AUDIT-0029 The local repository outbound event manager is initializing
Fri Jan 03 17:30:23 GMT 2020 myserver Exception OMRS-AUDIT-0034 The connector to the local repository failed with a org.odpi.openmetadata.frameworks.connectors.ffdc.ConnectionCheckedException exception and the following error message: OCF-CONNECTION-400-005 Unknown Connector Provider class org.odpi.egeria.connectors.ibm.igc.repositoryconnector.IGCOMRSRepositoryConnectorProvider1 passed in connection <Unknown>
Platform shutdown requested. Temporary deactivating any auto-started servers
```

though the `/instance` call itself actually confusingly returns a success `200` code, it shows the following response body:

```
{
    "class": "SuccessMessageResponse",
    "relatedHTTPCode": 500,
    "exceptionClassName": "org.odpi.openmetadata.adminservices.ffdc.exception.OMAGConfigurationErrorException",
    "exceptionErrorMessage": "OMAG-ADMIN-500-001 Method activateWithSuppliedConfig for OMAG server myserver returned an unexpected exception of org.odpi.openmetadata.repositoryservices.ffdc.exception.OMRSConfigErrorException with message OMRS-CONNECTOR-400-004 The connection null passed to the EnterpriseOMRSRepositoryConnector is invalid.",
    "exceptionSystemAction": "The system is unable to configure the OMAG server.",
    "exceptionUserAction": "This is likely to be either an operational or logic error. Look for other errors.  Validate the request.  If you are stuck, raise an issue."
}
```

planetf1:
I'd like to include this in 1.3 if ok? It would help in configuring connectors.

mandy-chessell:
This is a really nice change - howevber, I need to comment on the 200 response codes.
In the REST protocol it is only possible to return a response object if the status code is 200.  If we returned, say 500 then it would not be possible to return the error message.

Using 200 is not breaking the protocol because it does not mean "success" as many people think - it mean there is further information in the response object.

Issue #1828, opened 2019-10-23
Update httpclient > 4.5.2


planetf1:
Currently the janus connector is using org.apache.httpcomponents:httpclient:4.5.2 which is subject to a directory traversal security exposure documented as SONATYPE-2017-0359 (7)

Current versions appear to have addressed this flaw

grahamwallis:
Which connector is this referring to?

planetf1:
```
[INFO] org.odpi.egeria:janus-connector:jar:1.2-SNAPSHOT
[INFO] +- org.janusgraph:janusgraph-es:jar:0.3.1:compile
[INFO] |  +- org.elasticsearch.client:elasticsearch-rest-client:jar:6.0.1:compile
[INFO] |  |  +- org.apache.httpcomponents:httpclient:jar:4.5.2:compile
[INFO] |  |  +- org.apache.httpcomponents:httpcore:jar:4.4.5:compile
[INFO] |  |  +- org.apache.httpcomponents:httpasyncclient:jar:4.1.2:compile
[INFO] |  |  \- org.apache.httpcomponents:httpcore-nio:jar:4.4.5:compile
```

grahamwallis:
OK - thanks - it isn't obvious from the name that this is part of the OLS.

grahamwallis:
This issue should really be assigned to Guy or someone working on the OLS. 
I cannot change the assignment.

planetf1:
@GuyAckermans can you take a look (cc: @grahamwallis )

GuyAckermans:
Will do

planetf1:
This has been fixed everywhere we can control it -- we do still pull in 4.5.2 via hadoop - but this will be addressed by moving hadoop components to a new repo

PR #2386, opened 2020-01-09
Elastic search dependency internal library version update!

DimitriosMaimaris:
Change for elastic rest client version based on a security finding. Janus elastic search dependency is using 6.6.0 version of elastic-rest-client which is vulnerable according to 

https://nvd.nist.gov/view/vuln/search-results?adv_search=true&cves=on&cpe_version=cpe%3A%2Fa%3Aelastic%3Aelasticsearch%3A0.4.0)

Explicit update of the rest-client to version 6.8.6.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,10 +126,10 @@
         <!-- Versions of dependent libraries -->
         <junit.platform.version>1.5.2</junit.platform.version>
         <junit.jupiter.version>5.5.2</junit.jupiter.version>
-        <junit.version>4.12</junit.version>
+        <junit.version>4.13</junit.version>
         <surefire.plugin.version>3.0.0-M4</surefire.plugin.version>
         <mockito.version>3.2.4</mockito.version>
-        <testng.version>7.0.0</testng.version>
+        <testng.version>7.1.0</testng.version>
         <log4j.version>1.2.17</log4j.version>
         <slf4j.version>1.7.30</slf4j.version>
         <logback.version>1.2.3</logback.version>
@@ -161,7 +161,7 @@
         <quartz.version>2.3.2</quartz.version>
         <lucene-analyzers-common.version>7.0.0</lucene-analyzers-common.version>
         <guava.version>23.0</guava.version>
-        <tink.version>1.3.0-rc2</tink.version>
+        <tink.version>1.3.0-rc3</tink.version>
         <jsonld.version>0.13.0</jsonld.version>
         <commons-logging.version>1.2</commons-logging.version>
         <commons-codec.version>1.13</commons-codec.version>
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/types/TestConsistentTypeDef.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/types/TestConsistentTypeDef.java
@@ -72,8 +72,11 @@ public class TestConsistentTypeDef extends RepositoryConformanceTestCase
                 TypeDef restAPITypeDef = metadataCollection.getTypeDefByGUID(workPad.getLocalServerUserId(),
                                                                              typeDef.getGUID());
 
-
-                assertCondition((typeDef.equals(restAPITypeDef)),
+                /*
+                 * A full equality test is too strong because the type may be at a different version, in which case numeric version field will differ.
+                 * Test full equality only if the version is identical.
+                 */
+                assertCondition(( (typeDef.getVersion() != restAPITypeDef.getVersion()) || typeDef.equals(restAPITypeDef) ),
                                 getAssertionName(assertion1),
                                 typeDef.getName() + assertionMsg1,
                                 super.defaultProfileId,
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/types/TestSupportedTypeDef.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/types/TestSupportedTypeDef.java
@@ -233,7 +233,7 @@ public class TestSupportedTypeDef extends RepositoryConformanceTestCase
                         super.defaultProfileId,
                         super.defaultRequirementId);
 
-        verifyCondition(typeDef.equals(resultObject),
+        verifyCondition(( (typeDef.getVersion() != resultObject.getVersion()) || typeDef.equals(resultObject) ),
                         assertion15,
                         testTypeName + assertionMsg15,
                         super.defaultProfileId,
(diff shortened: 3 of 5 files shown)


pom.xml
```diff
@@ -1964,12 +1976,6 @@
                 <scope>compile</scope>
             </dependency>
 
-            <dependency>
```
planetf1: I notice you have removed this dependency again -- it was the reason for the last PR build failure and this one -- we need it. It defines a transient dependency

pom.xml
```diff
@@ -475,8 +474,21 @@
                 <artifactId>janusgraph-es</artifactId>
                 <scope>compile</scope>
                 <version>${janus.version}</version>
+                <exclusions>
```
planetf1: is this because this piece is not needed?

I see you have an entry in dependencyManagement for the rest client. If set that should ensure we use the correct version transitively too -- so you should not need the exclude here (adding excludes gets harder to manage ... though obviously they have their uses!)_

Can you check with 'mvn dependency:tree'. Is there nay difference if you remove the exclude? What is it?
Also are you getting the right versions.

I usually run this on master, and my PR, and then use a side by side compare (I tend to use atom) to verify. This is needed - I can run if required, but suggest not merging the PR until verified

We just need to be careful with dependencies as it's so easy for them

DimitriosMaimaris: OWSAP Scan declared that janusgraph-es is HIGH vulnerability. I opened the link I have attached and after I did the dependency tree to see , you can check it has elastic-search-client 6.6.0 that is why I excluded it like that. After the exclusion it does not appear in the tree anymore. Yes lets verify first and then merged it that is why I put a lot of people in the reviewers list. Some use elastic so may have to test it as well.

pom.xml
cong78: Hi Nigel, from the build failure it shows the dependency is missing from artifactory of odpi.jfrog as we can build it locally. Would that make sense for you to have a look on it? Thanks 
`org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project graph-repository-connector: Could not resolve dependencies for project org.odpi.egeria:graph-repository-connector:jar:1.4-SNAPSHOT: Could not find artifact com.sleepycat:je:jar:7.5.11 in central (https://odpi.jfrog.io/odpi/egeria-release)`

DimitriosMaimaris: If we have the dependency as it was we can open the graph in gremlin console it needs an older version of this dependency. Since it has it embedded in some library there is no need of having it explicitly. Locally is building with no issues.

pom.xml
DimitriosMaimaris: We are going also to check the OWASP again to see if it is still an issue or not before merging.

pom.xml
DimitriosMaimaris: @planetf1  Berkeley is using this one:
    - com.sleepycat:je:jar:7.5.11:compile

If we can find a version that is compatible with this then we can added as a separate dependency with no issue.

planetf1: > Hi Nigel, from the build failure it shows the dependency is missing from artifactory of odpi.jfrog as we can build it locally. Would that make sense for you to have a look on it? Thanks
> `org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal on project graph-repository-connector: Could not resolve dependencies for project org.odpi.egeria:graph-repository-connector:jar:1.4-SNAPSHOT: Could not find artifact com.sleepycat:je:jar:7.5.11 in central (https://odpi.jfrog.io/odpi/egeria-release)`

It's missing in maven central too - which would cause an issue for anyone else using berkeley. We need to use a version oracle has pushed to the repos. See https://github.com/odpi/egeria/issues/1116

Adding to our artifactory at odpi isn't sufficient

planetf1: > If we have the dependency as it was we can not open the graph in gremlin console it needs an older version of this dependency. Since it has it embedded in some library there is no need of having it explicitly. Locally is building with no issues.
 
Are you saying the newer berkely is causing an issue with gremlin? I know @grahamwallis uses gremlin, and that dependency has been in place for quite a while

Can you elaborate on this?
Do you have any instructions on how to repro? I'm happy to offer an additional opinion, but don't know how to setup the environment (simply)? Or can you share a transcript of what you see when it 'fails'? And how does your local build's dependency tree differ to master (or this proposal)?

pom.xml
planetf1: The first section seems to exclude elasticsearch-rest-client by default when janusgraph-es is used as a dependency (though it could be added explicitly). The latter says that if elasticsearch-rest-client is used, it must be at 6.8.6 -- is this what you are trying to achieve?

Also which CVE/issue are you addressing? I notice that our CLM scans aren't showing an issue with elastic-search-client, other than it's old. 

I'll try running the maven build with 'mvn clean install -DCVE' which should generate the CVE report to double check.

pom.xml
DimitriosMaimaris: To reproduce if you want you can keep the sleepycat dependency in with the latest version as it was and start the open lineage services. I can send you the post call if you like. Then if you try to open the graph from the gremlin console you will see this.

Caused by: com.sleepycat.je.VersionMismatchException: (JE 7.5.11) /Users/wd38tc/Workspace/egeria/egeria-lineage-repositories/bufferGraph/berkeley Expected log version 15 or earlier but found 17 VERSION_MISMATCH: The existing log was written with a version of JE that is later than the running version of JE, the log cannot be read. Environment is invalid and must be closed. Couldn't open file /Users/Workspace/egeria/egeria-lineage-repositories/bufferGraph/berkeley/00000000.jdb

pom.xml
cong78: https://nvd.nist.gov/view/vuln/search-results?adv_search=true&cves=on&cpe_version=cpe%3A%2Fa%3Aelastic%3Aelasticsearch%3A0.4.0
Hi Nigel, this is the CVE/issue that we are addressing.

pom.xml
planetf1: Thanks. Yes REST would be good.

Are you starting from a clean environment? ie if you have no lineage graph at all - completely clean config of egeria. It looks as if there is an incompatibility between files (the tx log) written with one version, and read by another?

Another possibility is that you have inadvertantly included 2 different versions of berkeley - but you would see that in the mvn dependency:tree plus our build validation *should* prevent it.

Or are you running both old and new code?

cong78:
I will close this PR to figure out whether the issue is valid.

planetf1:
Additional notes for the benefit of anyone reviewing at a later date...

it needs a little more evaluation. It’s interesting that our Nexus scanner didn’t note it, but OWASP (which is well respected) does. It looks as if this old vulnarability is in the /implementation/ of the elastic search rest api, yet janusgraph-es only contains client code? In which case it wouldn’t be vulnerable… but need to analyse in more depth.

Additionally for handling an issue like this, dependencyManagement may be sufficient (without exclusion, just a version requirement), especially as janusgraph-es happily states support  https://mvnrepository.com/artifact/org.janusgraph/janusgraph-es/0.4.0 for 7.5.1

It’s also related to this -> https://github.com/odpi/egeria/issues/1828 which I opened against OLS in October

PR #2385, opened 2020-01-09
Ensures underlying connector issues are available in audit log

cmgrote:
Audit logs connector issues as exception to ensure underlying causes are available, resolves #2384

PR #2387, opened 2020-01-09
Fixes LocalConnector validations and considers type version in CTS type compares

### LINKED ISSUE #2356: CTS failures after typedef version changes & fixes
In Release 1.2 I saw
 - 4 CTS failures for the graph repo (all related to the way deletes were handled - subsequently resolved in master)
 - 0 failures for the in-mem repo

However the current master is showing more failures. For example inmem currently shows:
```
Date of conformance test:          2020-01-03T18:28:48.861+0000
Number of tests that were run:     4495
Number of tests that failed:       445
Number of tests that were skipped: 0
```

cc: @grahamwallis @cmgrote @mandy-chessell 

We likely need to resolve this before cutting the 1.3 release.

grahamwallis:
Fixes #2356

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -129,7 +129,7 @@
         <junit.version>4.13</junit.version>
         <surefire.plugin.version>3.0.0-M4</surefire.plugin.version>
         <mockito.version>3.2.4</mockito.version>
-        <testng.version>7.0.0</testng.version>
+        <testng.version>7.1.0</testng.version>
         <log4j.version>1.2.17</log4j.version>
         <slf4j.version>1.7.30</slf4j.version>
         <logback.version>1.2.3</logback.version>


planetf1:
See comment above, but propose we merge this now in any case.

planetf1:
I will do the 1.3 backport.

cmgrote:
Looks good against IGC connector:

```
=======================================
 Open Metadata Conformance Test Report
=======================================
 ... contacting conformance suite server: cts (http://localhost:8080)

Conformance report from server:  cts

Number of tests: 582
Number of tests passed: 582
Number of tests failed: 0
Number of tests skipped: 0

Congratulations, technology under test is conformant
```

grahamwallis:
With regard to the question "when the typedef versions differ, what checks should be done. In theory we would expect compatible changes in a later version (ie additional attributes etc)? Does that matter?" I did start to implement a 'compatibility check' method but quickly realised that it would be a much bigger piece of work - and really needs to be checked when the type update is first made. For now I have implemented a check that says "if version is the same, the type must match exactly, otherwise it is not checked".

planetf1:
See https://github.com/odpi/egeria/pull/2385/commits for 1.3

planetf1:
And yes, when I thought about it - agreed, it's non trivial

Issue #2388, opened 2020-01-09
Investigate why Polymer build needs 3GB to run (node, windows, workaround in place)


planetf1:
Building on windows 10 (very current) with adoptopenjdk 11 openj9 the polymer build fails with:
```
[INFO]
[INFO] --- exec-maven-plugin:1.6.0:exec (Polymer build) @ ui-chassis-spring ---
info:   Clearing build\ directory...
info:   (dev) Building...
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 00007FF63BB3ACB5
 2: 00007FF63BB144E6
 3: 00007FF63BB14EF0
 4: 00007FF63BDA47FE
 5: 00007FF63BDA472F
 6: 00007FF63C2D8964
 7: 00007FF63C2CF0F2
 8: 00007FF63C2CD68C
 9: 00007FF63C2D65B7
10: 00007FF63C2D6636
11: 00007FF63BE81677
12: 00007FF63BF189D2
13: 000002DDCBFDC5C1
[ERROR] Command execution failed.
org.apache.commons.exec.ExecuteException: Process exited with an error: 134 (Exit value: 134)
    at org.apache.commons.exec.DefaultExecutor.executeInternal (DefaultExecutor.java:404)
    at org.apache.commons.exec.DefaultExecutor.execute (DefaultExecutor.java:166)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:804)
    at org.codehaus.mojo.exec.ExecMojo.executeCommandLine (ExecMojo.java:751)
```

A google search suggests this isn't uncommon... and may need changes to the build scripts.

As an aside: Seeing a few platform specific issues - may be worth extending our automated builds to include windows. OR to explicitly state we are not supporting/working with. Or somewhat inbetween, only support WSL2 See #1307

planetf1:
Raised whilst debugging #2350

planetf1:
Some tips at:
 - https://github.com/angular/angular-cli/issues/13734

planetf1:
When the 'Polymer build' execution occurs I see 'node' consuming lots of memory before crashing.
As per the linked articles it's possible to increase the memory for node by a command line parm, or setting NODE_OPTIONS

So for example one can include this in the maven exec plugin definition for Polymer build (within the configuration tag)
```
                                    <environmentVariables>
[...]

PR #2390, opened 2020-01-09
#2388 increase node heap to fix Windows UI build failure OOM

planetf1:
Fixes broken Windows build caused by Polymer build Out of memory


planetf1:
See #2388

Issue #2360, opened 2020-01-06
REST API response status codes?


cmgrote:
Per discussion on #2353:

> ... I need to comment on the 200 response codes. In the REST protocol it is only possible to return a response object if the status code is 200. If we returned, say 500 then it would not be possible to return the error message.

Is this a limitation on our side, or a difference in a standard that defines how REST API status codes differ from the HTTP protocol itself?  I can't see anything in the HTTP standard restricting message bodies (response objects?) to status code 200.  The overview of status codes (https://tools.ietf.org/html/rfc7231#section-6) explains:

> "HTTP clients are not required to understand the meaning of all registered status codes, though such understanding is obviously desirable.  However, a client MUST understand the class of any status code, as indicated by the first digit...  For example, if an unrecognized status code of 471 is received by a client, the client can assume that there was something wrong with its request and treat the response as if it had received a 400 (Bad Request) status code.  The response message will usually contain a representation that explains the status."

(Per the last sentence, a non-200 code of 471 has a response message that provides further explanation.)  Regarding a 500 response, specitically (https://tools.ietf.org/html/rfc7231#section-6.6):

> "Except when responding to a HEAD request, the server SHOULD send a representation containing an explanation of the error situation, and whether it is a temporary or permanent condition.  A user agent SHOULD display any included representation to the user."

> Using 200 is not breaking the protocol because it does not mean "success" as many people think - it mean there is further information in the response object.

Again, is this something specific to REST APIs vs the HTTP protocol itself?  From the standard (https://tools.ietf.org/html/rfc7231#section-6.3), it feels pretty strongly worded to me that 200 _is_ defined as success:

> "Successful 2xx: The 2xx (Successful) class of status code indicates that the client's request was successfully received, understood, and accepted.  The 200 (OK) status code indicates that the request has succeeded."

It's confusing to me that we would specify different HTTP status codes (like 500, 404, etc) on the definition of errors, but not actually use them as the response status code -- which per the standard references above seems like it should be possible? -- and I suspect will be confusing to users as well (who upon seeing a 200 response status may well simply ignore any body of the response itself)?

Suggestion:

1. if we can, we change our approach to make use of the HTTP status codes and record a message body in the response
2. if not, that we clearly document somewhere quite clearly that the 200 codes we use in responses do not necessarily mean success.

cmgrote:
[...]

PR #2393, opened 2020-01-10
HTTP status codes (for Spring)

cmgrote:
My understanding is that the way the status codes are handled is going to depend on the REST framework we're actually deploying -- this adds handling for Spring, and keeps it coupled only to where Spring is already a dependency (ie. the REST controller definitions).

Targeting issue #2360

Leaving the PR as draft for discussion and review purposes while I iterate through updating the various pre-existing REST resources with the same changes.

As described on the issue, the intentions of this change are to:

- return HTTP status codes other than 200, ie. for responses indicating some error
- retain the existing response bodies that are already in use for such errors (ie. only change the response status code, not the actual content of the response)
- tie the Spring-specific implementation of these response codes only to where Spring is already a dependency (ie. not to change any existing exception or error-handling, serialization, classes, etc)


mandy-chessell:
@cmgrote we should not be attempting a code change until we have agreed on a way forward.

cmgrote:
The intention was to use this as a means of discussing a potential way forward (with the expectation it would help to see an example) -- hence the draft status...

PR #2414, opened 2020-01-13
Implements updated search tests for CTS

### LINKED ISSUE #2413: CTS search test refactoring
In 1.4, the CTS search tests should be improved to:
a) improve performance - speeding up overall run time and/or enabling larger result sizes to be tested
b) increase similarity of approach for read-only repositories vs read-write repositories
c) simplify and reduce the search test cases to simplify maintenance and future enhancement

This change consolidates the previously 3 entity search test cases into one. Similarly, the 3 relationship search test cases are consolidated into one,

To cater for read-only/read-write repositories is to initially perform a 'wild' search for the type being tested. If that search returns instances they are used as the set of known instances upon which subsequent search tests are based. If the wild search does not return a known instance set, the test case attempts to create a set of known instances. If the repository supports read-write (formally, the metadata_maintenance profile) the created set is used. If the repository does not support creation of instances (and as has already been established) has no existing instances of that type, then the result for that type will be set to 'unknown_status' (as it is not possible to pursue the test for the requested type).

Provided search tests are possible, the test case will:

- Test that every primitive attribute can be used for property search, using one of the known values from the known instances. This is repeated for MatchCriteria.ALL and MatchCriteria.NONE to ensure that positive and negative filtering are both possible.

- Test that a pair of primitive attributes can be used in property search, with  MatchCriteria.ALL, MatchCriteria.ANY and MatchCriteria.NONE to ensure that logical combinations operate correctly.

- Test that for one string type attribute, property value search works correctly, using regular expressions from the RepositoryHelper regex methods; for Exact, Prefix, Suffix and Contains matches.

- For every primitive attribute, using one known value for each, type filtering (to the type under test) can be used to narrow the search compared to searching with no type filter.

- Test that for one string type attribute, property value search works correctly, using general regular expressions (i.e. testing the advanced_search profile); for Exact, Prefix, Suffix and Contains matches.

The above search tests for type Asset (for example) run in approximately 2 seconds.

grahamwallis:
Fixes #2413

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -124,8 +124,8 @@
         <classmate.version>1.5.1</classmate.version>
 
         <!-- Versions of dependent libraries -->
-        <junit.platform.version>1.5.2</junit.platform.version>
-        <junit.jupiter.version>5.5.2</junit.jupiter.version>
+        <junit.platform.version>1.6.0</junit.platform.version>
+        <junit.jupiter.version>5.6.0</junit.jupiter.version>
         <junit.version>4.13</junit.version>
         <surefire.plugin.version>3.0.0-M4</surefire.plugin.version>
         <mockito.version>3.2.4</mockito.version>


cmgrote:
In other places in the results, I can see that there are unexpected exceptions that probably are true issues on my side -- one or two `NullPointerException`s, and some `RepositoryErrorException`s.  However, the CTS seems to be swallowing the exceptions entirely -- even with INFO-level logging on there is nothing in the console or dev log.  This makes troubleshooting the root cause of such issues basically impossible (without going in and modifying the CTS itself myself), which for me is a non-workable situation for the CTS.

I think we should at least be `logException`ing such exceptions via the audit log, if we're not using a more dev-centric logger for the CTS?

grahamwallis:
Hi @cmgrote - there are tests that will pass general regexes - they are as you suggest to test for support of the advanced profile. I hadn't realised you were spotting them and throwing a FunctionNotSupportedException. 
Could you please send me your test result file so I can make sure I fix any/all offending tests?

cmgrote:
> there are tests that will pass general regexes - they are as you suggest to test for support of the advanced profile. I hadn't realised you were spotting them and throwing a FunctionNotSupportedException.

Just tried to dig through old issues to jog my memory on whether we'd formally agreed on this behaviour, but can't find anything that says one way or another...  Do we feel this is a valid approach for a repository saying it cannot support a regular expression?  (Guess we should at minimum update the JavaDocs if so?)

grahamwallis:
With regard to your question:

> Just tried to dig through old issues to jog my memory on whether we'd formally agreed on this behaviour, but can't find anything that says one way or another... Do we feel this is a valid approach for a repository saying it cannot support a regular expression? (Guess we should at minimum update the JavaDocs if so?)

My recollection is that the repository should "do it's best" and if that is to return nothing then that was what was wanted. It may be nicer to throw a FunctionNotSupportedException if the repository/connector knows immediately that it cannot parse/support the search attempt. We should discuss with Mandy and record the decision.

cmgrote:
The exception logging (in the areas it was added) is now great -- very helpful!

However, I can see in other areas that tests are failing (eg. too many results returned, or results that don't match are returned), but because no exception was thrown there isn't any detail on the specific parameters that were used for the test.  It would be very helpful for these non-exception failure scenarios to still capture the same level of detail: what method was run, with what parameters, what was expected (eg. how many results) and what was actually found.

grahamwallis:
Will raise additional issues to roll out unexpected exception logging to the other (non-search) testcases, and to add more detail to the assertion messages.

PR #2415, opened 2020-01-13
Fix User Interface build 

planetf1:
This fix ensures the UI builds properly, even when run for the first time.
Previously the 'official' build artifacts for the UI were bad, as they did not include the public resources - so no web page

The reason for this was that we were:
 * In the 'prepare-resources' phase, copying src/main/resources into target/classes (maven default behaviour - as this is intended for static resources)
* Creating our polymer UI, with static resources left in src/main/static
* copying src/main/static/build/dev (the built UI static files) resources into src/main/resources/public
* Building the jar following default behaviour
* spring-ifying the jar

The problem was the ordering -- at the time resources are copied into where we package from they hadn't been built! This is why, combined with the failure to cleanup after each build, it worked second time around....

This PR :
 * Ensures the built UI is deleted as part of the 'clean' target -- though for now we leave the downloaded npm/node tools intact, since the setup can take a while & require network.
 * Removes the extensive <profiles> section for skipUI from the ui-chassis-spring build & moves it up to a simple optional module inclusion in user-interface. The effect is the same, but it avoids complexity in the lower level module, and is more aligned to how we manage skipFVT
* Copies the built polymer UI directly into target/classes/public so that it is visible to the maven-jar-plugin 

Tested .. the jar is now complete each time and works fine.

### DIFF
--- a/open-metadata-resources/open-metadata-tutorials/omag-server-tutorial/task-starting-the-omag-server-platform.md
+++ b/open-metadata-resources/open-metadata-tutorials/omag-server-tutorial/task-starting-the-omag-server-platform.md
@@ -21,11 +21,11 @@ Check the version of Java you have with the command **java -version**
 and the **HotSpot** version for your operating system.
 You only need the JRE but select the JDK if you expect to also write some Java code.)
 
-Start the OMAG server platform as follows - the `-Dserver.port` parameter is optional, but is needed if you want multiple instances running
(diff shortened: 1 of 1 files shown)


planetf1:
As this is effectively a build break I would like to merge today, ideally this morning
Any comments/concerns appreciated

planetf1:
Merging after 2 approvals - thanks

PR #2441, opened 2020-01-17
BUMP springboot - resolves build fail in #2440

planetf1:

This replaces PR #2440 by also updating tomcat - resolving the build failure caused by conflicting dependent versions of tomcat


planetf1:
tested against notebooks/UI etc

Issue #2442, opened 2020-01-17
BLOCKING RELEASE - Release & merge pipelines failing 


planetf1:
Currently the Merge build is failing - see https://dev.azure.com/ODPi/Egeria/_build?definitionId=12&_a=summary

This is failing in the ArtifactoryMaven task with the following error:
```
[Info] Downloading bintray/jcenter/org/jfrog/buildinfo/build-info-extractor-maven3/2.13.13/build-info-extractor-maven3-2.13.13-uber.jar
[Info] Downloaded 1 artifact.
[Error] couldn't find plexus-classworlds-x.x.x.jar in Maven installation path, please check M2_HOME environment variable
##[error]Error: Command failed: /opt/hostedtoolcache/jfrog/1.31.2/x64/jfrog rt mvn "clean install -f /home/vsts/work/1/s/pom.xml --batch-mode -e -s .m2/settings.xml -Dmerge -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -P artifactory" /home/vsts/work/1/s/config --build-name="Egeria" --build-number="20200117.1"
Collecting environment variables...
```

This task is provided by the ArtifactoryMaven plugin from the marketplace, which is up to date.

There appear to be other reports of the failure including
https://github.com/actions/virtual-environments/issues/279

An attempt to release Egeria 1.3 also failed (see https://dev.azure.com/ODPi/Egeria/_releaseProgress?_a=release-pipeline-progress&releaseId=78 ) with:
```
2020-01-16T20:43:13.4730326Z [Error] couldn't find plexus-classworlds-x.x.x.jar in Maven installation path, please check M2_HOME environment variable
2020-01-16T20:43:13.4812777Z ##[error]Error: Command failed: /opt/hostedtoolcache/jfrog/1.31.2/x64/jfrog rt mvn "install -f /home/vsts/work/r1/a/_ODPI_Egeria_Commit/pom.xml --batch-mode -e -s /home/vsts/agents/2.164.3/.m2/settings.xml -P artifactory -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn" /home/vsts/work/r1/a/config --build-name="Egeria-Stage" --build-number="5e7b152de"
```

This fundamentally looks like an update is needed to the jfrog task definition (in their plugin).. or alternatively needs recoding 'manually'. 

[...]

PR #2443, opened 2020-01-17
URGENT - Workaround JFrog build failures

planetf1:

Attempts to workaround jfrog issue breaking build.
See #2442 for info


planetf1:
Note: @mandy-chessell I will merge ASAP. No guarantee this will address the problem, but worth a try & is currently blocking further activity. If it works I'll apply a similar fix to the release process.

PR #2445, opened 2020-01-17
Open Lineage Services: Shutdown Server Endpoint

GuyAckermans:


GuyAckermans:
The logToAudit method is strictly intended for audit logging where no further parameters are required. The addition of this method greatly reduces the number of lines of code in the class. I was not aware of the logException method, will update the code

GuyAckermans:
Updated

GuyAckermans:
Both loggers are being used, audit.logException() is called in the OCFCheckedExceptionToOMAGConfigurationError method

mandy-chessell:
> The logToAudit method is strictly intended for audit logging where no further parameters are required. The addition of this method greatly reduces the number of lines of code in the class. I was not aware of the logException method, will update the code

Having a standard method is fine.  However you will probably need a variety - when errors occur, the message needs enough information for the person who is debugging the problem to know what the incorrect values were.  The rest of the message describes how to fix the problem.

logException allows the exception to be passed to the auditLog.  It is for error cases where things like the stack trace are needed to debug.  The auditLog will add details of the exception to the audit log record that is passed to each of the audit log destinations.

GuyAckermans:
All exceptions are now written to the auditlog and when an error occurs with the GraphDB, the entire OLS config is written to the audit log as well. I will finetune the latter in a future PR with more precise details on which config values are invalid

PR #2447, opened 2020-01-17
Removes debug-logging that could result in NPEs during exception throwing

### LINKED ISSUE #2446: Potential for NPEs during exception construction
I found a few spots where our own Exceptions were debug-logging, and explicitly calling methods like `.toString()` on objects that could be null (and would thus result in an NPE as part of throwing such an exception).

I don't believe there's any reason for such debug-logging to remain in these exceptions, so will remove it; alternatively, we should do null-handling before attempting to call such methods.

cmgrote:
Resolves #2446


cmgrote:
Do we need the debug logging?  No other exception classes have this logging...

I'm also not sure I agree: looking at Java's own implementation of exceptions (tracing the various `super(...)` calls all the way up to `Throwable`) it does its own null-handling of `cause`s rather than relying on the caller to correctly determine whether they have a null cause or not when throwing the exception.  (In fact, it appears to do its own null handling of _all_ of its potential parameters.)

To me this makes more sense to have the exception handle its own potential nulls.  To me it's very dangerous to have the error-handling mechanism itself dependent on people calling it "properly" -- if it fails because it hasn't been called correctly this can be very confusing to track down.  (And there is nothing at the Java level currently that will check that it is called "properly" at compile-time, only possible at runtime without adding annotations.)

cmgrote:
Fundamentally I was trying to kill 2 birds with 1 stone -- as I thought we were trying to get rid of unnecessary logging (which this appeared to be)...  However if we still need it there's probably an easier fix: not to send concatenated strings to the logger, but to use the logger's own formatting operations (which implicitly call `.toString()` on objects passed, and do their own null checking before doing so).

cmgrote:
Let's agree on approach per discussion on the issue (#2446) and then I'll update the PR accordingly 🙂

mandy-chessell:
I thnk we are saying the same thing - the exception should validate that it is not being used with nulls and log the error.  This is to inform developers that they are using the class incorrectly and fix their code.  I think it is too much for the exception to throw an invalid parameter exception.

The debug logging shows there is an exception being create.  I don't mind it being removed for the right reason -ie it is not needed.  I do mind that it is being removed to allow incorrect use of the exception/

PR #2480, opened 2020-01-22
Open lineage flow

DimitriosMaimaris:
Add columns in main Graph for cases we do not have Lineage Mappings.


GuyAckermans:
Sleepycat is still in the POM, the removed one seems to have been a duplicate entry

planetf1:
Ah my error - i just realised. Apologies. I'd seen so many updates with sleepcat coming in and out in prior PRs I lost track... You are correct, so it's all good.

PR #2484, opened 2020-01-23
Updates anchorGUID description to resolve #2483

### LINKED ISSUE #2483: Further describe "anchorGUID" on SchemaElement
On today's call we discussed the intended behavior of `anchorGUID` in eg. SchemaElement.  We discussed that it should point to the `Asset` (when there is an `Asset` involved); however, the current types describe that it simply points to a `Referenceable`.  I believe we should create an update to the types to change this attribute's description to refer to the `Asset` level (when available) rather than a `Referenceable` in general?

(In other areas, particularly around discovery, that `anchorGUID` is used it does refer to `Asset` rather than `Referenceable`.)

From type archives 1.2:

```java
private EntityDef getSchemaElementEntity()
    {
        final String guid            = "718d4244-8559-49ed-ad5a-10e5c305a656";
        final String name            = "SchemaElement";
        final String description     = "An element that is part of a schema definition.";
        final String descriptionGUID = null;
        final String superTypeName   = "Referenceable";

        EntityDef entityDef = archiveHelper.getDefaultEntityDef(guid,
                                                                name,
                                                                this.archiveBuilder.getEntityDef(superTypeName),
                                                                description,
                                                                descriptionGUID);

        /*
         * Build the attributes
         */
        List<TypeDefAttribute> properties = new ArrayList<>();
        TypeDefAttribute       property;

        final String attribute1Name            = "anchorGUID";
        final String attribute1Description     = "Optional identification of the referenceable that this schema element is a part of.";
        final String attribute1DescriptionGUID = null;

        // ...
    }
```

cmgrote:
As discussed in call today and per issue #2483 -- though on thinking about it this is updating an existing property (just its description)...  I can't remember whether this is allowed via the type def patching (hopefully 🙏 🙂 )

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -23,6 +23,8 @@ jobs:
     steps:
       - name: automerge
         uses: "pascalgn/automerge-action@a4b03eff945989d41c623c2784d6602560b91e5b"
+        with:
+          args: "--trace"
         env:
           GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
           MERGE_LABELS: "automerge,!no-not-merge"


cmgrote:
> final String attribute1Description = "Optional identification of the referenceable that this schema element is a part of; should refer to an Asset when possible.";

does it not?

grahamwallis:
I assumed you were going to change it "optional identification of the asset...etc" 
Maybe that would be too much of a constraint?

cmgrote:
I understood from the call that the SchemaElement's `anchorGUID` could refer to non-Asset objects as well (hence the Referenceable endpoint rather than an Asset endpoint), so I was assuming we'd need the description to somehow cover both cases (since it's a single attribute used for either)...  Happy to rephrase the description to whatever you like, if you have a suggestion 🙂 

Now I'm wondering if it might be better to have an entirely new attribute that is something more like `assetGUID` that makes it explicit, for cases where an asset is involved?  (Though not sure what `anchorGUID` would refer to in those cases?)

grahamwallis:
I think you are correct - and it's fine.

mandy-chessell:
It is ok to add the attribute - but make it clear it is optional and is not set in the schema element that is directly linked to the asset.  If it is blank, then the process reading the schema needs to walk the entity/relaitonship chain

cmgrote:
> It is ok to add the attribute - but make it clear it is optional and is not set in the schema element that is directly linked to the asset. If it is blank, then the process reading the schema needs to walk the entity/relaitonship chain

Any strong preference between the two options (description change and re-use of `anchorGUID` vs adding a new, optional, asset-specific attribute)?

cmgrote:
I'll update per agreement from today's discussion (documented on the issue) once we're in a position to start modifying types under 1.5 🙂

cmgrote:
Closing this and will open new PR with agreed change to the description.

PR #2498, opened 2020-01-26
Make Egeria Graph DB configurable

lcpopa:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,8 +136,8 @@
         <jackson.version>2.10.2</jackson.version>
         <jackson.databind.version>2.10.0</jackson.databind.version>
         <kafka.version>2.4.0</kafka.version>
-        <janus.version>0.4.0</janus.version>
-        <gremlin.version>3.4.1</gremlin.version>
+        <janus.version>0.4.1</janus.version>
+        <gremlin.version>3.4.4</gremlin.version>
         <spring-boot.version>2.2.4.RELEASE</spring-boot.version>
         <spring.version>5.2.3.RELEASE</spring.version>
         <spring-security.version>5.2.1.RELEASE</spring-security.version>


open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/pom.xml
```diff
@@ -31,6 +31,12 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>kafka-clients</artifactId>
+            <exclusions>
```
planetf1: Can you clarify the reason for this exclusion - in the past a few issues have been hidden, only to resurface later, when excluding part of a dependency. It is sometimes needed of course, and a valid approach, but we should explain/document why

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
```diff
@@ -23,6 +23,18 @@
     <artifactId>graph-repository-connector</artifactId>
 
     <dependencies>
+
+        <dependency>
+            <groupId>org.janusgraph</groupId>
+            <artifactId>janusgraph-cassandra</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>hibernate-validator</artifactId>
```
planetf1: Same comment here on exclusions

pom.xml
```diff
@@ -137,6 +137,9 @@
         <jackson.databind.version>2.10.0</jackson.databind.version>
         <kafka.version>2.4.0</kafka.version>
         <janus.version>0.4.0</janus.version>
+        <snappy.version>1.1.7.2</snappy.version>
+        <antlr.version>3.5.2</antlr.version>
```
planetf1: Antlr 3 dates back to 2014 and is likely to have security issues. Please update to antlr 4 ie 4.8-1 or document why antlr3 is required

pom.xml
```diff
@@ -137,6 +137,9 @@
...
+        <joda.version>2.10.1</joda.version>
```
planetf1: Please update to 2.10.5 (otherwise dependabot will anyway) or document why an earlier version is required

pom.xml
```diff
@@ -137,6 +137,9 @@
         <jackson.databind.version>2.10.0</jackson.databind.version>
         <kafka.version>2.4.0</kafka.version>
         <janus.version>0.4.0</janus.version>
+        <snappy.version>1.1.7.2</snappy.version>
```
planetf1: Please update to 1.1.7.3 -- dependabot will kick in anyway

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
```diff
@@ -82,6 +124,12 @@
                                 <!-- Runtime dependencies for janusgraph-->
                                 <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-berkeleyje:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-lucene:*</ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-es:*</ignoredUnusedDeclaredDependency>
```
planetf1: WHy do we have these exclusions. they need to be clarified.

If a component is a runtime only dependency then I understand. I can see you recorded antlr as runtime in the pom - and the artifact is named runtime, so that makes sense -- but why the others?

There could be latent issues here that need investigation

lcpopa:
@planetf1  I am still working on the poms/dependencies to have them in a proper state and the project buildable. Merging is out of the question at the moment

planetf1:
> @planetf1 I am still working on the poms/dependencies to have them in a proper state and the project buildable. Merging is out of the question at the moment

Thanks for the confirmation. One tip - when creating the PR you can open it as 'draft'. It's not perfect as it doesn't notify reviewers, but it does make it clear the work isn't yet ready to merge

I've noticed some other old dependencies added too - like as asl mapper (now jackson databind). Ping me if you need any double-checking

Meanwhile I'm going to update the title as it's a little too easy to accidentally merge

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
lcpopa: Reverted, no longer applies

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
lcpopa: All these dependencies are runtime, used by janusgraph.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/pom.xml
lcpopa: Reverted, no longer applies

pom.xml
lcpopa: This version is required by janusgraph. Can test with a more recent one, but even if it is successful, I do not see it as a full proof solution

pom.xml
lcpopa: I found no other dependency to joda in the pom, and no error came from using 2.10.1 when using graph repo. Are you certain about the necessity of using a higher version ?

planetf1:
Targetting this for 1.4 as per request on status call today

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
lcpopa: This version of the validator will overwrite another and the result is Egeria not starting, failing with NoSuchMethod excepion, or similar

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/pom.xml
planetf1: Thanks. closing

pom.xml
planetf1: thanks - looking at options here. will update

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
planetf1: Reproduced:
```
Caused by: org.springframework.beans.BeanInstantiationException: Failed to insta                                     ntiate [org.springframework.validation.beanvalidation.LocalValidatorFactoryBean]                                     : Factory method 'defaultValidator' threw exception; nested exception is java.la                                     ng.NoClassDefFoundError: org.hibernate.validator.internal.engine.ConfigurationIm                                     pl (initialization failure)
```

planetf1: Looks like later versions of the validator cause spring issues. I'd concur that for now we go with the exclusion. When we repackage for the client/server model, or dynamically load all our connectors, then we can revisit.

pom.xml
planetf1: antlr3 is being used by a variety of components - janus itself, and then it's other packages ie for cassandra, tinkerpop, elastic search
I concur, let's go with that version. I'll take a look at the dependency reports for CVEs and if we need to do any further changes we'll do in a new PR

planetf1: snappy occurs in a few more places, including kafka. However we are close to current there, so less of a concern. Again, agree, let's go with this and can reevaluate the cve/dependency report after

pom.xml
planetf1: see above

open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
planetf1: Can you update the pom to add `<scope>runtime</scope>` for these - clarity and stops them being picked up at compile time.

planetf1: * joda time is pre-java 8 - but agree it's being used on our dependencies inc. tinkerpop and janus
Agree with others for now - 

Janus, cassandra , tinkerpop are bringing in rather a log of dependencies. I can see the rationale for all of these changes. When we are able to dynamically load - and also have the option of client/server, we'll be in a much cleaner state for the rest of the code.

For now, let's merge.. and then deal with any issues in later PRs as needed

planetf1:
@lcpopa just one minor update would be good I think - to make sure you make the runtime only dependencies as such in the pom for the graph connector.

Otherwise - see the above comment. Good job and apologies for the delay in reviewing..

planetf1:
Once updated & merged, If you need to get this into 1.4, you'll also need to create a PR against 1.4
 - checkout egeria-release-1.4
 - branch for your PR
- git cherry-pick -s <commitid of your change/changes>
- fix anything manually
- create PR as usual, but making sure to compare with egeria-release-1.4
 - note there is an enforced review before it can be merged

planetf1:
Can you clarify if you intend to create a 1.4 PR for this same change?

planetf1:
Confirmed on call 4 Feb that this is not needed for 1.4

If we can ideally update the pom to mark the runtime-only dependencies as scope=runtime it would keep us really crisp in our dependency definitions.

lcpopa:
The runtime dependencies have been marked appropriately.

PR #2502, opened 2020-01-27
Initial helpers for calculating differences between instances

### LINKED ISSUE #2486: Create general helper for diff'ing instances
Create a general helper that will identify differences between two instances (entities, relationships, etc) and provide some consolidated way of communicating / looking up just the differences.

cmgrote:
to resolve #2486

(See the issue for rationale on approach.)

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,23 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the ODPi Egeria project.
+---
+name: Check documentation links
+on:
+  schedule:
+    - cron: '0 1 * * *'
+jobs:
+  linkChecker:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Link Checker
+        uses: peter-evans/link-checker@v1
+        with:
+          args: -v -r *
+      - name: Create Issue From File
+        uses: peter-evans/create-issue-from-file@v2
+        with:
+          token: "${{ secrets.GITHUB_TOKEN }}"
+          title: Link Checker Report
+          content-filepath: ./link-checker/out.md
+          labels: documentation, consumability, tidy up


cmgrote:
One other very minor change I've included here is a new `PropertyContainer` interface that simply defines that instances that have InstanceProperties (ie. EntityDetail, Relationship) provide both a `getProperties` and `setProperties` method -- this will be helpful for areas where we might want to code something generic against an instances' properties, without having to create separate classes or methods for each of the different instance classes themselves.

(One example so far that I've hit: creating a properties comparator for ordering purposes for one of the repository connectors.  With this interface I can have a single method that implements the comparators rather than one method for EntityDetail and another for Relationship even though the logic of the implementation is identical because it just operates on the underlying instance properties.)

grahamwallis:
Don't EntityDetail and Relationship already provide getProperties and setProperties?

cmgrote:
> Don't EntityDetail and Relationship already provide getProperties and setProperties?

Yes, of course -- but not as part of a defined interface.  This minor change just adds the definition of this interface, with these 2 methods, and states that EntityDetail and Relationship both implement that interface (already).

cmgrote:
> No business logic should go into the instance objects - they must remain simple beans to avoid side-effects when Jackson is serializing/deserializing them. This is why we need the repository helper and verifier - they contain the logic that should be in the instance object if we could use OO design.

I was thinking that might be the case...  The downside to this would be that (without resorting to reflection) the properties that are distinct to each kind of instance then end up being managed in multiple places -- in the beans themselves, but also in the comparison methods (since they need to explicitly go through each instance property-by-property).  So any time the bean evolves, we need to also remember to evolve the comparison method (which will then live elsewhere in the code from the bean itself).  I guess we will have this potential challenge in other places as well given our strict bean-only approach, though, so better to be consistent?

> The comparison should only be between instance of the same category - so should only compare 2 relationships or 2 entities - do not need to compare an entity against a relationship. Therefore the repository helper should have methods such as compareRelationships/compareEntities/compareClassification and return a comparison object. This is in line with the existing implementation.

Yes, of course 🙂  Since many properties are actually the same at the header level (GUID, InstanceURL, Status, Type, etc) I was just going to short-cut a method that calculated those once rather than re-implementing that logic them for each different instance type (EntityDetail, Relationship, etc) that inherits from the header objects -- but I should be able to do that easily enough via some private helper methods as well (in a non-OO approach).

> I assume the compareRelationships would also check the relationship is pointing to the same entities at either end too?

👍 

> It would have been nicer to have a design in the issue rather than require reviewers to work through the code.

I tried to outline the basic idea on the approach briefly in the issue, sans diagrams, but I think sufficient to make clear that I was leaning towards this OO approach rather than the helper approach (?)  This was all I really wanted to discuss on the functionality call, to confirm whether the OO approach was one we wanted to take -- keeping in mind that if not I should be able to swap around the code without too much trouble.

I'll get to work on swapping it around now, no need to raise in the functionality call.

PR #2534, opened 2020-02-03
Update Janus & Gremlin (replaces dependabot #2521)

planetf1:
#2521 proposed a change to janus but failed during the build
This is because Janus now needs the current version of gremlin server.

Make this additional change

See comments on #2521


planetf1:
Builds cleanly, notebooks like ok with graph. CTS launched and running cleanly so far
Will update with final results.

planetf1:
The CTS (using graph repo) completed succesfully when run locally with this PR
 - 3936 tests, 0 skips/fails.

PR #2535, opened 2020-02-03
Type updates

### LINKED ISSUE #2483: Further describe "anchorGUID" on SchemaElement

### LINKED ISSUE #2508: Adding an "anchorGUID" property to Process
In further discussions from today's call, we identified that it would be useful to have something like the `anchorGUID` property on a Process, which could be used by a low-level (child) Process to point directly at its containing (parent) Process.

This would allow us to:

- Short-cut calculating this hierarchical relationship through what could be many relationships traversing many types in-between.
- Short-cut the determination that when a containing parent Process is deleted, it likely means that its contained (child) Processes should also be removed. **

However, I believe this approach would also differ from elsewhere we have `anchorGUID` since in this case it would be pointing to other instances of the same type.

**Are we amenable to adding such a property to Process (even if we call it something else, perhaps `owningProcessGUID`)?)**

** Taking DataStage as an example, this won't _always_ be the case (ie. at all levels).  For example, a Job owns the stages that it runs -- so in that case if the Job is deleted then any stages with the property set to that job's GUID should also be deleted; however, a Sequence (highest level) does not own the Jobs it runs, so deleting the highest level (Sequence) should not delete any Jobs that happen to be used within the sequence.  We may therefore want to only populate such an attribute where there is truly a containment / ownership relationship (in this example, between Stage and Job, but not between Job and Sequence)?

Also note that this will not preclude the fix that ensures we have multiple PortAliases and PortDelegation relationships between the different levels of Processes -- which will ensure that we can still figure out the parent-child relationship between eg. Sequence and Job in the example above even without this new property being populated (for the reasons outlined above).

cmgrote:
- Updates description of `anchorGUID` on `SchemaElement` to resolve #2483 
- Adds a new `ProcessHierarchy` relationship between processes to resolve #2508 

(Also documents the existing Ports and Lineage areas, for which there did not yet appear to be diagrams or markdown documentation.)


open-metadata-publication/website/open-metadata-types/0598-Lineage-Relationships.md
```diff
@@ -0,0 +1,24 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
```
mandy-chessell: These relationships really belong in Area 2 - they describe the linkage between assets and the linkages to other attachments.   All similar definitions are in Area 2.

cmgrote: They were already defined (from v1.2 onwards) as part of 0598...  Though I guess the area is really just down to the documentation (nothing inherent in the types themselves), so I could change this in the diagrams and markdowns (?)

cmgrote: (Or sorry, did you mean only the new ProcessHierarchy relationship should be in Area 2?)

cmgrote: Have pushed a new commit with the entire set of relationships moved to `0298` instead of `0598` + linked these new markdowns and diagrams from the Area-2 base markdown.

PR #2546, opened 2020-02-05
Data engine proxy and connector fixes to resolve #2536

### LINKED ISSUE #2536: DataEngineConnector/Proxy issues
This issue is about the DataEngineConnector definition located in module `data-engine-proxy-connector`.  This connector describes a plug point supported by the `data-engine-proxy-services`.  This means the classes and interfaces that support this connector should use the java package root of these services - that is they should be under:
```
org.odpi.openmetadata.governanceservers.dataengineproxy.
```
rather than
```
org.odpi.openmetadata.openconnectors.governancedaemonconnectors.dataengineproxy
```
where they are today.  This second package name would be where implementations of this connector would reside.

Another comment on this connector:
It passes the Data Engine OMAS's client to the connector.  This means:
   * the connector implementations are affected by any change to the Data Engine OMAS's client.
   * the connector implementation has to deal with server specific attributes such as the userId of the data engine proxy server.
  If the connector had a context object that wrapped the client:
     * it would offer a simpler interface to the connector implementation
     * the data engine proxy server could protect its connector implementations from changes
     * the data engine proxy server could ensure attributes such as the userId and externalMetadataCollectionId are set consistently on all requests.
This would be especially valuable since there seem to be userId properties throughout the property beans found in the model package.

cmgrote:
All of these changes should be transparent / have no impact on the way the proxy is configured with a connector, etc.

- Changes Java package for base connector implementation to match the rest of the proxy areas' packaging
- Removes beans that simply wrap a userId around the DE OMAS objects (uses the userId present on the configuration of the data engine connection itself instead)
- Replaces logged warnings and runtime exception throwing with the use of the audit log (logException)
- Makes mandatory methods in the interface abstract to ensure they are implemented, for others only produces logged information if they are expected to be implemented (ie. if the connector indicates that it requires polling)
- Moves the startup of the connector into the operational services themselves, before starting up a polling thread (previously the connector startup was inside the polling thread)


cmgrote:
Swap userId over to the `localServerUserId`

PR #2550, opened 2020-02-05
Open Lineage Services: Catch all possible exceptions

GuyAckermans:


open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/factory/GraphFactory.java
```diff
@@ -119,21 +117,20 @@ private JanusGraph initializeGraph(JanusGraph janusGraph, String graphType) thro
...
-     * @param type  - type to be labeled
-     *
+     * @param labels     - set of labels
+     * @param management - management instance of Janus Graph
+     * @param type       - type to be labeled
      */
-    private JanusGraphManagement checkAndAddLabelVertexOrEdge(Set<String> labels, JanusGraphManagement management,String type){
-        if(type.equals("vertex")){
-            for (String label: labels) {
+    private JanusGraphManagement checkAndAddLabelVertexOrEdge(Set<String> labels, JanusGraphManagement management, String type) {
```
lcpopa: This can be separated into different methods, with diminished parameters and specific responsibilities. Also, do not understand why management is returned, void can used. Streams can be used, even this method can be removed entirely.

lcpopa:
recheck

GuyAckermans:
@planetf1 There seems to be a problem with the build infrastructure, could you please have a look?

planetf1:
> @planetf1 There seems to be a problem with the build infrastructure, could you please have a look?

That's very odd - as you say it does look like an infra issue given a collision on the build ids (artifacts already existing).

I do note that future PRs are working ok so I would be inclined to retry. I'm not sure if just rerunning the job will help if the issue is a duplicate - though I will try.
If that fails I suggets pushing another change

@bramwelt any idea why we'd hit this issue with an artifact already existing?

planetf1:
I've started a new build at https://dev.azure.com/ODPi/Egeria/_build/results?buildId=4589&view=results - let's first see how this goes.

planetf1:
I was able to kick off an entirely new build pipeline - this worked perfectly.
I tried to rerun the hooks (which is how github knows the tests have passed), but this starts a new build, and I guess re-uses the same identifier and then fails.

I have no permissions on artifactory to manually clean up, so the most pragmatic suggestion I can make is that you do some kind of dummy change on the PR (or close/reopen) to kick off a fresh check which will most likely work.

GuyAckermans:
That worked, thanks!

PR #2548, opened 2020-02-05
RELEASE 1.3 - pom corrections for JCenter & maven central publishing

planetf1:
Signed-off-by: Nigel Jones <nigel.l.jones+git@gmail.com>

Misc pom changes to support publishing to maven central - #2499
 * Ensuring javadoc (empty) created for open types test generator
 * Adding missing packaging=pom to a few poms

Opening as draft pending further clarifications from jfrog

Issue #2499, opened 2020-01-27
Problems posting maven artifacts to maven central


planetf1:
Whilst our release 'shipped' on github 10 days ago (c. 17 Jan), and the release pipeline completed to push artifacts to jfrog artifactory & then bintray - from where they should be published to maven central, in fact the process hit some issues that we had no knowledge of.

Maven central publishing can take a few days for indexes to be rebuilt.

I raised IT issue https://jira.linuxfoundation.org/servicedesk/customer/portal/2/IT-18759?sda_source=notification-email 7 days ago ("IT-18759 ODPi Egeria - release completed in Azure. Not on maven central") and have had the following update from the support team:

The LF team are investigating updating the necessary pipelines to help with some of this:

Some further notes at:
https://confluence.linuxfoundation.org/pages/viewpage.action?pageId=41582704

In the meantime the artifacts were handled manually with the following issues arising:
```
All the 1.3 packages have been synced to Maven-Central with the exception of the following:

New Packages Have been Requested to be included in JCenter and are waiting on approval from JFrog:

audit-log-event-topic-connector
audit-log-slf4j-connector
cassandra-data-store-connector
cassandra-metadata-extractor-connector
data-platform-services-connector
discovery-service-connectors
gaf-metadata-client

open-connector-archives
Packages Missing JavaDoc:

open-types-test-generator
Packages Missing Sources:

governance-action-connectors
stewardship-action-connectors
Packages with Invalid POM file:

gaf-metadata-management
ui-chassis
Packages with Corrupt JAR:

open-lineage-janus-connector
ui-chassis-spring
I'll attempt to redistribute open-lineage-janus-connector and ui-chassis-spring from Artifactory to see if that resolves those package issues. Once we get confirmation from JFrog on the requested packages for JCenter we'll try syncing those out as well.```

planetf1:
I should have access to jcenter to help in future (if anyone else needs this add a note here - @cmgrote )

* We need to figure out the process around adding a package. Can we detect during the normal dev process, or is it left to final validation? 
* We do try and always build javadoc - though there is a discrepancy above? Where did this go wrong
* We also add sources - how did this fail?
* What is an invalid pom when we don't see build issues
* Can we automatically fix corruption by re-posting?? And was this error transient or permanent?

As we formulate answers to these questions I'll raise any necessary PRs for quick fixes or issues for longer discussions, 

cc: @mandy-chessell  @cmgrote @bramwelt

planetf1:
[...]

Issue #2477, opened 2020-01-22
Create release 1.4


planetf1:
* Branch for release 1.4 (last week of Jan)
* Update master to release 1.5
--
When 1.4 ready
* Update release 1.4-SNAPSHOT to 1.4
* update release notes
* final testing
* ship to maven 
* create release on github

planetf1:
@mandy-chessell I plan to update the release notes in master. I need to move 'dedup' from 1.4 forwards. Probably 1.5? I'll also move 1.4 to the current release

planetf1:
Agreed on team call 4 Feb that no more fixes needed for 1.4 so will move forward on final testing & creating the release

planetf1:
Release 1.4 is now published. closing

PR #2570, opened 2020-02-10
RELEASE 1.4 - updates for release

planetf1:
See #2477

This change

Moves the version from 1.4-SNAPSHOT to 1.4
removes the release-notes and replaces with a link to master
Note: ideally waiting for a fix for #2499 - PR for 1.3 is #2548) to allow for complete publishing to maven central.If this doesn't arrive in the next 2 days will do a partial fix & release


planetf1:
Release 1.4 is now ready for deployment. @mandy-chessell 

I will push through the release cycle BEFORE adding the git archive so that we can flush out any issues on this cycle.

Issue #2574, opened 2020-02-10
Only run actions on odpi/egeria - not forks


planetf1:
We're seeing actions run against all forks of our repository. Some appear to work, other's (like the linkchecker) fail.

This causes noise, and to stop requires every fork owner to disable actions.

From https://help.github.com/en/actions/automating-your-workflow-with-github-actions/contexts-and-expression-syntax-for-github-actions it looks as if we could use a conditional expression to only run the action in question when the repository is 'odpi/egeria' (or similar for our igc/atlas connectors).

We need to be careful this doesn't cause any issues with PRs -- since the fork itself is owned by a user other than odpi.

I think this is worth trying out.

I will propose a change
cc: @cmgrote @grahamwallis

planetf1:
This added on a step would work
```
if: github.repository == 'odpi/egeria'
```
more more generally
```
if contains('github.repository','odpi/')
```
looks better

Seems to be needed on each step to be skipped...

planetf1:
Doesn't seem to be possible on entire action, though a cleaner way to avoid conditional clutter is covered in https://github.community/t5/GitHub-Actions/conditional-jobs-and-workflows/td-p/32166

planetf1:
After merging, the overnight schedule run of the link checker still ran on my fork

I'm going to propose additional changes to add the conditional at the <step> level.

cmgrote:
> After merging, the overnight schedule run of the link checker still ran on my fork

... and somehow it looks like it did _not_ run on the master repo (?)

planetf1:
I can confirm this is working as expected.
 - odpi/egeria - stale/linkcheck actions ran successfully
 - planetf1/egeria (and anyone else's fork) - actions run but are a complete no-op and do not execute steps.

There's no way to completely stop an action from running (unless disabled for the forked repo), but this is a decent compromise.

cmgrote:
> I can confirm this is working as expected.

I don't think it is -- looking at the detailed logs, the steps that had the condition against them are entirely skipped _on master_ as well: https://github.com/odpi/egeria/runs/440017789?check_suite_focus=true

I think this is related to the conditional doing a literal-value check rather than checking the value within a variable?  See PR #2584

planetf1:
Last night actions ran completely on forks+odpi
Submitted new PR to resolve

PR #2575, opened 2020-02-10
#2574 only run actions on odpi repo

planetf1:

See #2574

Adds a conditional execution on both our 'stale' and 'linkchecker' github actions to only execute if the repository is 'odpi/*'

See issue for more details
cc: @grahamwallis @cmgrote - please review - not tried this before and it's trickier to test actions


planetf1:
Ok it's in as low-risk - worst case the actions don't run/error.
As they're scheduled we'll find out the effect overnight.

Issue #1734, opened 2019-10-13
Add support for PR build of egeria docker image


planetf1:
It would be really useful if, after building egeria, we could also run the egeria (only) docker build and publish to an alternate registry with a PR-related tag - this would aid in testing of PRs in a notebook/k8s environment. This should now be feasible given the speed images can now be created

NOTE: An alternative would be to modify the egeria docker image to pull the correct maven artifact specific to the PR in question.

cmgrote:
In terms of an alternative registry, this would probably be a good scenario to use our Artifactory setup -- which can also act as a Docker repo.

planetf1:
#2579 has made improvements to the docker build, but does not directly enable a PR build.

planetf1:
As our code continues to develop, with more collaborators, and more components such as the new UI, it's becoming more important that any contributor can easily test that their changes do not break out lab notebook environment or docker images.

Building a new docker image per PR is relatively quick (seconds) and could make it much easier than building manually, and avoids any extra overhead on the developer

planetf1:
still useful idea

planetf1:
Posted discussion topic at https://github.com/orgs/community/discussions/40818

Issue #1595, opened 2019-09-19
Public Docker Hub images versioning


cmgrote:
At the moment our various Egeria-owned Docker images publicly available via Docker Hub are using a `1.1` tag, despite Egeria itself not having a 1.1 release...  I think this is confusing / problematic from a consumption perspective?

Each of these also has a `1.1-SNAPSHOT` tag: not clear why we're not updating this version alongside `:latest` for consistency across our unreleased artifacts' versions?

- https://hub.docker.com/r/odpi/egeria-configure/tags
- https://hub.docker.com/r/odpi/egeria/tags

(Note that the images for other projects, like Apache Ranger and Apache Atlas use `:latest` and the version number for those projects' releases on which the images are built.)

planetf1:
I agree - we should align the versioning along with the code

Note I am currently fixing an egeria image build break - should be sorted in next day.

planetf1:
Adding some notes as we prep for 1.1 release

The docker image builds typically have lines like
```
# Default to latest snapshot from official odpi repository
ENV version ${version:-"1.1"}
ENV repo ${repo:-"https://odpi.jfrog.io/odpi/egeria-snapshot/"}
```
to retrieve artifacts. This is fine during development - but here we've moved on from snapshot just before release. Arguably part of our final stage prior to release is to update this repo first to snapshot, then to staging. This could be done here, or in the pom.

Then in the pom itself we tag 'latest' which we only want to do with master, not all releases.

I will patch up manually for the 1.1 release, whilst we consider automating this for the next release.

planetf1:
Additionally, we may want to seperate out where the docker images go - technically my quick patch will push a 1.1 release to dockerhub early (whilst being 'staged')

planetf1:
I think we'll need to revise the naming scheme along the lines of

* Only tag master builds as 'latest'
* Use tags of the form ${project-release}_${component.version}
* For build images use ${project-release}_${component.version}-build
* Do the same for gaian, ranger, atlas 
* Use egeria-${component-name} for the repository name

Do this in 1.1 & master
We then update the Dockerfile (where build images used), docker compose & helm charts accordingly
For now stopping 'latest' will be a manual removal in the branch

What I think we end up with is the 1.1 build resulting in 
egeria-apache-atlas:1.1_2.0.0
egeria-apache-atlas:1.1_2.0.0-build
egeria-apache-ranger_1.2
egeria-apache-ranger_1.2_build
egeria:1.1
[...]

PR #2579, opened 2020-02-10
Improve docker images & fix distribution

### LINKED ISSUE #2215: User interface is missing from egeria packaging
_'a clean build of 1.3 does not produce access-services-user-interface-1.3-SNAPSHOT.jar as I expected it to.  In fact the user-interface directory is missing from the egeria-1.3-SNAPSHOT-distribution/egeria-omag-1.3-SNAPSHOT  directory'_

-- User report from: https://odpi.slack.com/archives/CBPRDQPRC/p1576056215113300

My response

_we are doing a lot of work on it currently to make it useable. This includes refactoring/renaming which unfortunately is what caused the problem you saw :-(
One alternative way to run is to either
 a) Use our egeria docker container (the docker compose & k8s examples show how to do this - I can get more specific it if helps)
 b) Launch via java -jar ~/src/egeria/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/target/ui-chassis-spring-1.3-SNAPSHOT.jar  Note that in fact the UI currently needs to have some basic config provided for the server url and server name. Personally I start it with OMAS_SERVER_URL=http://localhost:8081 OMAS_SERVER_NAME=cocoMDS1 java -jar ~/src/egeria/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/target/ui-chassis-spring-1.3-SNAPSHOT.jar but this will depend on your configuration. Soon the configuration of the UI will change quite a bit more._

### LINKED ISSUE #2375: Use assemblies in containers
Currently our docker image for egeria is built by pulling a few specific files (jars) from the build.

As we move to making egeria load more capability dynamically - not just connectors, but OMASs etc too - and depend less on more 'static' maven dependencies, we should start using the same assemblies that we tell clients to use ourselves....

This will likely mean we'll need to  store the built archive as an artifact in the build environment - for example from the merge build - so that it can then be used by the docker build. 

This will ensure that our tutorial environments, and any additional similar setups we use for testing are most representative.

planetf1:
Improvements (changes?!)  to docker image build
 * Now uses the 'proper' egeria distribution - the same as we tell others to use, rather than copying files . Fixes #2375
 * Removed javadoc & source jars from the main distribution - for runtime these are not needed & add to the size. 
 * Added the UI (missing) Fixes #2215
* Added Janus connector for OLS (missing)
 * An image is now self-contained. Rather than downloading the jar from jfrog, we now build in-situ. This makes it easier to build a dev image for testing, in future for PRs if we wish, as well as easier to integrate into a build (merge) without complexity around intermediate repos - will help with #1630
 * Moved 'latest' tag to an optional profile (-Ptaglatest or -Dtaglatest) - this makes it easier to build docker images for a release or master, and then only use latest on master - so we can have good 1.3, 1.3-SNAPSHOT etc images as well as latest, and in future refer from our charts/compose . Should also help with #1595 - as will other changes
 * Use OCI spec metadata instead of label schema - more cross platform standard
 * Inject git metadata into image metadata - much easier to answer 'what am I running' - have had issues with this. We also record the repo & if dirty as well as build/latest git update time.  Set to 'unknown' if built outside
 * set LOADER_PATH by default for loadable connectors
 
May help with #1734 in future

Draft for now - will need to test more, check charts, but opening for review comments


planetf1:
Further changes
 * Updates 'configure' image to follow same pattern as for 'egeria'
 * Updates vdc & lab charts to work with new docker images
 * drops omrs monitor from vdc chart
 * Allow profiles to select whether to build core images (egeria, configure) , extended images(others) or all (by default to preserve existing behaviour)
 * removes build of extended images from daily build
 * adds build of core images to merge build - so that new, self-contained images will now be built immediately after a merge

This is probably enough changes for this PR (still to do - notebooks). Open for comments @grahamwallis @cmgrote @bramwelt . There's a fair chance something will go wrong. will monitor and add new PRs as needed - esp any impact on merge build. PR build not touched

Tested notebooks in k8s & compose, that correct build profiles are used for images, that images are tagged correctly, and that vdc chart deploys properly with igc etc.

planetf1:
@cmgrote I've tried not to change anything around connectors -- if we can get the basics working ok, then we can think about how to handle the connectors.

There are broader changes possible in future - like moving out the extended images into a different repository. I also could have gone more radical in using different docker repos (maybe artifactory) or publishing the omag archives to be picked up in a diff pipeline, but I opted for the simple approach

bramwelt:
@planetf1 This all looks good, and will be a good basis for more improvements!

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
```diff
@@ -57,7 +57,7 @@ spec:
           image: "{{ if (.Values.image.egeria.registry | default .Values.imageDefaults.registry) }}{{ .Values.image.egeria.registry | default .Values.imageDefaults.registry }}/{{ end }}\
                   {{ if (.Values.image.egeria.namespace | default .Values.imageDefaults.namespace) }}{{ .Values.image.egeria.namespace | default .Values.imageDefaults.namespace }}/{{ end }}\
                   {{ .Values.image.egeria.name }}\
-                  :{{ .Values.image.egeria.tag | default .Values.imageDefaults.tag }}"
+                  :{{ .Values.image.egeria.tag | default .Values.egeria.version }}"
```
cmgrote: Is this the right way around?  I would expect to take `.Values.image.egeria.version` and only if that is empty to default to the default tag (which I would expect would always be `latest`)...

Even then, though, we need to be careful here keeping in mind the specific semantics of `latest` and the pull policy (ie. this makes it further critical that the default pull policy remains `Always` so that non-`latest` images will be forcibly replaced by k8s and it won't leave stale / old `x.x-SNAPSHOT` images around)...

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
```diff
@@ -74,7 +74,7 @@ spec:
           image: "{{ if (.Values.image.configure.registry | default .Values.imageDefaults.registry) }}{{ .Values.image.configure.registry | default .Values.imageDefaults.registry }}/{{ end }}\
                   {{ if (.Values.image.configure.namespace | default .Values.imageDefaults.namespace) }}{{ .Values.image.configure.namespace | default .Values.imageDefaults.namespace }}/{{ end }}\
                   {{ .Values.image.configure.name }}\
-                  :{{ .Values.image.configure.tag | default .Values.imageDefaults.tag }}"
+                  :{{ .Values.image.configure.tag | default .Values.egeria.version }}"
```
cmgrote: One example (others in solr, egeria, ibm-igc, and the init jobs) -- for consistency this needs to be `.Values.image.egeria.tag` rather than `.Values.image.configure.tag` (to pickup the singular Egeria version rather than a configure-specific tag); though see previous comments about whether this condition is not backwards at the moment...

planetf1:
@cmgrote thanks for the comments. will correct & update.

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
planetf1: On the version choice:

It was my intent to do it this way around. 

My reasoning - it is generally sensible only to use the image version that is the same as that of the chart. We ship that as 1.5-SNAPSHOT. Since we're moving to a more release-specific model I do not believe 'latest' makes so much sense. This is why the expression no longer uses the imageDefaults for the tag.  However I still wanted to allow the specific image version to be overriden on demand which can optionally be set.

planetf1: I did anticipate the second part -- the vdc chart does now use 'Always' as the default setting for exactly that reason.

I did consider whether to add an expression to use 'Always' for snapshots and 'ifNotPresent' for release, but was keen to make significant steps forward in the general aspects around release specific images, whilst maintaining compatability with our existing charts - even though vdc for example, is somewhat superceeded by events and will be reworked around the 1.8 time so wanted to move forward. We can refine the setting of that value through an additional issue/PR if you feel it would be helpful?

The whole intent of this change is to make -SNAPSHOT something only us devs on the team use, whilst the 'masses' use specific released versions. 

As a mitigation we could change the flag in values as the point we cut a release ? That vaguely crossed my mind.

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
planetf1: Actually this was my intent - to allow the configure image tag to be overriden if required, but generally to default to the release version, as encoded by default in values.yaml.

So normally the 1.5-SNAPSHOT chart will be using the 1.5-SNAPSHOT docker image (which itself is made available very soon after merge - if the PR works!)

So it's not consistent with ranger, atlas - but intentionally so. The versioning of those images may need further work, but there's not an immediate pressure as we aren't really revisiting that scenario until 1.8 - of course that may change, but if so will revisit them then.

For now the old versioning and build process remains in place... and when we change it will be more radical than just a version change, I think we'll pull the build etc completely out into a seperarate third party repo.

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
cmgrote: Fine for now, just need to keep in mind for down the road...

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
cmgrote: I'm not sure I follow...  This would mean that within the same chart we'll end up with different versions of the `configure` image being deployed into different pods, which I think is confusing (and potentially error-prone, ie. if we delete a particular version or fix a bug in a particular version but there is then a mix of versions deployed)?  These aren't the versions of the ranger, atlas, solr etc images themselves -- just the configure image that contains various utilities (curl, etc) used for running setup / initialization.

I would suggest we keep the same version of this configure image across everything.  This would also mean that as we upgrade that image with new base images (alpine) or other fixes to eg. vulnerabilities that we gain those fixes across all places `configure` is used rather than just the core Egeria ones.

(Even if we break these other images out into their own repo, they'd still need such a `configure` image to do their configuration, which I'd suggest re-using from Egeria core rather than creating a side version of.)

planetf1: I agree we wouldn't want mixed versions of a specific image in a deployment

I've gone through all the yaml's again.

Each time we use the egeria image I use the form
```
:{{ .Values.image.egeria.tag | default .Values.egeria.version }}
```
Each time I use the configure image I use:
```
:{{ .Values.image.configure.tag | default .Values.egeria.version }}
```

My build changes have changed/improved the build of these two images - configure & egeria. They are built after merge. Both are built with the version of egeria they are built with (this can be overriden, they are also tagged as latest for master only)

The default values.yaml sets 'egeria.version=1.5-SNAPSHOT', which is aligned with the release currently in master. 

Thus we will use this version for both egeria and configure across both charts & all jobs/pods.

The user may also set either, or both of
 image.configure.tag
 image.egeria.tag

If set these override the 'inherited' egeria.version set above (I name it like this to be very clear it's a common setting not image specific). They can be different, but each time configure is used it will always be the same. Ditto for egeria.

There's no change to the way our other images are used (though I do plan to change notebooks in the near future)

Did I miss an update?
Have I interpreted the golang expression wrongly?

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-atlas.yaml
```diff
@@ -76,7 +76,7 @@ spec:
           image: "{{ if (.Values.image.configure.registry | default .Values.imageDefaults.registry) }}{{ .Values.image.configure.registry | default .Values.imageDefaults.registry }}/{{ end }}\
                   {{ if (.Values.image.configure.namespace | default .Values.imageDefaults.namespace) }}{{ .Values.image.configure.namespace | default .Values.imageDefaults.namespace }}/{{ end }}\
                   {{ .Values.image.configure.name }}\
-                  :{{ .Values.image.configure.tag | default .Values.imageDefaults.tag }}"
+                  :{{ .Values.image.egeria.tag | default .Values.egeria.version }}"
```
cmgrote: This value is different from the others -- there were one or two other spots with this setup, too.  Here you have:

```
:{{ .Values.image.egeria.tag | default .Values.egeria.version }}
```

whereas elsewhere you have:

```
:{{ .Values.image.configure.tag | default .Values.egeria.version }}
```

I had assumed you meant to use the former everywhere (having a single tag for all Egeria images: both the Egeria core and the configure image), but perhaps it's these few places with the first setup that were a mistake (?)

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
cmgrote: > Each time I use the configure image I use:  :{{ .Values.image.configure.tag | default .Values.egeria.version }}

I think I misinterpreted which was correct, then -- there are a few places where you're setting up the configure image differently than this (see example in comment below on the apache-atlas template)

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-atlas.yaml
planetf1: Thanks for clarifying. You are correct - I did have an error in that specific atlas file - but it seems the only one.

But I had intended to allow override on the two images independently, so the first example you list above is used for egeria, the second for configure. Both default to the egeria version

I've just added the additional commit

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
planetf1: I think it was just one? Should now be corrected.

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-atlas.yaml
cmgrote: Leave it to me to find the one that is wrong and assume it is correct and that all the others are incorrect 😝

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-vdc/templates/apache-ranger-admin.yaml
cmgrote: I went back to double-check and couldn't find another, so I think you're right!

planetf1:
Phew! I'm going to merge and see what happens... ;-)

planetf1:
All seems good after merge
 - egeria/configure build as part of merge job & pushed to maven central with both latest and 1.5-SNAPSHOT labels
 - vdc & lab environments working fine (UI, notebooks, IGC, viewing terms etc)
 - all containers/jobs look started ok.

Note that the other images are built by the original job now - it actually runs after a merge too (was changed a few weeks ago), though that can be refined when we do more scenario work.

PR #2581, opened 2020-02-11
#2574 move conditional on jobs to steps

### LINKED ISSUE #2574: Only run actions on odpi/egeria - not forks

planetf1:

I've copied the condition to steps. Decided to leave in-situ on jobs because there are some indications that should work - and plenty saying it should so may work at some point? !

Fixes #2574

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -7,7 +7,6 @@ on:
     - cron: '0 1 * * *'
 jobs:
   markdown-link-check:
-    if: contains('github.repository','odpi/')
     runs-on: ubuntu-latest
     steps:
       - name: checkout action code
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,7 +9,6 @@ on:
 
 jobs:
   stale:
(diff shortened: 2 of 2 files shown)


planetf1:
> > Decided to leave in-situ on jobs because there are some indications that should work - and plenty saying it should so may work at some point? !
> 
> I'm wondering if we should remove from the job-level, given that the actions appear to have _not_ run on the master repo (and maybe this was somehow the cause?)

Oh weird... Fair point - could they have really got the conditions inverted? .... I'll remove the job level

PR #2584, opened 2020-02-12
Action conditional check

cmgrote:
I think I spotted the error -- I think the conditions we had in there were checking the literal string `'github.repository'` rather than the context variable `github.repository`, and since the literal string `'github.repository'` does not ever contain the literal string `'odpi/'` within it, the conditional always evaluated to false.

Trying this change (and also bumping back up to a job-level conditional) to hopefully fix...


planetf1:
yep good spot. Now let's see if job level works....

PR #2585, opened 2020-02-12
Add <scm> link to poms 

planetf1:
See more details in #2499

This adds the <scm> link to github to all egeria pom files


planetf1:
Proposing to merge this shortly. Since it's a fairly global change it's prone to conflicts. As it's only a build change it's safe.

Issue #1514, opened 2019-09-10
create helm repository for chart sharing


planetf1:
The egeria-palisade project needs to make use of our 'lab' helm chart
To enable this easily, the egeria lab chart needs to be served via a helm repo as per https://helm.sh/docs/chart_repository/

Ultimately this needs to be done as part of the build process, to ensure these charts are always current. However for expediency in integration the generated file is being checked in to enable this reuse quickly and will be refined in time.

For now only the 'lab' chart is exposed.

Part 1 - create repo (can only be tested after merge due to the way github pages works....)

planetf1:
See https://github.com/kiwigrid/helm-maven-plugin

planetf1:
For integration I think we will focus on a core egeria operator, which other solution charts can use.
We should have such a solution chart combining Palisade with egeria, so moving this to samples repo

planetf1:
Keeping open - we should determine where our Helm charts are published to

planetf1:
**Update June 2021**

When testing the postgres connector I wanted to 'package' up the deployment I had created for future re-use, and to act as an interesting sample. This is in addition to the 'odpi-egeria-lab' and 'egeria-base' charts we have, and the 'vdc' chart we used to use. We also have a test chart for postgres testing (not currently used)

I figured the 'egeria-base' (or lab) could be a good starting point, to which I just wanted to 
 - use a custom image so I also got the connector (other approaches like wget are feasible)
 - add in a postgres deployment (helm chart) configured to use a sample data source
 - extend the configuration with additional startup scripting to setup the data manager & integration daemon
 
 I did this manually by installing the egeria-base chart, a postgres chart, then issueing the right REST requests
 
 I can't simply write a new chart for postgres which depends on one of our other charts since at this point:
 - Helm does not support installing charts directly from git, only from a local filesystem, or packaged in a repository (though there has been a PR being worked on for a few years which is getting closer https://github.com/helm/helm/pull/9482 )
  - There is no longer a public chart repository
  - charts need packaging and indexing as part of a build process
  - I need somewhere to push them too
  
  This means I have quite a specific workflow specific to helm charts. Just the same as base egeria is primarily java (with a little docker), and the k8s operator is in go - those sit best in different repositories so the CI/CD process can be customized to their needs.
  
  There are some evolving examples of automation of this form for helm ie https://github.com/marketplace/actions/helm-chart-releaser . This would allow us to publish charts to github pages for direct installation via URL but I note
[...]

PR #2589, opened 2020-02-13
Improve build, performance, resilience of notebook environment

### LINKED ISSUE #1639: Incorporate git notebooks into docker build more tightly.
Currently when we deploy the notebook environment via docker-compose or k8s
 - the egeria code comes from the last docker build
 - the notebooks are downloaded on first-use via git.

One problem I've seen is that as the code (in master) is evolving rapidly along with the notebooks, sometimes the 'package' will break. This has varied from minutes to days (if the docker build doesn't work) but is typically < 1 day.

Recently the build of the egeria docker image has been made a lot quicker (#821) and ideally will be built after merge (#1630) which narrows the window, as long as the user keeps updating  the image - a suggestion that was implemented via docs/defaults in #1617 

the current blog proposal #1492 covers this scenario

However one way of avoiding this is to 'bake in' the notebooks as part of build process -this could be done in a few ways
 - by strong version identification - so that we pull from git by sha hash , or a specific maven snapshot revision
 - by including the notebooks within the image (we might need to build a custom notebook image layered on the base one for this).

This is certainly doable (and interesting timing given other changes in this area I'm working through)

Any comments @cmgrote @mandy-chessell 

My git feel is to go ahead with getting the blog post off, the notebooks working as best they can right now, and only then change - what we have is pretty good & ready for more people to use, and then reconsider along with other feedback.

### LINKED ISSUE #1640: Create a 'zip' version of notebooks (compose & k8s)
When we ship the next release of Egeria (1.1, or a later release) consider if, as part of the assets made available to interested parties, we create simple bundles for
 a) docker-compose tutororial environment
 b) k8s chart for tutorial

so that a user can download a simple .zip to get started without even having to use git to retrieve the (few) files needed

Note that b) may be unnecessary if helm charts are published to a central repo as per #1514

Followup from @cmgrote suggestions in #1634

### LINKED ISSUE #1776: Docker-compose tutorial environment fails on Windows
When running the docker compose lab environment via `docker-compose -f ./egeria-tutorial.yaml up` on windows, the notebook server fails to start with:
```
notebook_1   | Starting to load open metadata labs from github
notebook_1   | Hit:1 http://archive.ubuntu.com/ubuntu bionic InRelease
notebook_1   | Hit:2 http://security.ubuntu.com/ubuntu bionic-security InRelease
notebook_1   | Hit:3 http://archive.ubuntu.com/ubuntu bionic-updates InRelease
notebook_1   | Hit:4 http://archive.ubuntu.com/ubuntu bionic-backports InRelease
notebook_1   | Reading package lists...
notebook_1   | Reading package lists...
notebook_1   | Building dependency tree...
notebook_1   | Reading state information...
notebook_1   | E: Unable to locate package git
```
The notebook server is then not running, so the URL http://localhost:18888 cannot be reached, and the tutorials cannot be used.

This occurs (in my case) on Windows 10 (18999) with docker for windows 2.1.4.0 (docker engine 19.03.3)

However the following environments work perfectly

Linux - Ubuntu 19.04, docker 19.03.4-rc1
MacOS - Catalina, docker 19.03.3

cc: @CDaRip2U

planetf1:
Redistributable assemblies
 * Creates a new assembly for docker compose, helm charts & notebooks
    - this is a first pass. We may want to rename, or change contents in future. For now
       it is primarily used to import the notebooks into the docker build environment
       Fixes #1640 

Reliability improvements
 * Creates a new 'jupyter' docker image
   - notebooks from the build are now baked in to the image (and therefore consistent with the egeria code). Fixes #1639 
    - notebook is versioned, 1.5-SNAPSHOT - aligned with our other core images
    - build is done alongside egeria/configure images as part of the merge build
    - Avoids unreliability & time taken to installl packages, download notebooks (which would also be out of sync with the code) 
    - works with restricted network connection - no need to go to git
    - No longer requiring a data volume means the Docker compose environment now works on windows again - Fixes #1776
    - docker-compose environment updated accordingly. This removes the need for the post-launch script
    - odpi-egeria-lab k8s helm chart updated. This removes the need for initContainers, volume, and a customized launch script

Cleanup
 * Removes old ranger chart (not used - an initial experiment). Ranger still forms part of the vdc chart - unchanged
 * Removes old helm index (index.yaml) - need to address #1514 to do this properly

Tested notebooks (configure/start) in
 - k8s lab chart
 - docker-compose (macOS)
 - docker-compose (Windows)

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -7,15 +7,17 @@ on:
     - cron: '0 1 * * *'
 jobs:
   markdown-link-check:
-    if: startsWith(github.repository,'odpi/')
     runs-on: ubuntu-latest
     steps:
       - name: checkout action code
+        if: startsWith(github.repository,'odpi/')
         uses: actions/checkout@master
       - name: Check links
+        if: success()
         id: check_links
         uses: cmgrote/github-action-link-check@master
       - name: Create issue from file
+        if: success()
         uses: peter-evans/create-issue-from-file@v2
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -9,10 +9,10 @@ on:
 
 jobs:
   stale:
-    if: startsWith(github.repository,'odpi/')
     runs-on: ubuntu-latest
     steps:
       - name: check stale issues and prs
+        if: startsWith(github.repository,'odpi/')
         uses: actions/stale@v1
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/ffdc/ConnectionCheckedException.java
+++ b/open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/ffdc/ConnectionCheckedException.java
@@ -5,7 +5,6 @@ package org.odpi.openmetadata.frameworks.connectors.ffdc;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 /**
  * ConnectionCheckedException provides a checked exception for reporting errors found in connection objects.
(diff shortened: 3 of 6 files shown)


planetf1:
( github status is 'stuck'. pushing a no-op to kick)

PR #2603, opened 2020-02-17
Add first draft of admin guide

mandy-chessell:
This PR provides the first draft of the admin guide plus updates to the admin services to improve their consumability.  There is also an initial implementation of the admin services clients.  These are untested as yet so should not be used.


open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources/application.properties
```diff
@@ -12,7 +12,7 @@ strict.ssl=true
 
 #userId used to startup the list of configured servers default is 'system'
 startup.user=system
-#CSV of servers to be started
+# Common separated names of servers to be started
```
planetf1: minor though should be 'comma'

PR #2608, opened 2020-02-18
Security process

### LINKED ISSUE #1386: [vulnerability_report_process] define vulnerability reporting process
(This is required for lowest level **passing** for CII.)

The project MUST publish the process for reporting vulnerabilities on the project site. (URL required for "met".)

E.g., a clearly designated mailing address on https://PROJECTSITE/security, often in the form security@example.org. This MAY be the same as its bug reporting process. Vulnerability reports MAY always be public, but many projects have a private vulnerability reporting mechanism.

If private vulnerability reports are supported, the project MUST include how to send the information in a way that is kept private. (N/A allowed.) (URL required for "met".) [vulnerability_report_private]

Examples include a private defect report submitted on the web using HTTPS (TLS) or an email encrypted using OpenPGP. If vulnerability reports are always public (so there are never private vulnerability reports), choose "not applicable" (N/A).
The project's initial response time for any vulnerability report received in the last 6 months MUST be less than or equal to 14 days. (N/A allowed.) [vulnerability_report_response]

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#vulnerability_report_process)

Furthermore (required for **silver** level):
The project MUST have a documented process for responding to vulnerability reports.

This is strongly related to vulnerability_report_process, which requires that there be a documented way to report vulnerabilities. It also related to vulnerability_report_response, which requires response to vulnerability reports within a certain time frame.

This is inspired by Apache Project Maturity Model QU30.

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/other.md#vulnerability_response_process)

cmgrote:
Documents our security vulnerability handling process, which should resolve #1386, and was agreed on during our initial security call.


cmgrote:
@mandy-chessell one point in here that is currently a placeholder is a separate mailing list for reporting security vulnerabilities (privately), for those who are not part of the security call itself.  Would you be able to setup such a list for us at `lists.odpi.org`?  I was thinking something like `odpi-project-egeria-security@lists.odpi.org` for consistency.

(If there's a way to put myself as an admin for that list specifically, I'm happy to help maintain the receivers.)

I can then link that into the docs above and we should be all set.

mandy-chessell:
John suggested that we use the maintainers mailing list as we are still a small group

cmgrote:
> John suggested that we use the maintainers mailing list as we are still a small group

Fair enough -- what's the email address for that list (as I don't seem to see it on lists.odpi.org)?

PR #2612, opened 2020-02-18
Fix repository proxy classification and add admin client sample

mandy-chessell:
This PR covers the testing of server start up and configuration around the metadata server.  The testing revealed a case where a bad connector provider class name coming from the client is swallowed and only a log.error() is made.  Now returns a proper exception that is received in the client.

Also changed the server type classifier to recognize the repository proxy.  I have been able to test that the other server types seem to be ok - but need @cmgrote  to test that his real repository proxies work


cmgrote:
So this gets further along now, but proxy doesn't quite get fully setup as it bombs out with a non-existent server name error at some point -- though I suspect this is more likely related to #2614 (?)

Will raise a separate specific issue.

Issue #1682, opened 2019-10-08
Glossary View OMAS not supporting FFDC


mandy-chessell:
Egeria is a middleware product.  This means it is embedded in other tools.  As a result it needs to have a certain level of resilience to avoid and correct errors.  Where this is not possible, and it needs to report an error, there are standards for how errors are reported.   These standards follow the principle of first failure data capture (FFDC).

This issue documents the gaps in the FFDC support for the Glossary View OMAS based on a quick review of the code.

This is an extract of the GlossaryResource from the spring module.  Because spring is optional, there should be no logic in this module - just a straight delegation to the service in the server module.  We are only using Spring to marshall the REST call.

```
@RequestMapping(method = RequestMethod.GET, path = "/glossaries/{glossaryGUID}/external-glossary-links")
    public GlossaryViewEntityDetailResponse getExternalGlossaryLinks(@PathVariable("serverName") String serverName,
                                                                     @PathVariable("userId") String userId,
                                                                     @PathVariable("glossaryGUID") @NotBlank String glossaryGUID,
                                                                     @RequestParam(name="from", defaultValue=PAGE_FROM_DEFAULT_VALUE) @PositiveOrZero Integer from,
                                                                     @RequestParam(name="size", defaultValue=PAGE_SIZE_DEFAULT_VALUE) @PositiveOrZero @Max(PAGE_SIZE_MAX_VALUE) Integer size) {
        StopWatch watch = StopWatch.createStarted();

        GlossaryViewEntityDetailResponse response = glossaryService.getExternalGlossaryLinks(userId, serverName, glossaryGUID, from, size);

        watch.stop();
        log.debug("Method: getExternalGlossaryLinks; Duration: " + watch.getTime()/1000 + "seconds");

        return response;
    }
```
There are two main problems
 * the debug and performance measurement should be in there server (also logging the start of the service call.
 * the parameter validation is being done with Spring annotations which means that FFDC style exceptions are not thrown.  (Also it is not using the max page size defined for the server)

This next extract is within the server module.  The separation between a parameter error, security or repository error is lost as they are grouped into a single type of exception.  This reduces the ability for the caller to react.

```
protected Optional<EntityDetail> getEntityDetail(String userId, String serverName, String guid, String entityTypeName, String methodName)
            throws GlossaryViewOmasException {

        Optional<EntityDetail> entityDetail ;
        try {
[...]

PR #2621, opened 2020-02-20
issue-1682 Make gv-omas ffdc compliant

lcpopa:
Exposing to requester incoming known checked exceptions from OMRS. The rest are collected into the GlossaryViewOmasException with request data and returned to requester


lcpopa:
#1682

PR #2627, opened 2020-02-21
Fix up "why Egeria" page

mandy-chessell:


open-metadata-publication/website/why-egeria/README.md
```diff
@@ -23,7 +23,7 @@ The name needed to be available for trademark and a member of the
 community came up with "Egeria".
 
 Egeria was a Roman Water Nymph and a wise counselor.
-(See [Wikipedia for more details](https://en.wikipedia.org/wiki/Egeria_(mythology\))).
+(See [Wikipedia for more details](https://en.wikipedia.org/wiki/Egeria_(mythology%29)).
```
cmgrote: I think the issue is that without the `\)` then markdown interprets this first `)` as the closing of the link -- so it's a necessary escape mechanism for markdown (which works for markdown), but the interpreter checking for broken links doesn't properly handle it...

I've added it to a whitelist of links to ignore instead.

(Or if you've tested this change and confirmed it works, it seems like we have an extra ')' mark at the end still?)

mandy-chessell:
Had another go at this - problem is that the obvious mark up does not work in GitHub but does work on the main site.  I have been experimenting with different styles of markup to try and get it to work in both places - however each attempt needs to go through the PR cycle

PR #2632, opened 2020-02-21
Open Lineage Services:  auditlog original exceptions instead of omag config exceptions

GuyAckermans:


GuyAckermans:
The context is passed to the logException method via the auditcode argument. In the logException method the context message is passed to the auditlog together with the exception

GuyAckermans:
/**
     * Write an exception to the audit log.
     *
     * @param auditCode         Reference to the specific audit message.
     * @param actionDescription Describes what the user could do to prevent the error from occurring.
     * @param e                 The exception object that was thrown.
     */
    private void logException(OpenLineageServerAuditCode auditCode, String actionDescription, Throwable e) {
        auditLog.logException(actionDescription,
                auditCode.getLogMessageId(),
                auditCode.getSeverity(),
                auditCode.getFormattedLogMessage(localServerName, openLineageServerConfig.toString()),
                null,
                auditCode.getSystemAction(),
                auditCode.getUserAction(),
                e);
        log.error(auditCode.getSystemAction(), e);
    }

GuyAckermans:
The relevant value for auditCode is set in the catch block of the method where the exception occurred

GuyAckermans:
In short, any logging of an exception will always be accompanied with a message stating whether it occurred during initialising, starting or disconnecting each of the three connectors

GuyAckermans:
Catch blocks now catch Exception instead

Issue #1876, opened 2019-10-28
fix infinite loop while sending message 


Shrinivas-Kane:
Scenario: 
If kakfa throws message is too big or kafka specific exception it goes in infinite loop 
 
We need to fix infinite loop here: 
https://github.com/odpi/egeria/blob/master/open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java#L101

mandy-chessell:
Thanks for spotting this.  It could stop immediately if it gets an exception that is not an `ExecutionException` or `WakeupException`.  Alternatively it could retry a few times with a sleep in the hope that it may clear).

Any preferences?

planetf1:
From https://kafka.apache.org/23/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html the exceptions from Producer.send seem to be:
```
AuthenticationException - seems fairly unlikely that will fix itself (though there could be a transient issue - if so it's more likely to last for a while due to bad config - so retrying quickly probably isn't that helpful)
AuthorizationException - as above
IllegalStateException - coding error?
InterruptException - could be worth a retry or two?
SerializationException -unlikely to help?
TimeoutException - probably worth a retry or two
KafkaException -hard to know.
```

I would be inclined to only try a couple of times then abort. Maybe even only on timeout & interrupt? Ideally I'd consider retrying after a delay - which means more async handling would be needed for the retry - perhaps unnecessarily complex?

planetf1:
& further from
ApiException, BufferExhaustedException, CommitFailedException, ConfigException, ConnectException, InterruptException, InvalidOffsetException, SerializationException, StreamsException, WakeupException
 - as Mandy mentioned, probably wakeup, interrupt worth a short retry - the others less so

mandy-chessell:
Billy is looking at this ...

wbittles:
There may be some confusion around which exceptions we are actually concerned with I've copied and pasted the javadoc below and  I assumed that it was the 3 exceptions thrown from Future.get

V get()
      throws InterruptedException,
             ExecutionException

Waits if necessary for the computation to complete, and then retrieves its result.

Returns:
    the computed result
Throws:
    CancellationException - if the computation was cancelled
    ExecutionException - if the computation threw an exception
[...]

PR #2638, opened 2020-02-24
1876 infinite loop fix 

wbittles:
This fix is for #1876


open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
```diff
@@ -97,58 +96,97 @@ private void publishEvent(String event) throws ConnectorCheckedException
...
+                                                                            error.getMessage()),
+                                           null,
+                                           auditCode.getSystemAction(),
+                                           auditCode.getUserAction());
+                    }
+
+                    eventRetryCount++;
+                }
+            }
+            catch (WakeupException error)
```
planetf1: Though this occurs at shutdown, could it occur in other situations too? 
For example another thread could use wakeup() against this thread - though I think we do?
I'm wondering if we should pause and retry ... though at shutdown I'd be concerned we wait too long
If we really only think wakeup will be at shutdown then this handling is ok-ish, but it's one route out of the method where we don't log an audit message at all, which probably would be desirable?

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
```diff
@@ -206,11 +244,19 @@ public void run()
             catch (Throwable   error)
             {
                 log.error("Bad exception from sending events " + error.getMessage());
-                this.recoverAfterError();
+
+                if( isExceptionRetryable(error) ) {
+                    this.recoverAfterError();
```
planetf1: I noticed in this method we are using time in 2 different units - around line 39. Seconds for recoverytime, milliseconds for regular sleep time. though handled correctly would be better to use consistent units?

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
```diff
@@ -97,58 +96,97 @@ private void publishEvent(String event) throws ConnectorCheckedException
...
+                /*
+                 * This may be a simple timeout or something else more
+                 */
+                log.debug("Kafka had trouble sending event: " + event + "exception message is " + error.getMessage());
+
+                if( !isExceptionRetryable(error)) {
+                    /* kafka thinks this isn't a retryable problem */
+                    /* so let the caller try */
+                    log.error("Exception in sendEvent " + error.toString());
+                    KafkaOpenMetadataTopicConnectorErrorCode errorCode = KafkaOpenMetadataTopicConnectorErrorCode.ERROR_SENDING_EVENT;
```
planetf1: I notice the text for this event says the message isn't sent. True in terms of this connector's responsibility, but if queueing occurs further up the stack the message is confusing to the audit log reviewer

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
wbittles: @planetf1 This code is actually original code that was removed as part of this first attempt to fix this issue. I put the behaviour back after getting feedback on the first fix. There is nothing the producer is doing that would cause kafka to throw this.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
wbittles: @planetf1 again this code was there , but I'm happy to change it if Mandy agrees.

wbittles:
@planetf1 So far I've just had a sniff test by running through the conformance suite lab and the egeria server config lab and killed and then restarted  the kafka container during the run, after I was sure everything was up and connected.. In both cases egeria recovered and was responsive to further commands.

planetf1:
I had a think about error conditions. I think we should walk through the design considering the following -- and also look at what testing we can do for each of these:
```

* connection hang (timeout) connecting to broker (producer & consumer) 
* Connection denied connecting to broker (both) 
* Message too big (send) - as previously reported
* topic doesn't exist (ie if auto create disabled which is the case in a production environment) (send)
 * duplicates (can happen with producers invoking retries) (receive)
 * connection closed (both)- ie during the middle of an operation
 * badly formatted message

There's also some permanent errors 
* incompatible versions
* ??

AndMore advanced tests that I think are out of immediate scope, but to look at when we do more testign around full HA & high scalability scenarios - where kafka is of course just one part

 * failed node/partition failover (of leader partition)
 * leader isolation from rest of cluster 
 * buffer full - leading to blocking
 * HA - multiple platforms. kill one (more an application test)
 * slow processing (CPU or network latency) 
 * offset out of range (ie all messages expired before platform restarted)

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
planetf1: Good to stick with this for our immediate kafka fix - can consider revisiting with a new connector implementation

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
planetf1: Good to stick with this for our immediate kafka fix - can consider revisiting with a new connector implementation

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
planetf1: Good to stick with this for our immediate kafka fix - can consider revisiting with a new connector implementation

Issue #2696, opened 2020-03-03
VDC Helm chart - update connector versioning logic


planetf1:
We have a VDC helm chart that deploys egeria, atlas, ranger, gaian etc.
Whilst we intend to focus on this more in the 1.8 timeframe according to the current release plan, there are people using it currently, so we continue to ensure it keeps working as-is.

When a release of egeria is made all references to a release version like '1.5-SNAPSHOT' are changed to '1.5'. This is a full text change so covers poms, but also readmes & source files.

In the helm chart definition for vdc currently we have a few areas to consider
```
egeria:
  version: 1.5
```
This is the version to be used for the docker containers (egeria, configure, jupyter) - and is updated correctly.

There is an issue open to create this version of container during the release cycle automatically
 ( #2583 )

```
connectors:
  repo: snapshot
  version: 1.5
```
This requires consideration. The global change has updated the version tag to 1.5, but this will not immediately be available -- until @cmgrote has had a change to test

Additionally the repo once releases should be 'release'

The easiest option IMO would be to use some scripting in our templates to automatically determine whether to use 'snapshot' or 'release' as the *jfrog* repo - this could be based on whether the version includes '-SNAPSHOT' (and dd a note to the config file/readme to warn the connector may not be immediately available ). This is a small change, and probably the simplest, albeit at the exposure of a few days where the VDC chart doesn't work

There are other options that might include changing in which project the helm charts are placed, their build process, what containers we use, repositories for artifacts, but this is probably overkill for now especially as the vdc chart is going to be revisited in a few releases
```
# Default values used for images
# 'tag' is only used for images other than connectors/egeria where the version above is used
imageDefaults:
  tag: latest
```
This remains appropriate - for atlas, ranger etc

My *short term* proposal for 1.5 is to change repo from 'snapshot' to 'release' so that it works 'soon' (I've published the 1.5 docker images early - chicken & egg challenge). This needs to be done imminently (I will do this)

It would be good to determine the right way moving forward for 1.6/7

cc: @cmgrote

cmgrote:
Part of the reason I've left the `repo` section separate rather than auto-determining it is that during final testing it is useful to use Artifactory's staged artifacts (prior to their release), in which case the `repo` is `staging` while the release identifier is just `1.x` (ie. no `-SNAPSHOT`, but also not in the `release` repo, yet...)

[...]

PR #2697, opened 2020-03-03
Release 1.5 - update repo for connectors

planetf1:
Updates the reference to the Artifactory repository to use for the 1.5 connector from 'snapshot' to 'release'

See #2696 for a discussion.

This is a quick tactical fix for 1.5, pending a better fix for 1.6
It will work once the connector is released 

@cmgrote - if you think this works, feel free to approve, otherwise we can abandon and discuss further in the issue


cmgrote:
I'm hesitant given that it seems it can take us a matter of weeks after we've branched before the core release is available, meaning that this won't be a matter of hours (or even a day or two), but weeks during which the chart will be broken (even if I manage to release the connectors on the very same day that core hits Maven Central).

I can't even build the connectors to stage a non-snapshot release (ie. Azure pipelines are all going to fail) until Egeria core's modules are available in MC...

planetf1:
Whilst the version is always overrideable by the end user, and the audience for these charts are very low, it would be good to figure out some approach that works

I'll close this PR and let's revisit in the issue.

PR #2707, opened 2020-03-04
Include Glossary Terms in lineage result

lcpopa:


GuyAckermans:
Has it been confirmed that we want to display glossary terms as separate nodes in the UI, rather than including them as properties for the columns and tables? This may greatly clutter the screen of the user when viewing end to end lineage

lcpopa:
@GuyAckermans  I do not have that information. Who can provide it ?

planetf1:
Where is the issue that documents this problem or proposal? 

We should be able to track the work we are doing at a broader level than just in PRs. Indeed one issue may require a number of PRs (and even more commits) to complete.

Current process is documented in https://github.com/odpi/egeria/blob/master/developer-resources/Issue-Tracking.md

lcpopa:
#2716

lcpopa:
> On hold awaiting confirmation whether glossary terms should be properties or separate nodes

Confirmation received.

lcpopa:
> On hold awaiting confirmation whether glossary terms should be properties or separate nodes
This was settled. Option to display glossary terms as either nodes or properties will be implemented in a separate issue. At the moment, they will be displayed as nodes

PR #2708, opened 2020-03-04
UI add about-view

bogdan-sava:
This view provides information about the current version and build of the UI


planetf1:
We should have appropriate issues to track the proposed work in the UI so as to contain the discussion & feedback before getting to submission of code changes in the PR

More detail in the current process flow at https://github.com/odpi/egeria/blob/master/developer-resources/Issue-Tracking.md (which of course we can review & discuss)

bogdan-sava:
@planetf1 you are right.
However, link for about-view is already there and I mentioned it in the past.
I have create also an issue: #2711

bogdan-sava:
Please also note that is not affected current functionality, it only add a new view with information about version of the UI running and build time.

PR #2715, opened 2020-03-06
Migrates Data Engine Proxy to new audit log framework

cmgrote:


cmgrote:
@mandy-chessell ^^ sounded like from today's call you might be doing this work as well -- just a heads-up that it should be done already for the Data Engine Proxy area per this PR...

PR #2718, opened 2020-03-08
Asset Lineage OMAS: Reclassify, Declassify

GuyAckermans:


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
```diff
@@ -167,39 +165,62 @@ private void processDeletedEntity(EntityDetail entityDetail) throws ConnectorChe
...
 
+
     private void processNewRelationship(Relationship relationship) {
-        log.debug(PROCESSING_RELATIONSHIP_DEBUG_MESSAGE, "NewRelationship", relationship.getGUID());
+        log.debug(PROCESSING_RELATIONSHIP_DEBUG_MESSAGE, "newRelationship", relationship.getGUID());
     }
 
     private void processUpdatedRelationshipEvent(Relationship relationship) throws OCFCheckedExceptionBase, JsonProcessingException {
-        log.debug(PROCESSING_RELATIONSHIP_DEBUG_MESSAGE, AssetLineageEventType.UPDATE_RELATIONSHIP_EVENT, relationship.getGUID());
+        log.debug(PROCESSING_RELATIONSHIP_DEBUG_MESSAGE, "updatedRelationship", relationship.getGUID());
```
popa-raluca: Just a question here, why don't you want to use directly the event type(e..g AssetLineageEventType.UPDATE_RELATIONSHIP_EVENT) for debugging? And use another string instead?

GuyAckermans:
recheck

GuyAckermans:
@planetf1 Checks seem to be frozen

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
GuyAckermans: AssetLineageEventType does not refer to the input events of the omas, but to its output events. Some input events, such as classify and reclassify, may lead to an event of the same AssetLineageEventType (CLASSIFICATION_CONTEXT). Therefore these constants are not sutied for classifying the input events

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/admin/AssetLineageAdmin.java
```diff
@@ -119,6 +126,18 @@ public void initialize(AccessServiceConfig accessServiceConfig,
         }
     }
 
+    private List<String> getLineageClassificationTypes(AccessServiceConfig accessServiceConfig) {
```
lcpopa: My version of this method (not actually tested, but the result should be identical)
`private List<String> getLineageClassificationTypes(AccessServiceConfig accessServiceConfig) {
        if (accessServiceConfig.getAccessServiceOptions() != null) {
            return AssetLineageConstants.immutableDefaultLineageClassifications;
        }
        return accessServiceConfig.getAccessServiceOptions().get(AssetLineageConstants.LINEAGE_CLASSIFICATION_TYPES_KEY);
    }`

Steps taken to reach this code:
- noticed else block is smaller than if block, therefore switched between themselves
- used a fast return in if block, making 1 less usage of returned list and making else block unnecessary, thus flattening landscape (aka decreasing identation) and increasing code readability
- since the returned list/object is already initialized as null, the null check becomes unnecessary and therefore can be removed
- as a consequence, the assignment to returned list/object is also unnecessary and the result of get() method can be immediately returned

I am only stating this is an improved version, not the best possible one, only by staying in the context of this method alone.
Also, my approach is to always use curly brackets/accolades, however they are called, when it comes to single line if statements. The web debates about this case, but in my opinion they provide code structure, thus increasing readability. Also, they help with spotting failed code merges, but if they are not used the bugs rising from these situations can turn into nightmares. The choice is ultimately the developers.

lcpopa: How do I make GitHub properly indent code in comments ? :)

GuyAckermans: Thanks for the feedback! It made me realise there was a bug, the body of the else block should be run when either of the if statements are false. I refactored to:

    ```
private List<String> getLineageClassificationTypes(AccessServiceConfig accessServiceConfig) {
        Object lineageClassificationTypesProperty = null;
        if (accessServiceConfig.getAccessServiceOptions() != null)
            lineageClassificationTypesProperty = accessServiceConfig.getAccessServiceOptions().get(AssetLineageConstants.LINEAGE_CLASSIFICATION_TYPES_KEY);
        if (lineageClassificationTypesProperty != null)
            return (List<String>) lineageClassificationTypesProperty;
        return AssetLineageConstants.immutableDefaultLineageClassifications;
    }


```

To me code is more readable when minimising the number of lines, so using single line statements. I have never run into a failed code merge situation with this, but I can imagine it would be a big problem

Issue #2717, opened 2020-03-06
Kafka topic connector - restarts on startup


planetf1:
When running the coco pharma notebook (configure, start) I see 126 shutdown events from topic connectors ie:

```
core_1       | Fri Mar 06 12:44:36 GMT 2020 cocoMDS2 Shutdown OCF-KAFKA-TOPIC-CONNECTOR-0011 The Apache Kafka producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic is shutting down after sending 1 messages and with 69 unsent messages
```

The pattern is:
```
core_1       | Fri Mar 06 12:44:41 GMT 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC-CONNECTOR-0002 8 properties passed to the Apache Kafka Producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic
core_1       | Fri Mar 06 12:44:41 GMT 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC-CONNECTOR-0010 The Apache Kafka producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic is starting up with 0 buffered messages
core_1       | Fri Mar 06 12:44:42 GMT 2020 cocoMDS2 Shutdown OCF-KAFKA-TOPIC-CONNECTOR-0011 The Apache Kafka producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic is shutting down after sending 1 messages and with 25 unsent messages
core_1       | Fri Mar 06 12:44:42 GMT 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC-CONNECTOR-0002 8 properties passed to the Apache Kafka Producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic
core_1       | Fri Mar 06 12:44:42 GMT 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC-CONNECTOR-0010 The Apache Kafka producer for topic egeria.omag.openmetadata.repositoryservices.cohort.cocoCohort.OMRSTopic is starting up with 0 buffered messages
```
repeated

The full output is at https://gist.github.com/a81f41773b688b63546571556d2102e4

In this case kafka is working, and the topics do exist - so something else is destabilizing the environment during startup. There is no manipulation to introduce kafka errors.

[...]

PR #2725, opened 2020-03-10
It infrastructure

wbittles:
The change moves the control of the kafka producer lifecycle to inside a single method, now Producer are created when needed and closed on Error. I've also added more debug so we can track the lifecycle of the producer more closely.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -184,7 +184,7 @@
         <maven-compiler.version>3.8.1</maven-compiler.version>
         <maven-jar-plugin.version>3.2.0</maven-jar-plugin.version>
         <maven-pmd.version>3.13.0</maven-pmd.version>
-        <owasp.version>5.3.0</owasp.version>
+        <owasp.version>5.3.1</owasp.version>
         <spotbugs-maven.version>3.1.12.2</spotbugs-maven.version>
         <spotbugs.version>4.0.0</spotbugs.version>
         <maven-download.version>1.5.0</maven-download.version>


open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java.orig
```diff
@@ -0,0 +1,548 @@
+/* SPDX-License-Identifier: Apache 2.0 */
```
cmgrote: Do we need this file?  I would have expected that we would let Git manage the change history for us, rather than having `.orig` files scattered around (?)

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
```diff
@@ -99,6 +100,10 @@ private void publishEvent(String event) throws ConnectorCheckedException
         boolean                  eventSent = false;
         long                     eventRetryCount = 0;
 
+        if( producer == null){
```
cmgrote: Formatting isn't following the convention already present in the code:

```java
if (producer == null)
{
   // ...
}
```

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
```diff
@@ -119,7 +124,11 @@ private void publishEvent(String event) throws ConnectorCheckedException
                 if( !isExceptionRetryable(error)) {
                     /* kafka thinks this isn't a retryable problem */
                     /* so let the caller try */
-                    log.error("Exception in sendEvent " + error.toString());
+
+                    producer.close();
+                    producer = null;
+
+                    log.error("Execution Exception closed producer ");
```
cmgrote: I don't think we need to log an error here given that we're throwing an exception in the lines immediately following?

wbittles:
@cmgrote Thanks for the review , I've made the code changes you suggested and my thoughts on your questions are below.

1) The retry loop count ideally should be a property, but a lot of tests had been run with this already hard coded, so when I changed it originally it caused a lot of confusion , so I put it back in. I'm more than happy to address it in a different issue , but for this code I'm trying to keep behavioural changes to a minimum.

2) Again I believe this relates to a kubernetes environment where kafka was being temporarily starved of resources and this approach addressed that specific problem, I'm unsure as to the impact if we are asleep when kafka does start processing  In addition any delay in the thread  will cause blocking on any thing trying to send an event and finally I'm unsure what impact that would have when trying to shutdown egeria while in that loop.  
Currently when we reach the retry limit we give up and perform an effective delayed reconnect process by shutting down and restarting anyway

So I'm far more nervous of this change especially because of any unforeseen consequences.

However , if you do feel it would be useful again I'm happy to address it in another issue.

planetf1:
> @cmgrote Thanks for the review , I've made the code changes you suggested and my thoughts on your questions are below.
++ Agreeing with Chris's comments

> 1. The retry loop count ideally should be a property, but a lot of tests had been run with this already hard coded, so when I changed it originally it caused a lot of confusion , so I put it back in. I'm more than happy to address it in a different issue , but for this code I'm trying to keep behavioural changes to a minimum.
Agree with you both. I think we need it but can be a seperate PR - i think it should probably continue indefinately by default (though not without a delay - see below)

> 2. Again I believe this relates to a kubernetes environment where kafka was being temporarily starved of resources and this approach addressed that specific problem, I'm unsure as to the 

I don't think kafka was starved as such. It was seen in k8s initially though I suspect that was more related to overall performance . Also seen, but less elsewhere (and not at all when gathering debug.. as that in itself slowed things down)


> impact if we are asleep when kafka does start processing  In addition any delay in the thread  will cause blocking on any thing trying to send an event and finally I'm unsure what impact that would have when trying to shutdown egeria while in that loop.
>    Currently when we reach the retry limit we give up and perform an effective delayed reconnect process by shutting down and restarting anyway
> 
> So I'm far more nervous of this change especially because of any unforeseen consequences.

If we don't feel we understand it well enough, more experimentation may be desirable before we merge to master. But I understand your reasoning -- and given we have the slower outside loop, agree with the proposal to leave as-is initially.

> However , if you do feel it would be useful again I'm happy to address it in another issue.
We can continue discussing in the issue? But it could be a new PR? Or if we really believe it's independent and the main issue is fixed just with this change, then ok to close and open new one

Generally also
 * We should cross reference the issue related to this change #2717 (there, done..!)
 * Update the title of PR to be slightly more meaningful (I initially thought this was new omas)

I'll give the PR a go.

cmgrote:
> 1. The retry loop count ideally should be a property, but a lot of tests had been run with this already hard coded, so when I changed it originally it caused a lot of confusion , so I put it back in. I'm more than happy to address it in a different issue , but for this code I'm trying to keep behavioural changes to a minimum.

Makes sense -- happy for it to live as a separate issue and eventual PR.

> 2. Again I believe this relates to a kubernetes environment where kafka was being temporarily starved of resources and this approach addressed that specific problem, I'm unsure as to the impact if we are asleep when kafka does start processing  In addition any delay in the thread  will cause blocking on any thing trying to send an event and finally I'm unsure what impact that would have when trying to shutdown egeria while in that loop.
>    Currently when we reach the retry limit we give up and perform an effective delayed reconnect process by shutting down and restarting anyway
> 
> So I'm far more nervous of this change especially because of any unforeseen consequences.
> 
> However , if you do feel it would be useful again I'm happy to address it in another issue.

Fair enough -- fine by me to leave for now in this PR, was just wondering if this could have caused any resource starvation because we were hammering something that was already incapable of handling what we were throwing at it to begin with.  (I'll let you and @planetf1 decide whether that's worthy of an issue at this stage for tracking purposes or not?)

planetf1:
I don't believe I can quanitfy a specific scenario based on what I've seen recently that could lead to a new issue.

The points around overloading, configuration etc are things we need to look at - there's a lot of controls around # requests in flight, buffer sizes, use of acks, message history size, working with clusters, introducing network disturbances  but I'd attach those to the new generic issue we have to discuss a new connector?

I tried this change in my environment and it's looking good so far. I too don't see the disruption at startup any more. 

I'd be inclined to go with the merge - but until we have a lot more automated and disruptive testing, we do incur some risk if our analysis is incomplete or wrong.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataEventProducer.java
cmgrote: If I can be super-picky (sorry!) I'd suggest the conditionals should be:

```java
if (producer == null)
{
   // ...
}
```

At the moment all the `if ...` statements look more like function calls with the `(` coming immediately after `if` without any space in-between, and instead a space between the `(` and the variable used for comparison -- just looks messy...

wbittles:
@cmgrote absolutely no problem with being picky , I've made the changes you requested

cmgrote:
Seems to be a step in the right direction compared to existing behaviour, so I will merge.

Issue #2795, opened 2020-03-25
Egeria compilation failure if JAVA_HOME is missing in the env


fpompermaier:
if the JAVA_HOME env variable is not set the compilation of ALF fails because of the javadoc maven plugin. The error given is the following:

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:jar (attach-javadocs) on project audit-log-framework: MavenReportException: Error while generating Javadoc: Unable to find javadoc command: The environment variable JAVA_HOME is not correctly set. `

As stated by https://stackoverflow.com/questions/13961615/unable-to-find-javadoc-command-maven it is sufficient to add the following configuration to the plugin:

```
<plugin>
        <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
         <version>3.0.1</version>
          <configuration>
               <javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>
          </configuration>
</plugin>
```

fpompermaier:
This does not seem to work on JDK < 1.9 because javadoc is not under jre/bin (see https://issues.apache.org/jira/browse/MJAVADOC-595)

planetf1:
We came across this in #1500. Unfortunately I did try this fix, and we found one issue was that it sets the path to javadoc in relation to the java install executing maven, which in some cases isn't a JDK - but could be a JRE. And in that case there is no javadoc executable.

That being said it's good to revisit - so will suggest looking at this again - should probably hold off any merge for the moment

Ah! I just saw your post - exactly what I was going to say!

planetf1:
This was the PR that updated the dev docs 2316 but clearly if we can find an approach within maven that can cover java 8,11 & beyond, use of jre or jdk, and works cross-platform, which still ensuring javadoc is always built (for validation) then that's great :-)

Changing the build env and removing from the normal mvn clean install is tempting, but means developers won't see javadoc warnings, which tends to mean issues - even warnings - don't ever get addressed....

planetf1:
May be possible to add a plugin that could check JAVA_HOME is set, and if not produce a clearer, quicker message ..?

fpompermaier:
I think the only way to deal with this here is to use maven profiles with activation rules that tries to guess where the javadoc executable is. If the user is using a jre to compile we can disactivate the javadoc plugin

planetf1:
A few thoughts
[...]

PR #2796, opened 2020-03-25
Fixed javadoc-plugin failure on missing JAVA_HOME env variable

fpompermaier:
Closes issue #2795

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2541,6 +2541,7 @@
                 <configuration>
                     <additionalJOption>-J-Xmx1536m</additionalJOption>
                     <quiet>true</quiet>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -2541,6 +2541,7 @@
                 <configuration>
                     <additionalJOption>-J-Xmx1536m</additionalJOption>
                     <quiet>true</quiet>
+                    <javadocExecutable>${java.home}/bin/javadoc</javadocExecutable>
```
planetf1: See also #1500 - issues if run under JRE

planetf1:
Suggest closing the PR for now until/if we address the discussion in #2795 
(can always be reopened)

Issue #2669, opened 2020-02-27
netty issues


planetf1:
netty components used via dependencies in Egeria are at a variety of levels:
 - 4.1.25 - from graph-repository-connector, open-connector-framework)
 - 4.1.45 - from cassandra-data-store-connector, repository-services-api
 - 4.0.44 netty-all - (including via io.netty:netty:jar:3.7.0.Final)

The third party artifacts we use that cause the dependency to be pulled in are:
org.apache.cassandra:cassandra-all:jar:2.2.13 (this is where we get 4.0.44 - very old)
org.janusgraph:janusgraph-cql:jar:0.4.1 (pulls in com.datastax.cassandra:cassandra-driver-core:jar:3.7.2) - this causes 4.1.25 versions to be pulled in 
com.datastax.oss:java-driver-core:jar - this is ok as it's pulling in 4.1.45 (current)

netty-all is showing up in security reports as being subject to CVE-2019-20444, CVE-2019-20445, CVE-2019-16869, CVE-2020-7238. 

Latest versions generally of most of the components are 
 4.1.45 (Jan 2020)

io.netty:netty is used in one location by
[INFO] org.odpi.egeria:gaian-ranger-plugin:jar:1.5-SNAPSHOT
[INFO] |  |  \- io.netty:netty:jar:3.7.0.Final:compile
This is deprecated and replaced by netty-all
Will address as part of other ranger issues

To fix we have a few options:

* We can force transitive dependencies up to 4.1.45 for the affected artifacts. Given netty is pretty stable and mostly security fixes there's a good change this will be good - but we are affecting janus and cassandra -- so may cause testing concerns

* The ranger dependency can be handed by moving that code outside the core repo

* All instances of netty are pulled in through connectors  - these need not be expressed as maven dependencies, but loaded at runtime. this will reduce the potential scope of any exposure by isolating where the module is used/present, though doesn't fix the underlaying issue.

planetf1:
Added additional entries into dependency management for the other netty subcomponents - whilst janus uses netty-all, cassandra uses individual components. This brings them all in sync & current. Minor fixes so hopefully ok.

I have not changed the old version io.netty:netty -- this is used only within the ranger plugin which pulls in hadoop & within that zookeeper -- we have MANY issues in this area of the code so there seemed little point in just addressing netty. Instead this will be addressed in part by #958

grahamwallis:
Thanks @planetf1 - sounds like a useful synchronisation.

Issue #2828, opened 2020-03-31
Update JanusGraph to 0.5.0


grahamwallis:
Requires updates to graph repo and OLS POMs, as well as root POM for dependency management of version of netty.

PR #2829, opened 2020-03-31
Updates JanusGraph to 0.5.0 plus dependencies

grahamwallis:
This PR has the POM changes for the above change, for graph repo and OLS.
It updates the version of netty.

It has been built and given a sniff test in which the new version was brought up with an existing JG 0.4.0 graph repo database. The test ran OK.

I have not run a full conformance test on this yet.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,7 +141,7 @@
         <netty.version>4.1.42.Final</netty.version>
         <spring-boot.version>2.2.6.RELEASE</spring-boot.version>
         <spring.version>5.2.5.RELEASE</spring.version>
-        <spring-security.version>5.3.0.RELEASE</spring-security.version>
+        <spring-security.version>5.3.1.RELEASE</spring-security.version>
         <spring-data.version>2.2.6.RELEASE</spring-data.version>
         <tomcat.version>9.0.33</tomcat.version>
         <commons-io.version>2.6</commons-io.version>


grahamwallis:
Fix for issue #2828

pom.xml
```diff
@@ -136,8 +136,9 @@
         <jackson.version>2.10.3</jackson.version>
         <jackson.databind.version>2.10.0</jackson.databind.version>
         <kafka.version>2.4.1</kafka.version>
-        <janus.version>0.4.1</janus.version>
+        <janus.version>0.5.0</janus.version>
         <gremlin.version>3.4.6</gremlin.version>
+        <netty.version>4.1.42.Final</netty.version>
```
planetf1: I note netty latest is 45.1.48-Final . I expect dependabot will point this out later. Is there a reason for not going there?

Hopefully we'll fix #2669 but I will pull & check

pom.xml
```diff
@@ -158,7 +159,7 @@
         <ranger.version>1.2.0</ranger.version>
         <plexus.version>3.3.0</plexus.version>
         <quartz.version>2.3.2</quartz.version>
-        <lucene-analyzers-common.version>7.0.0</lucene-analyzers-common.version>
+        <lucene-analyzers-common.version>7.7.2</lucene-analyzers-common.version>
```
planetf1: See #1106 - this doesn't take us to 7.8 but hopefully will address some of the issues

I'll suggest running an owasp check (-DCVE on the build) - I can pull and check - though I think your proposed change is first step in any case

planetf1:
Just adding tag whilst testing

pom.xml
grahamwallis: I selected 4.1.42.Final only because it seems to be what JG, Cassandra et al have been tested with. We could push to the latest netty update (4.1.48.Final) if you prefer.

planetf1:
Just taking a look through.... will add any observations

Note that netty still has a mix of versions - but that was the case before and we can leave #2669 open to persue. Janus core isn't affected, though cassandra is (sql), & hadoop

planetf1:
janusgraph-driver looks like a valid compile & runtime dependency - so think it's fine as it is. Also noted a few other dependencies (including a few I added for swagger) aren't specifying the scope, though again compile is the default. I should fix....

I've not tested the function - but I think we're good to merge this.
I'll look at doing the netty change shortly also. I can't test lineage without docs but will leave change open for review.

pom.xml
planetf1: Either... ok as is, but I will plan to look at #2669 straight away and pull up there?

pom.xml
planetf1: owasp checks are failing. Will review in nexus-iq reports after merge

planetf1:
Fixed merge with spring-boot.

grahamwallis:
Thanks @planetf1

Issue #1124, opened 2019-06-13
Improve swagger docs


planetf1:
The Swagger docs for Egeria could do with some improvement
 
* Better descriptions for operations, parameters , responses, body
* Removing of framework APIs (like basic-error-controller)
* Adding egeria specific title, link to github etc
* Only documenting REST API calls served by any particular server chassis (if possible?)

Original issue: Swagger was completely broken, see https://github.com/odpi/egeria/issues/1121

It may be possible to incorporate javadoc into the swagger docs - needs further investigation, perhaps using the springfox-javadoc plugin. However this needs base javadoc build tobe fixed, and needs to correctly pull in the right javadoc into the springfox doclet. See https://github.com/springfox/springfox-javadoc (which won't work as-is in our multi module project). https://github.com/odpi/egeria/issues/1110 will fix base javadoc as a prereq

planetf1:
It may be worth looking at redoc https://github.com/Redocly/redoc which offers a nice presentation - I first came across this with the formal docker documentation at https://docs.docker.com/engine/api/v1.25/

Note that swagger is somewhat behind the curve in spring versions too which has been causing an issue when trying to uplift our dependencies.

planetf1:
It's worth reading https://github.com/springfox/springfox/issues/3070 .

Springfox is a neat tool, but requires a community to allow it to move forward/keep up. That probably means either being happy to jump in to help, or evaluating alternative approaches with an existing community.

planetf1:
Incorporate comments from https://github.com/odpi/egeria/issues/213 -- to ensure that the generated docs adequately address how to appropriate compose the request

planetf1:
another spring doc project. see : https://github.com/springfox/springfox/issues/2932#issuecomment-559446881 which refers to https://github.com/springdoc/springdoc-openapi (also a small project)

Docker (ie https://docs.docker.com/engine/api/v1.40/#operation/Session) meanwhile uses 'redoc' - see https://github.com/Redocly/redoc which seems richer and has a more active community

planetf1:
No springfox updates since 2018 and is subject to CVE-2019-17495

The docs themselves are rudimentary and offer no clarity over which APIs are in development, in tech preview, or released. 

I'm thinking we should remove springdoc as a first step and evaluate the alternatives
There's a few suggestions at https://stackoverflow.com/questions/36634281/list-of-swagger-ui-alternatives

planetf1:
[...]

PR #2835, opened 2020-04-01
#1124 Add tags to split up swagger docs into sections

planetf1:

As part of #1124 this change improves the swagger docs for the REST services provided by the server chassis.
 * REST endpoints are now split by module ie 'Asset Owner' or 'DevOps' or 'OMRS'
 * Links to the documentation are included
 * defaults to a collapsed layout 
 * deprecated operators are striked through
* Modules & operations are alpha sorted

Caveats
 * Formatting isn't great - each Module's header is independently formatted - so they do not align. 
 * Don't expect too much - this is a small step forward, and it's far from great...
 * No change to the docs for specific operations

I will shortly setup on a server and add a link below for comments

Questions for reviewers
 * I've named the OMASs 'Asset Owner', 'Asset Consumer' etc but not included OMAS in the first label. We could add this, or we could use 'OMAS : Asset Owner' which might help with ordering given that I can only order by alpha (without writing a new sort method)
 * APIs can be hidden if we deem them irrelevant - please identify any candidates (for example I've included view services for now)
 * I've taken the first sentance from docs as a description of the module
 * Are we ok including all our APIs in a single page -- or do we want a harder split - ie access services via OMRS. 

In future PRs I would like to look at:
 * Layout
 * Duplicate tag definitions (expedient for first draft)

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/type-explorer/type-manager.js
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/type-explorer/type-manager.js
@@ -66,7 +66,6 @@ class TypeManager extends PolymerElement {
      */
     loadTypes(serverName, serverURLRoot, enterpriseQuery) {
 
-        alert("type-manager cpt: serverName and serverURLRoot fields are "+serverName+" "+serverURLRoot);
         if (this.validate(serverName) && this.validate(serverURLRoot)) {
 
             if (serverName.length > 0 && serverURLRoot.length > 0) {


mandy-chessell:
We will need the OMAS in the name so distinguish between related access, governance and view services.  For example we have the Data Engine OMAS and the Data Engine Services.  

I would use their names as appears in the component descriptions, documentation etc so that related services are grouped.  So we may have Discovery Engine OMAS, Discovery Engine OMVS, Discovery Engine Services, ... all grouped together.

planetf1:
Test server : http://158.176.66.122:54321/swagger-ui.html

planetf1:
Names updated in PR & on test server

PR #2864, opened 2020-04-03
Adds data engine proxy polling optimizations

cmgrote:
Adds a new data engine proxy configuration option to optimize how many changes the data engine proxy attempts to do during any one polling interval.

This should allow us to avoid a scenario (ie. during initial load) where too many changes are attempted to be processed and we are unable to checkpoint changes that have been synchronized due to running out of resources (memory, etc).


cmgrote:
@DimitriosMaimaris as the feature lead, I'm going to let you be responsible for merging when ready.

Issue #1087, opened 2019-05-30
Monitor deprecation of RestTemplate


cmgrote:
Monitor the eventual deprecation of `RestTemplate` from the Spring framework (see description in https://docs.spring.io/spring/docs/current/javadoc-api/index.html?org/springframework/web/client/RestTemplate.html) -- no specific release or timeline seems to be defined (yet), but the intent is there...

While we're keen to retain synchronous behaviour in the main clients to ease integration, it may be worth considering alternative clients that would also expose asynchronous capabilities (eg. to integrate with high-concurrency UIs or the like).

mandy-chessell:
The Egeria code has a client connector that includes RestTemplate() - this is a place where the replacement calls and extended APIs can be added

cmgrote:
Haven't seen any urgency that this is truly marked as deprecated, so suggest we push out beyond 1.4.

grahamwallis:
The RestTemplate doesn't appear to have been deprecated and the comment in the javadoc that @cmgrote linked to talks about 'maintenance mode' rather than deprecation (with a view to future removal). I would suggest we let this close and (re-)open it if and when there we need to move from RestTemplate.

planetf1:
Given that all of our use should be captured in the spring rest connector (other than a few anomalies found in #3056 ) a change is likely fairly simple.

There's a noticeable difference though in the design (sync vs async at the detail level) and different dependencies.

Lots of interesting articles such as https://dzone.com/articles/resttemplate-vs-webclient

I wonder if it might be better to tackle this up-front in any case since we know it will go away

PR #2884, opened 2020-04-07
Changes URI encoding for RestTemplate for SpringRESTClientConnector

### LINKED ISSUE #2883: Proposed change to RestTemplate for SpringRESTClientConnector
This issue is a proposed change to the SpringRESTClientConnector.

It currently news up a RestTemplate.
The proposed change is to use create the RestTemplate with the RestTemplateBuilder so that the URI encoding can be controlled.

The reason for this is that, if a regular expression is used for searchCriteria on a call to findEntitiesByPropertyValue or findRelationshipsByPropertyValue, the searchCritera string is appended to the URI as a query parameter. 

If the regex contains a '+' character (which means the string to be matched must contain at least one, and optionally multiple,  occurrences of the previous character - the '+' is encoded as a space.

For example, the regex "bond_0+7" should match "bond_07", "bond_007", "bond_0007", etc.
Without the above proposal the query parameter is converted to "bond_0 7", so the search fails.
With the proposed change, the "+" is preserved, and the regex is delivered to the Spring server interface as "bond_0+7", and the search would succeed.

The change is to replace the creation of the RestTemplate with the following:
```
 DefaultUriBuilderFactory builderFactory = new DefaultUriBuilderFactory();
 builderFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.VALUES_ONLY);
 restTemplate = new RestTemplateBuilder()
                .uriTemplateHandler(builderFactory)
                .build();
```
The associated PR has been built and tested and addresses the problem. 
With the PR, searches work.

grahamwallis:
Fixes #2883

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/graphrepository/repositoryconnector/GraphOMRSClassificationMapper.java
+++ b/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/graphrepository/repositoryconnector/GraphOMRSClassificationMapper.java
@@ -124,7 +124,8 @@ public class GraphOMRSClassificationMapper {
             TypeDef typeDef = repositoryHelper.getTypeDefByName(repositoryName, typeName);
 
             // For the type of the entity, walk its type hierarchy and construct a map of short prop name -> qualified prop name.
-            Map<String, String> qualifiedPropertyNames = GraphOMRSMapperUtils.getQualifiedPropertyNamesForTypeDef(typeDef, repositoryName, repositoryHelper);
+            GraphOMRSMapperUtils mapperUtils = new GraphOMRSMapperUtils();
+            Map<String, String> qualifiedPropertyNames = mapperUtils.getQualifiedPropertyNamesForTypeDef(typeDef, repositoryName, repositoryHelper);
 
             // This is a full property update - any defined properties that are not in the instanceProperties are cleared.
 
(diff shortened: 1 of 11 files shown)


open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/pom.xml
```diff
@@ -52,6 +52,11 @@
             <groupId>org.codehaus.plexus</groupId>
             <artifactId>plexus-utils</artifactId>
         </dependency>
+
+        <dependency>
```
planetf1, line 56: This dependency is pulling in many things - 100s of classes, jetty etc. Since this is client code this seems unnecessary. Is there a more precise dependency that could be used for any needed classes? https://mvnrepository.com/artifact/org.springframework.boot

May be worth doing a maven dependency:tree before and after to see transitive dependencies also, as that would be on top of the classes added here?

immmus: This dependence is already present there (https://mvnrepository.com/artifact/org.springframework/spring-web/5.2.5.RELEASE) and, in theory, carries all the necessary functionality for this change. So this addiction is not necessary.

immmus: May be use:
`      DefaultUriBuilderFactory builderFactory = new DefaultUriBuilderFactory();
        builderFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.VALUES_ONLY);
        RestTemplate restTemplate = new RestTemplate();
        restTemplate.setUriTemplateHandler(builderFactory);`

Then there will be no need for spring-boot

grahamwallis: Many thanks @planetf1 and @immus.

I've made the above change and removed the dependency on spring-boot. That is much nicer :-)

planetf1: We are of course using restTemplate which is deprecated - https://docs.spring.io/spring-framework/docs/current/javadoc-api/index.html?org/springframework/web/client/RestTemplate.html though there is an issue open to track at https://github.com/odpi/egeria/issues/1087

So that IMO is totally out of scope here, but just a reminder.

The new solution makes me happy :-)

Issue #2857, opened 2020-04-03
Add new find methods to enable more complex search criteria


cmgrote:
In particular, to allow searching using operators (`=`, `>`, `<`, `!=`, `between`, `in`, etc).

(Thought there was already an issue for this, but couldn't find it @grahamwallis -- do you know?)

From a verbal discussion, leaning towards implementing this as a new set of find methods rather than updating the existing ones (eg. by sub-classing the existing InstanceProperties), as the latter approach would probably break existing repositories that have no way of handling such an extended InstanceProperties.

grahamwallis:
@cmgrote - I thought we had raised an issue about that but I cannot find it.

This would be a nice enhancement. My original thinking was to subclass the InstanceProperties to create a more powerful MatchProperties object. However, I do take your point about how a repository connector would manage to tell the difference and behave sensibly if it does not support the sub-class.

A key consideration for the design of the enhanced API is whether, instead of matchCriteria applying to all property-conditions, there should be separate conditions and combinators allowing more complex conditions to be constructed (e.g. (A or B) and !(C and D). Need to be careful though not to make either the class interface (or grammar) too complex for its own good.

cmgrote:
Good point — wondering if we do this in stages and do not initially handle nested / sub-expressions (leave that to if / when we eventually develop a search-specific “language” or for a search index)?

cmgrote:
Based on our Tuesday call, I'm moving ahead with a structure like the following:

```json
{
  "class": "EntityFindRequest",
  "typeGUID": "0db3e6ec-f5ef-4d75-ae38-b7ee6fd6ec0a",
  "pageSize": 10,
  "matchProperties": {
    "conditions": [
      {
        "property": "displayName",
        "operator": "like",
        "value": {
          "instancePropertyCategory": "PRIMITIVE",
          "primitiveDefCategory": "OM_PRIMITIVE_TYPE_STRING",
          "primitiveValue": "(?i).*\\Qaddress\\E.*"
        }
      },
      {
        "property": "updatedTime",
        "operator": ">",
        "value": {
          "instancePropertyCategory": "PRIMITIVE",
          "primitiveDefCategory": "OM_PRIMITIVE_TYPE_DATE",
          "primitiveValue": "123456789"
        }
      },
      {
        "nestedConditions": {
          "conditions": [
            {
              "property": "...",
              "operator": "...",
              "value": {
                "instancePropertyCategory": "...",
                "primitiveDefCategory" : "...",
[...]

PR #2888, opened 2020-04-08
Revised find operations for #2857

cmgrote:
Creating as draft request so we can debate the naming, endpoints, etc if desired...

- Interface definition, including new search objects and operators
- New `findEntities` method and REST endpoints
- Initial implementation of `findEntities` for in-memory repository

Still todo:

- [x] Implementation of new `findRelationships` method and REST endpoints
- [x] Initial implementation of `findRelationships` for in-memory repository
- [x] ~~Unit tests (is there something already there I should extend?)~~
- [x] Inclusion of the method as a new profile in the CTS (probably as a separate issue and PR: see #2895 )

Also discussed:

- I have purposely _not_ extended the InstanceElementHeader class in the new beans (other than the top-level request bean), because this would force us to need to explicitly provide the "class" property in many places within the payload, and I believe this is redundant (and overly complicates the request) -- therefore I've gone directly for Serializable in the new beans. (Confirmed on Thursday call that this is acceptable as we do not anticipate subclassing these sub-beans, as they are only used by this search interface and if they were to be subclassed then we would invalidate the backwards compatibility of the interface.)

### DIFF
--- a/developer-resources/License-Example-Files/License_for_POM_Files.xml
+++ b/developer-resources/License-Example-Files/License_for_POM_Files.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>1.7-SNAPSHOT</version>
+        <version>1.8-SNAPSHOT</version>
     </parent>
 
     <name>Add module name here</name>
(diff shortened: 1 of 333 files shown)


cmgrote:
Merging into master for 1.8, as discussed on Tuesday call...

PR #2933, opened 2020-04-16
Glossary view ui implementation

pmarius22:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/glossary/glossary-view.js
```diff
@@ -17,13 +17,77 @@ class GlossaryView extends PolymerElement {
...
     }
 
+    _retrieve_all_glossaries() {
+        console.debug('searching: '+ this.q);
+        var types = [];
+
+        this.$.combo.selectedItems.forEach(function(item){
+            types.push( item.name);
+        });
+
```
bogdan-sava, line 86: please remove this code.
It is not needed.

PR #2937, opened 2020-04-17
Reference Data Management

mandy-chessell:


mandy-chessell:
This pull request includes the addition of valid values management to Asset Owner OMAS and Digital Service OMAS.  There is a major update to the UML models for Area 4 (governance) and Area 7 (lineage)

Issue #2938, opened 2020-04-18
swagger with filter set on is not working


bogdan-sava:
swagger UI with filter option is not working properly and ends up in a javascript error.

<img width="1051" alt="Screen Shot 2020-04-18 at 8 32 29 AM" src="https://user-images.githubusercontent.com/40362260/79629135-52d03e00-814f-11ea-9cc1-a8164fe1df8a.png">

Without filter parameter seems to work fine.

bogdan-sava:
sometimes loads with 'No operations defined in specs' message:
<img width="1312" alt="Screen Shot 2020-04-18 at 9 04 43 AM" src="https://user-images.githubusercontent.com/40362260/79629711-66ca6e80-8154-11ea-943c-8b08e0aa50fd.png">

planetf1:
Agreed - let's remove it

PR #2939, opened 2020-04-18
switch off swagger ui filter option as it is not working properly

bogdan-sava:
https://github.com/odpi/egeria/issues/2938


bogdan-sava:
yes

PR #2943, opened 2020-04-20
UI enable loading.path

bogdan-sava:


bogdan-sava:
the response is on the issue #2944

planetf1:
Agree with comments in issue. I misinterpreted change. Seems a good change

PR #2948, opened 2020-04-20
Asset Catalog OMAS - map and array properties are not displayed correctly

danielaotelea:


open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/builders/AssetConverter.java
```diff
@@ -325,4 +336,13 @@ private Element lastElementAdded(Element tree) {
                 ADDITIONAL_PROPERTIES_PROPERTY_NAME,
                 instanceProperties, methodName));
     }
+
+    private String listToString(List<String> list) {
+        return String.join(",", list);
+    }
+
+    private String mapToString(Map<String, Object> map) {
+        return map.keySet().stream().map(key -> key + "=" + map.get(key))
```
bogdan-sava, line 345: Are you sure you want to use '=' and not ':' ?
An issue that briefly shows current incorrect behavior and the wanted behavior would be useful.
Also a unit test that assert the new output is correct would be useful.

PR #2950, opened 2020-04-20
Asset Lineage OMAS - small refactoring

danielaotelea:


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/util/Converter.java
```diff
@@ -46,36 +59,34 @@ public LineageRelationship createLineageRelationship(Relationship relationship)
...
-
-            String propertyValue = propertyValueToString(property.getValue());
-            if (propertyValue.equals(""))
-                continue;
-
-            if (property.getKey().equals("name"))
-                attributes.put("displayName", propertyValue);
-            else
-                attributes.put(property.getKey(), propertyValue);
+        String methodName = "extractProperties";
```
GuyAckermans, line 73: Methodname is instancePropertiesToMap

Issue #2935, opened 2020-04-17
Unexpected Exception from Governance Engine OMAS


mandy-chessell:
The following errors are occurring while running the Coco Pharmaceuticals labs:

```
Fri Apr 17 10:54:38 BST 2020 cocoMDS1 Security OPEN-METADATA-SECURITY-0006 User Governance Engine is not authorized to issue a request to server cocoMDS1
Fri Apr 17 10:54:38 BST 2020 cocoMDS1 Exception OMAS-GOVERNANCE-ENGINE-0008 The Governance Engine Open Metadata Access Service (OMAS) encounter an exception while processing incoming event {0}
```

Firstly the Governance Engine OMAS is not using the server's userId (serverUserName)
but the `componentName` in multiple places in `GovernanceEngineOMRSTopicListener`.  This userId is not valid and so the UserNotAuthorizedException is thrown.

The second problem is that the `OMAS-GOVERNANCE-ENGINE-0008` message is not being formatted correctly.  It is produced by the `logExceptionToAudit()` method.  Note the call should be `logException()` rather than `logMessage()`.  The `instanceEvent` should be passed to `getMessageDefinition()` and the throwable should be the last parameter on the `logException()` call.

```java
private void logExceptionToAudit(OMRSInstanceEvent instanceEvent, Throwable e) {
        auditLog.logMessage("Governance Engine OMAS is processing an OMRSTopic event.",
                GovernanceEngineAuditCode.EVENT_PROCESSING_ERROR.getMessageDefinition());
    }
```
In addition `logExceptionToAudit()` is called from multiple places and there is no distinguishing property in the message to identify the origin of the error.   A simple fix would be to add methodName to the message.

planetf1:
This is noticeable when running the lab notebooks if/when a user refers to the logs - similarly in a real system.

We should fix this asap 

I'm proposing to add a reference in the release notes for 1.7

mandy-chessell:
There has been a fix go in for this problem and so it is very close to being remedied.  Now we have a proper audit log message and the UserNotAuthorizedException stack trace is logged.

```
Mon May 04 16:38:38 BST 2020 cocoMDS1 Security OPEN-METADATA-SECURITY-0006 User cocoMDS1 is not authorized to issue a request to server cocoMDS1
[...]

PR #2953, opened 2020-04-21
Unexpected Exception from Governance Engine OMAS #2935

danielaotelea:


mandy-chessell:
This fix corrects the problem with the exception logging but does not resolve the root problem that the userId being passed is a fixed string rather than the server's userId.

PR #2966, opened 2020-04-23
Use correct types for CSVFile properties

mandy-chessell:


planetf1:
PR build is failing in Data Platform OMAS API though I appeared to build ok....

planetf1:
CTS(inmem) failed with 16 errors. See https://ibm.box.com/s/7mqfgcbhe18lbbwfjrm9cb4b15d52v5h

The errors are EntityNotKnown exceptions in reference copy tests

Will repeat without this PR to see if it's a PR_specific regression, or a current master issue - in which case we may as well merge anyway

planetf1:
Prior to this PR the CTS (inmem) passes with 0 failures. Retesting with PR again.

planetf1:
Something has gone odd in the github - merge . 
When the PR was initially opened it built ok. I was also able to do this locally, and experienced the cts failures.
Later in the day I clicked on 'This branch is out-of-date with the base branch' as we often do -- the result was a failed build
I pulled down the PR manually and merged with master. No conflicts detected but the build result is the same locally as seen on the build system - problems in the data platform OMAS.
It looks as if something could be a bit odd with the git history or prior merges. 

@mandy-chessell suggest you pull or rebase from master and test locally, then push/force-push. One of us should then retest the CTS - in case the previous failure was a 1-off or an artifact of merge oddities.

planetf1:
Ok. I think what's happened is that commit 8e2627630ca196141beb403faef301e5bc1c75eb made some 'in-flight' changes ie to DataPlatformProperties.java which were pushed to origin, and thus into this PR -- before the PR had been merged, and this commit contains an incomplete (work in progress change)  and isn't buildable.

Need a stable branch for the PR - an option is to create a new branch and then cherry-pick the selected changes in needed for this fix. 

once any *upstream* branch contains the fix I'll cherry-pick for the release

mandy-chessell:
I think we need to separate the PR from the specific commit needed to go into the 1.7 branch.

On this PR - I thought it was merged yesterday and I have continued with the work on the Data Platform OMAS fix, not realising that my intermittant changes were going into the open PR. I will get it into a buildable state again sometime today so it can be merged.

It should be possible to just take the commit for 1.7 and test it on the branch.
The specific commit for 1.7 on its own should not break the CTS.  The change to OMRS fixed the formatting of an error message.  The change to the type of the properties was in the OMAS layer which is not tested by the CTS.

planetf1:
1.7 PR created just from the csv fix commit. 

One option - If you need to 'freeze' for a PR, whilst concurrently continuing development at pace :-) , you can create a branch from your development branch, and use that for the PR. cherry-picks can be added for any fixes found during review and added back to the PR branch from dev branch. This allows for review time without blocking dev work. Just one possibility of so many with git!

planetf1:
I ran a test of CTS (inmem) after merge - all runs cleanly. Unsure why the early test failed - perhaps due to work-in-progress or a glitch.

PR #2788, opened 2020-03-24
Add CloudInformationModel for tutorial environments

planetf1:
See issue odpi/egeria-jupyter-notebooks#34 for more info

* Updates path in information architect notebook for CloudInformationModel (archives)
* Builds CloudInformationModel.json during build
* Adds CloudInformationModel.json to assembly, so that it's findable in container/archive environment

Testing ...

Issue #2778, opened 2020-03-23
Data Engine OMAS - update integration tests


popa-raluca:
* Review and update use cases
* Add tests for failing requests
* Add tests for update use cases

PR #2978, opened 2020-04-24
Git2941 Owl RDF Canonical glossary parser  

davidradl:


.gitignore
```diff
@@ -89,12 +89,10 @@ egeria-lineage-repositories/**
 *-graph-repository/**
 
 # Ignore archives
-CloudInformationModel.json
-CloudInformationModelGUIDMap.json
-DataStoreConnectorTypes.json
-DataStoreConnectorsGUIDMap.json
-OpenMetadataTypes.json
+*.json
```
planetf1, line 92: I wonder if this exclude is rather too general -- will we not have .json files in future?

If the files are GENERATED they should be under the target tree - those files are already entirely excluded from git.

See below for unit test too - we should generate them into a target directory

This helps keep excludes under control, and also makes it easier for packaging and assembly

planetf1:
In #2788 I tried to fix the creation -- not completed as I hit an issue -- of the archive for CIM, and the fix is bad hence closed, but will revisit under #2778 in the 1.8 timescale (also changes needed to other clients) and so this could be good to cover also.

.gitignore
planetf1: Another reason for this is that other build lifecycle aspects like 'clean' will work properly without extra work, as well as making it easier to run other utilities across the source - for docs etc.

Unfortunately in both the UI and the subject area tests this wasn't done which caused a fair bit of heartache to fix up some consequential issues.

davidradl: @planetf1 I agree normally. I have changed the logic so the json file is generated from the model name - so we do not know it advance. It is the existing logic for the archive writer to write the output here - I have not changed that. As far as my change -there is a json file which is the archive which is picked up from here; this is the same as a the CIM model. There is a guid file is generated (also existing logic) - but you would not necessarily want it deleted as part of a maven clean - as it is the file that ensures subsequent builds use the same guids.    I suggest we raise a separate issue to review the input and output files locations as I see this as a separate issue; we can amend the .gitignore at that stage if required. Is that ok with you? 
I will look at the junit as that should write to target.

davidradl:
@planetf1  I have amended the code so the junit does not write to the file system (googling around this seems like bad practice). I have changed the gitignore to be more targetted.

planetf1:
> @planetf1 I have amended the code so the junit does not write to the file system (googling around this seems like bad practice). I have changed the gitignore to be more targetted.

Thanks. even better if we don't need to write.

I notice in the gitignore we have:
```
*Archive.json
*GUIDMap.json

# Ignore any json files generated during junit
open-metadata-resources/open-metadata-archives/design-model-archives/*.json
```
Where do these get written to (I haven't had a chanbce to try it)

They should only get written to the target directory. Adding anything created elsewhere is bad practice, and can cause problems with packaging.

Part of this is that we need to ensure that 'mvn clean' restores the files to the same state beforehand.. but there are other implications ie on wildcards for packaging. Best practice is to output all files into 'target' only.

If that's the case no gitignores would be needed.

We do breach this slightly on the top level with the config/registry files (there is an issue open to address this)

davidradl:
@planetf1. The reason for the need for the .gitignore is not to ignore files that are created during the build. The reason is that, in a dev environment in IJ, when you run the archive parser writer builder, then it looks for the the guid file in the root folder, it generates the archive where the model file is. I then copy the archive file into the root folder. I then start the omag server. I then issue the rest call to load the archive - the filename being in the rest body. This is how I get Egeria to pick up the archive. I have not changed this mechanism, this was the same for cim, but it had hard coded names.

So the .gitignore is required to ensure that in a dev environment these files are not inadvertently put into git. I suggest any enhancement to this existing mechanism be pursued in a separate issue.

planetf1:
Agreed - thanks for the explanation. JUst wanted to check I wasn't missing anything done in the build -- which as you say we'll tackle in the related issue.

davidradl:
I intend to merge this at the beginning of April the 29th, unless I receive more feedback

open-metadata-implementation/repository-services/repository-services-archive-utilities/src/main/java/org/odpi/openmetadata/repositoryservices/archiveutilities/OMRSArchiveGUIDMap.java
```diff
@@ -88,8 +88,21 @@ public String  getGUID(String id)
 
         if (guid == null)
         {
-            guid = UUID.randomUUID().toString();
-
+            boolean duplicate =true;
```
mandy-chessell: What is it likelihood of getting a GUID clash from RandomUUID given it uses a timestamp?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/api/rex/RepositoryExplorerController.java
```diff
@@ -496,7 +496,7 @@ public RexTraversalResponse rexTraversal(@RequestBody RexTraversalRequestBody bo
         try {
             serverName          = body.getServerName();
             serverURLRoot       = body.getServerURLRoot();
-            enterpriseOption    = body.getEnterpriseOption();
+             enterpriseOption    = body.getEnterpriseOption();
```
mandy-chessell: Fix alignment

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
```diff
@@ -295,18 +326,65 @@ protected String addTerm(String       glossaryId,
                              List<String> categoryIds,
                              String       qualifiedName,
                              String       displayName,
-                             String       description)
+                             String       description) {
```
mandy-chessell, line 330: Formatting

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
```diff
@@ -326,6 +404,9 @@ protected String addTerm(String       glossaryId,
             {
                 if (categoryId != null)
                 {
+                    if (categoriesAsNames) {
```
mandy-chessell, line 409: Formatting

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
```diff
@@ -754,6 +835,70 @@ protected void addLinkToBead(String                 beadId,
...
+
+        archiveBuilder.addRelationship(archiveHelper.getRelationship(ISA_RELATIONSHIP_NAME,
+                                                                     idToGUIDMap.getGUID(generalizededTermId + "_to_" + specializedTermId),
+                                                                     null,
+                                                                     InstanceStatus.ACTIVE,
+                                                                     end1,
+                                                                     end2));
+    }
+    protected void addHasARelationship(String conceptQName, String propertyQName) {
+        String conceptId = idToGUIDMap.getGUID(conceptQName);
```
mandy-chessell, line 862: Formatting - problem continues futher down file

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/owlcanonicalglossarymodel/CanonicalGlossaryOwlArchiveWriter.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private String modelLocation;
+
+
+    /**
+     * Default constructor
+     *
+     * @param modelLocation directory name for the model's JSON-LD files.
+     */
+    CanonicalGlossaryOwlArchiveWriter(String modelLocation)
```
mandy-chessell, line 25: Should this name be more specific as only works for industry models ?

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/owlcanonicalglossarymodel/properties/ModelElement.java
```diff
@@ -0,0 +1,226 @@
+/* SPDX-License-Identifier: Apache 2.0 */
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.archiveutilities.designmodels.owlcanonicalglossarymodel.properties;
+
+import java.util.*;
+
+/**
+ * Attribute describes an attribute in the model
+ */
+public class ModelElement
```
mandy-chessell, line 10: Javadoc missing on methods - this is a generic comment as occurs in other files

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/test/JenaTest.java
```diff
@@ -0,0 +1,549 @@
+/* SPDX-License-Identifier: Apache 2.0 */
```
mandy-chessell: Seems this is in the wrong package - should be in the OWL package

davidradl: I think you are looking at an old version of the fix this file is not in the current fix

open-metadata-implementation/repository-services/repository-services-archive-utilities/src/main/java/org/odpi/openmetadata/repositoryservices/archiveutilities/OMRSArchiveGUIDMap.java
davidradl: I think you are looking at an old version of the fix this code is not in the current fix

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/api/rex/RepositoryExplorerController.java
davidradl: I think you are looking at an old version of the fix this file is not in the current fix

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
davidradl: fixed

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
davidradl: I think you are looking at a back level fix - this is not in the current fix

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/owlcanonicalglossarymodel/CanonicalGlossaryOwlArchiveWriter.java
davidradl: We hope that others in the community will adopt this way of modelling if it suits them, there is nothing in the format that is intended to be IBM specific.

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/base/DesignModelArchiveBuilder.java
davidradl: fixed

open-metadata-resources/open-metadata-archives/design-model-archives/src/main/java/org/odpi/openmetadata/archiveutilities/designmodels/owlcanonicalglossarymodel/properties/ModelElement.java
davidradl: fixed

PR #2979, opened 2020-04-24
Git2977 amend getEntityNeighbourhood behaviour for inmemory repositories to handle classifications properly

davidradl:


planetf1:
may be useful to run CTS (it was running clean on master yesterday) on in-mem after a fix?

davidradl:
@grahamwallis I am not sure I understand your concern.  You say "Although it will permit the root entity to pass there is a logical AND condition in the overall predicate, so if the neighbour entity is not appropriately classified it would fail, I think." That is the behaviour I was trying to achieve: 
- if either of the entities is root then that entity is included.
- if neither of them are root then we apply the classification test to each
- if either of them is root then the non-root entity should have the classification test applied to it.  

What am I missing?

grahamwallis:
@davidradl Hi David - You are correct - apologies, I hadn't followed all the way through to the validator method, which will pass on any of the required classifications being present.

PR #2982, opened 2020-04-27
git2981 rename subject area Lines to short names

davidradl:


davidradl:
subject area OMAS only

Issue #2989, opened 2020-04-28
Verify that  Kafka is running before starting the KafkaTopicConnector


wbittles:
A number of different issues have been reported which have been caused by Kafka not being available when Egeria attempts to produce or consume events.

These "bring up" issues have been encountered primarily during Egeria cohort initialisation, but issues have  been reported sending and consuming events.  Resulting in important cohort initialisation message being lost.   

In order to avoid handling each of these events individually, with each occurrence requiring it's own retry handler with error and audit  logging code.

In order to avoid these bring up error handlers, this change will delay the start of the Egeria KafkaTopicConnector until the Zookeeper cluster reports that Kafka has initialised enough to have running brokers.

mandy-chessell:
What happens to the calling components?  Do they get an exception?  Are the events buffered?  
Ho do things restart when kafka/zookeeper come back?

planetf1:
would be helpful to link/respond to the previous issues raised, especially if this proposal supercedes the previous discussion/fix

wbittles:
@planetf1 
There have been a number of reported "bring u"p type issues encountered in the past and this fix will handle all of those. If we image there are different stages to bring up
1. Zookeeper bootstrap
issues HOST_NOT_FOUND
2. Zookeeper reachable but cluster not configured yet.
issue TOPIC_NOT_FOUND
3. Zookeeper cluster is up, but no brokers are running.
issue a Kafka Exception (Execution or timeout)

This fix means that all of the above stages must have completed successfully before we complete successfully. If we can't complete the Exception that occurred last will be reported. 
So If initially Egeria couldn't bootstrap to zookeeper , this exception would be stored. If Egeria bootstraps then encounters another issue then it's that last issue that will be reported, until successful.
In Jeff's specific case , I tested the creation of an Egeria Cohort without a running Kafla cluster , and the cohort failed to initialise because of this fix catching the exception, so I'm confident this fix prevents Egeria from encountering the problem Jeff reported as we will either ride out the condition by waiting or failing.

wbittles:
@mandy-chessell The calling components do receive an exception 
Here is an abridged version the start and end of the exception trace.
[stacktrace.txt](https://github.com/odpi/egeria/files/4551834/stacktrace.txt)

In the case of the cohort test, the cohort failed to start, because this fix detected that a Kafka Cluster wasn't running.
In the case of a single server , the first OMAS that tries to connect receives the exception and the server stops.
[...]

Issue #2595, opened 2020-02-14
When Kafka is temporarily unavailable during startup, OMRS cohorts never get initialized


jnhagelberg:
I have put together a fix for this issue which I will be submitting for review shortly.

We have seen this in a few cases where the Kafka pod gets into a weird state in our environment.  This is not a common scenario, but we have seen it enough times that it makes sense to do what we can to automate the recovery when Kafka does come back up.

What we have done here is put in a retry mechanism that kicks in if the initial connection to the cohort fails.  The retry mechanism is totally configurable and can be turned off If desired.  With the changes, the first initialization attempt is done synchronously, the same way it is now.  If the initialization attempt fails, then an additional attempt is scheduled to run in a background thread if certain conditions are met.  If that fails, we schedule an additional retry attempt.  We do this until the configured maximum number of retries have been exhausted.  The default behavior is that there is one retry thread shared between all of the cohorts.  This can be configured programmatically by applications that embed egeria.   In addition, the default number of retries is 20, and by default we wait 10 seconds between retries.  These are also configurable.  There are certain kinds of initialization errors we exclude from the mechanism.  These are for cases where we are sure that doing a retry will not resolve the issue.

The changes also introduce a new `CohortConnectionStatus`: `AWAITING_RETRY`.  If the intialization of the cohort manager fails in a way that is retryable (and the max retries have not been exhausted), the connection state is set to `AWAITING_RETRY`.  When the retry starts, the state will go to `INITIALIZING` and resume its usual progression.  If the initialization fails with an error that is not retryable, the state will stay at `CONFIGURATION_ERROR` (and no retries will be scheduled)

jnhagelberg:
I've created https://github.com/odpi/egeria/pull/2596

planetf1:
I added some comments to PR #2596 - seen before I saw the issue (catching up)

planetf1:
We believe #2989 will address this scenario, and the PR in #2990 is now merged.

Please re-open if the fix is insufficient

PR #2990, opened 2020-04-28
Delay the start of the KafkaTopicConnector, untill kafka has running …

wbittles:
…brokers Issue #2989

Signed-off-by: William Bittles <wbittles@uk.ibm.com>


wbittles:
@jnhagelberg @planetf1 @mandy-chessell 
Here is the change that was discussed earlier

I tested a number of scenarios

1. Tried to create an  Egeria cohort with no Zookeeper/Kafka cluster available 
Worked as expected, with Exception showing that the new code prevented any cohort initialisation messages being sent.
2. Ran through the Postman tutorial without a zookeeper/Kafka cluster available
Again this worked as expected with this fix preventing the Asset Lineage OMAS from connecting to a topic.
3. Ran the Postman tutorial with a zookeeper/cluster available.
Worked as expected.
4. Ran the Postman tutorial , starting the instance before starting the Zookeeper/Kafka cluster. The zookeeper/egeria cluster was then started while the KafkaTopicConnector was retrying and when the cluster was available Egeria connected.

I did encounter one issue, while testing 
https://github.com/apache/kafka/pull/2861

On my system ,the default timeout on a client connection to Zookeeper was extremely long, so it took a long time to be informed that the attempt to connect to Zookeeper had failed.

mandy-chessell:
With this fix, how does the server recover?  It seems the initialization of each of these component will fail and this will result in the server start up failing.

planetf1:
#2595 specifically relates to a fix where the hostname of the kafka broker cannot be resolved - temporarily. This results in a different response to a connection timeout. How would you expect the code to respond to this, is it tested, and is this fix a suitable replacement for that PR in that regard?

planetf1:
I tried out the PR with a single test -- start the cohort with the broker config pointing to a host called 'banana' (which didn't exist).
Started up.... I had some exceptions from ASSET_CONSUMER reporting the AdminClient failure (as host not found) but server still seemed to start.
I then allows the host to resolve, but no recovery occurred and therefore no cohort membership
This is what I'd expect from the code.

I think it's useful to have better error reporting, but the server did not end up in a recovered state so this is not replacing the other fix?

Will try a few more things later/tomorrow.

planetf1:
I would approve this fix if we can ensure the following
 - The current defaults are 2 retries 10ms apart. I would recommend something more like 10 retries, 10000 ms apart (100 seconds total)
 - An audit log entry is written during the retry phase, at least every  minute (and not every loop iteration)
It benefits from being a focused, simple & sufficient fix (=good)
 - We align the naming of the new audit event with the other events

Asynchronous startup is a longer discussion outside the scope of this issue/pr, as is the server status reporting - which if present is a bug I'll raise separately.

wbittles:
@planetf1 
You need to consider the Kafka timeout, which on a default RHEL system defaults to 1588025312853
With a down Kafka Cluster, initialising an Egeria Cohort  would take (1588025312853 + 10000) x 10
before it failed.  In a lot of scenarios Egeria will need to fail fast. There is also the issue that if a Kafka Cluster does become available , it could be 100 seconds before we are able to attempt another connection. For most situations I'd advise a high retry count with a small sleep time.
With these defaults I was able to start Egeria begin the retry loop , and then go and start a zookeeper docker image, then start a Kafka broker get them up and running , before Egeria exited the retry loop.
I posted a link above to a, two year old, not merged, Kafka PR, but I suspect it's related to the OS's TCP/IP settings. 
Sorry I don't follow the next point , In waitForBrokers we call auditLog.logMessage just before we call getRunningBrokers

Fair point on the last issue, it is a bit "It's Kafka's Fault"

planetf1:
Agree the timeout needs to be considered - though to clarify in my case I was verifying the 'HostNotFound' case. Here the kafka API call will fail immediately -- 

Other cases will behave differently -- a host not responding but not refusing connections is always going to be a difficulty with all protocols, and so they'll all have a timeout. In addition kafka itself may do some internal retries. perhaps it does this in the 'refusing connections' case. 

What this may mean is that rather than a fixed retry/iterations we need to have a 'minimum time between retries' as well as a number, since
 - the low default in the current implementation is crazy fast for the HostNotFound case - 2 iterations of 10ms delay is of no help. I think therefore a change is needed. Such a tight 10ms delay imo could be problematic too. 

 - I wasn't clear on your calculation but I think a min interval of 10s or so is as low as it makes sense going. perhaps a little smaller, but I'm wary of tight loops. BUT by interval I mean the time between successive calls (to which a small 10ms guard could be added) since we now now some fail immediately, others with a timeout
 - my suggestion does indeed lead to a high delay *by default* - and here I think the best approach is to add to the documentation of the connector to at a minimum clarify our behaviour with this retry
 - I didn't observe any retry AUDIT log message. I will retry (perhaps after we've finished the above discussion/any changes?)

planetf1:
Here's an example part of my log (from yesterday):
```
Wed Apr 29 18:19:02 BST 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC-CONNECTOR-0001 Connecting to Apache Kafka Topic egeria.omag.openmetadata.repositoryservices.cohort.devCohort.OMRSTopic with a server identifier of 73b97297-114f-48e3-857c-191781c7692a
Wed Apr 29 18:19:02 BST 2020 cocoMDS2 Error OCF-KAFKA-TOPIC-CONNECTOR-00014  Connecting to bootstrap Apache Kafka Broker egeria.omag.openmetadata.repositoryservices.cohort.devCohort.OMRSTopic}
Wed Apr 29 18:19:02 BST 2020 cocoMDS2 Exception OMRS-AUDIT-0006 Configuration error detected while connecting to cohort devCohort, exception org.odpi.openmetadata.frameworks.connectors.ffdc.ConnectorCheckedException was caught with error message: OCF-KAFKA-TOPIC-CONNECTOR-400-002  Egeria was unable to initialize a connection to a Kafka cluster.  The message in the exception was: Failed to create new KafkaAdminClient
Wed Apr 29 18:19:02 BST 2020 cocoMDS2 Exception OMRS-AUDIT-0006 Supplementary information: log record id 3b494d8f-4e16-497b-ab00-4bd4d12df209 org.odpi.openmetadata.frameworks.connectors.ffdc.ConnectorCheckedException returned message of OCF-KAFKA-TOPIC-CONNECTOR-400-002  Egeria was unable to initialize a connection to a Kafka cluster.  The message in the exception was: Failed to create new KafkaAdminClient and stacktrace of
OCFCheckedExceptionBase{reportedHTTPCode=400, reportingClassName='org.odpi.openmetadata.adapters.eventbus.topic.kafka.KafkaOpenMetadataTopicConnector', reportingActionDescription='KafkaMonitor.waitForThisBroker', reportedErrorMessage='OCF-KAFKA-TOPIC-CONNECTOR-400-002  Egeria was unable to initialize a connection to a Kafka cluster.  The message in the exception was: Failed to create new KafkaAdminClient', reportedErrorMessageId='OCF-KAFKA-TOPIC-CONNECTOR-400-002 ', reportedErrorMessageParameters=[org.apache.kafka.common.KafkaException, egeria.omag.openmetadata.repositoryservices.cohort.devCohort.OMRSTopic, Failed to create new KafkaAdminClient], reportedSystemAction='The system is unable initialize.', reportedUserAction='Ensure that Kafka is available', reportedCaughtException=org.apache.kafka.common.KafkaException: Failed to create new KafkaAdminClient, reportedCaughtExceptionClassName='org.apache.kafka.common.KafkaException', relatedProperties=null}
    at org.odpi.openmetadata.adapters.eventbus.topic.kafka.KafkaOpenMetadataTopicConnector.start(KafkaOpenMetadataTopicConnector.java:249)
```

wbittles:
@planetf1 to be honest I can't recreate it, I boot strapped against nohost:2181 and Zookeeper ignored it and Egeria started normally connected to Kafka.
It would be really useful to see the exception to be absolutely sure, but I could ride out that OS level type of problem separately, I could verify that the server is reachable and responsive before attempting to bootstrap to Zookeeper. 
If in a Kubernetes environment it's common for an Egeria Cohort to be started  before a zookeeper server has been attached to the network. I'm happy to follow your guidance on this and I like the extra layer of robustness.
The decision to be made  would be what server to actually check for, the first one of the bootstrap.servers list ? 

If I put the above check in, then we will know that server is responding to tcp/ip requests and we won't need to worry about the minimum time after that stage in bring up has been reached.
If someone goes and actually starts configuring their OS/Kafka default settings, we need to honour that  and not add extra sleep time. Once we hit the Kafka code properly that tight loop becomes 5 minutes long, with my default settings, and starting a cohort could take an hour to fail.

I agree re the documentation, just not sure where and how it should be delivered. Considering the changes that have taken place , I'd like to re-verify our existing default settings, in an attempt to make the quality of service a bit more predictable, but that for another PR.

So depending on the results of your testing, does this sound like a way forward ?

planetf1:
@wbittles We don't directly communicate with zookeeper from Egeria. We specify bootstrap servers, but these are kafka brokers. Even for the AdminClient we specify a *broker*. Of course at the back-end kafka makes use of zookeeper - but that's hidden using these APIs. Indeed I know that on one cloud service (IBM Event Streams) there is no direct access to zookeeper - we must go through Kafka.

Therefore I don't understand 'I boot strapped against nohost:2181 and Zookeeper ignored it and Egeria started normally connected to Kafka.'

I can explain what I did. I was running Egeria locally. I have kafka and zookeeper installed and running locally also, on default ports, and installed via homebrew. I ran up the 'coco' topology ie three egeria servers on ports 8080,8081,8082. In the server configuration notebook I changed 'eventBusURLRoot' to be 'banana:9082'. 'banana' doesn't exist. I started egeria & had the behaviour above. (later on I added banana to /etc/hosts to map to 127.0.0.1 and allow it to resolve)

So in this case the AdminClient has no chance of connecting to anything -- the host cannot be resolved

This isn't identical to Jeff's situation, but it was my attempt at doing something similar.

So this was unrelated to k8s. Jeff does have a deployment there... as to whether it's common? Generally I would say in most cases kafka is a very foundational capability and we'd expect it to be reliable and running. In our labs/charts we include a deployment of kafka. For tutorials we do no waiting, but for the vdc (virtual data connector) chart we have specific code in the pod initialisation to wait for kafka to become available. This just uses 'netcat' to ensure the host:port specified is reachable. no more. - its this local install if anything that could result in a slight timing issue. 

In terms of which server to check, I think the answer (as you had already come up with?) is 'any'. For example I've deployed into IBMCloud and I think got 10 brokers! Locally there's only 1. When we played with VDC we had 3. Any broker can be the 'bootstrap' since the first server connected to also informs the clients (based on the broker's cached metadata from zookeeper) of the other servers in the cluster.

@jnhagelberg can you elaborate again on your scenario - so we can figure out if this works for you?

Finally, users may of course change their kafka settings which we don't want to interfere with -- but we do need to explain the interaction with our logic. I still think there needs to be a clear way to configure an appropriate retry time period -- and I don't think we have that nailed yet. A number of retries with a minimum clock time between retries would be enough... and I would err on defaulting around the several minutes timescale

wbittles:
There is now a minimum sleep time, which is only called if the bring up checks blocks for less that 5000ms  configurable and Egeria will retry for 10 attempts again by default. If a bring up check takes longer than 5000ms Egeria won't sleep at all , I've also aligned the message id 

Sat May 02 11:02:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC-CONNECTOR-0001 Connecting to Apache Kafka Topic egeriaTopics.server.cocoMDS1.omas.assetlineage.outTopic with a server identifier of 29ef0305-716d-4cba-bae2-f5db2cf15ce0
Sat May 02 11:02:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 1
Sat May 02 11:03:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 2
Sat May 02 11:04:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 3
Sat May 02 11:05:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 4
Sat May 02 11:06:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 5
Sat May 02 11:07:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 6
Sat May 02 11:08:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 7
Sat May 02 11:09:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 8
Sat May 02 11:10:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 9
Sat May 02 11:11:14 BST 2020 cocoMDS1 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 10
Sat May 02 11:12:14 BST 2020 cocoMDS1 Error OCF-KAFKA-TOPIC-CONNECTOR-00014  Connecting to bootstrap Apache Kafka Broker egeriaTopics.server.cocoMDS1.omas.assetlineage.outTopic}

planetf1:
I will aim to test later today & post an update review. thanks

wbittles:
good spot @grahamwallis , it's fixed in latest push

planetf1:
I've tested this locally
 - Host Not Found: we do retries on the configurable interval & number of times specified 
 - Added resolution for the host but with zk and kafka down. retry continues (correctly)
 - started zookeeper - as expected retry continues
 - stopped zk, started kafka - as expected retry continues
 - started both zk and kafka -- initialisation completes and server comes up

One MINOR issue - I to see the retry loop as reporting 
```Mon May 04 15:54:40 BST 2020 cocoMDS2 Startup OCF-KAFKA-TOPIC_CONNECTOR-0015 The local server is attempting to connect to Kafka, attemp 33```

Note the typo on 'attempt' - as @grahamwallis mentioned. This was with the fix download within the last two hours (no additional pushes)

This can be fixed in this PR if you can push - or I can open a new issue to fix

On this basis I think the fix achieves what it set out to do, and improves resiliancy of the connector.



I did note that if we run out of retrying across all servers the higher level of the stack doesn't behave correctly - but this is outside the scope of the connector. Opened #3021 to track

File: open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnectorAuditCode.java [hunk 1]
```diff
@@ -100,9 +100,19 @@
...
+             "Add the missing property to the event bus properties in the server configuration."),
+
+    SERVICE_FAILED_INITIALIZING( "OCF-KAFKA-TOPIC-CONNECTOR-00014 ",
+              OMRSAuditLogRecordSeverity.ERROR,
+             "Connecting to bootstrap Apache Kafka Broker {0}}",
+             "The local server has failed to started up the Apache Kafka connector, Kafka Broker is unavailable",
+             "Ensure Kafka is running and restart Egeria"),
+    KAFKA_CONNECTION_RETRY( "OCF-KAFKA-TOPIC_CONNECTOR-0015",
+              OMRSAuditLogRecordSeverity.STARTUP,
+             "The local server is attempting to connect to Kafka, attemp {0}",
```
planetf1, line 112: This is the remaining typo

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnectorAuditCode.java
```diff
@@ -100,9 +100,19 @@
...
-
-    ;
+             "Add the missing property to the event bus properties in the server configuration."),
+
+    SERVICE_FAILED_INITIALIZING( "OCF-KAFKA-TOPIC-CONNECTOR-00014 ",
+              OMRSAuditLogRecordSeverity.ERROR,
+             "Connecting to bootstrap Apache Kafka Broker {0}}",
+             "The local server has failed to started up the Apache Kafka connector, Kafka Broker is unavailable",
+             "Ensure Kafka is running and restart Egeria"),
+    KAFKA_CONNECTION_RETRY( "OCF-KAFKA-TOPIC_CONNECTOR-0015",
```
planetf1, line 110: This is out of sync in numbering - 00014 vs 0015 - different 0s

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnectorAuditCode.java
```diff
@@ -100,9 +100,19 @@
...
-             "Add the missing property to the event bus properties in the server configuration.")
-
-    ;
+             "Add the missing property to the event bus properties in the server configuration."),
+
+    SERVICE_FAILED_INITIALIZING( "OCF-KAFKA-TOPIC-CONNECTOR-00014 ",
+              OMRSAuditLogRecordSeverity.ERROR,
+             "Connecting to bootstrap Apache Kafka Broker {0}}",
+             "The local server has failed to started up the Apache Kafka connector, Kafka Broker is unavailable",
+             "Ensure Kafka is running and restart Egeria"),
```
planetf1, line 109: we don't need to restart Egeria, just the logical server

PR #3008, opened 2020-04-30
Correct the project scope description and Subject Area FVT  

davidradl:


davidradl:
I plan to merge at the start of the 5th of May unless there is feedback.

Issue #1498, opened 2019-09-06
Egeria UI has debug enabled by default


planetf1:
When the Egeria UI is launched, debug output  is enabled by default it
```
  egeria git:(h2update2) ui
 ODPi Egeria

       ______                        _                  __  __   ____
      / ____/ ____ _  ___    _____  (_) ____ _         / / / /  /  _/
     / __/   / __ `/ / _ \  / ___/ / / / __ `/        / / / /   / /
    / /___  / /_/ / /  __/ / /    / / / /_/ /        / /_/ /  _/ /
   /_____/  \__, /  \___/ /_/    /_/  \__,_/         \____/  /___/
           /____/


 :: Powered by Spring Boot (v2.1.6.RELEASE) ::

2019-09-06 10:13:48.565  INFO 57955 --- [           main] o.o.o.u.a.EgeriaUIApplication            : Starting EgeriaUIApplication on Nigels-MBP with PID 57955 (/Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/user-interfaces/access-services-user-interface/target/access-services-user-interface-1.1-SNAPSHOT.jar started by jonesn in /Users/jonesn/IdeaProjects/egeria)
2019-09-06 10:13:48.571  INFO 57955 --- [           main] o.o.o.u.a.EgeriaUIApplication            : No active profile set, falling back to default profiles: default
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.codehaus.groovy.reflection.CachedClass (jar:file:/Users/jonesn/IdeaProjects/egeria/open-metadata-implementation/user-interfaces/access-services-user-interface/target/access-services-user-interface-1.1-SNAPSHOT.jar!/BOOT-INF/lib/groovy-2.4.15-indy.jar!/) to method java.lang.Object.finalize()
WARNING: Please consider reporting this to the maintainers of org.codehaus.groovy.reflection.CachedClass
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
2019-09-06 10:13:50.892  INFO 57955 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Bootstrapping Spring Data repositories in DEFAULT mode.
2019-09-06 10:13:51.096  INFO 57955 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Finished Spring Data repository scanning in 171ms. Found 3 repository interfaces.
[...]

PR #3022, opened 2020-05-04
UI logging level

bogdan-sava:
https://github.com/odpi/egeria/issues/1498


planetf1:
Actually I realised I fixed the original issue in d55e3cef96, and set logging to be on for spring (so we could get the critical errors like a port clash) but off for egeria - which is our general policy for Egeria

So the question here is what the UI's design is. Will it be using egeria's audit logging - or a similar mechanism, in which case logging should be off.

Or is it going to use regular logging -- which is different to the platform. And in this case would you only want to log issues from the UI perspective rather than inherit the somewhat ad-hoc (where it exists) logging from other modules - which would make identifying issues harder?

I would prefer consistency with the chassis?

bogdan-sava:
I saw that! For Spring yes. But this solves only a corner case. Like error in spring framework.
What about the INFO:
`EgeriaUIPlatform    : Started EgeriaUIPlatform in 7.25 seconds (JVM running for 8.048)`
This is an important message to be seen in the logs.
Also the warning regarding the fact that strict ssl is set to false.
`Strict SSL is set to false! Invalid certificates will be accepted for connection!`

Initially design of the UI was to be a spring boot app that is one of the consumers of the egeria APIs.
Regarding the consistency, of course hope everybody is in the favor of it. But they are different things. My understanding of the need of having audit logging in egeria server-chassis is that because that logging is to be read and used by other than technical ops engineers. This is not the case for the UI. By definition this is the purpose of the UI to be used by the end-user. The logs of the UI are technical logs used only by ops. The action of the user in the UI produces audit log on server chassis.

Issue #2693, opened 2020-03-02
Inadequate packaging of utilities & FVTs in egeria-omag assembly


planetf1:
We have a few 'utilities' which are designed to be 'run' at the command line, so typically include a 'main'

Currently it is not described how to run these utilities - other than knowing that they work in an IDE environment. The required classes are not setup in an easy to use manner.

In my view the simplest approach to this is to build a jar with dependencies. That way the 'utility' can be run as `java -jar myutility.jar parm1 parm2` etc - they are self contained

First this is how the directory currently looks:
```
  egeria-1.5-distribution git:(egeria-release-1.5) ✗ cd egeria-omag-1.5
ls
➜  egeria-omag-1.5 git:(egeria-release-1.5) ✗ ls
LICENSE           clients           samples           user-interface
NOTICE            conformance-suite server            utilities
➜  egeria-omag-1.5 git:(egeria-release-1.5) ✗ cd utilities
➜  utilities git:(egeria-release-1.5) ✗ ls
design-model-archives-1.5.jar   open-connector-archives-1.5.jar open-metadata-types-1.5.jar
➜  utilities git:(egeria-release-1.5) ✗ pwd
/Users/jonesn/src/egeria/open-metadata-distribution/open-metadata-assemblies/target/egeria-1.5-distribution/egeria-omag-1.5/utilities
➜  utilities git:(egeria-release-1.5) ✗ ls -la
total 424
drwxr-xr-x   5 jonesn  staff     160  2 Mar 11:11 .
drwxr-xr-x  10 jonesn  staff     320  2 Mar 11:11 ..
-rwxr-xr-x   1 jonesn  staff   38636 27 Feb 20:14 design-model-archives-1.5.jar
-rwxr-xr-x   1 jonesn  staff   10722 27 Feb 20:14 open-connector-archives-1.5.jar
-rwxr-xr-x   1 jonesn  staff  163171  2 Mar 11:02 open-metadata-types-1.5.jar
➜  utilities git:(egeria-release-1.5) ✗
➜  utilities git:(egeria-release-1.5) ✗ jar -tf open-metadata-types-1.5.jar
META-INF/MANIFEST.MF
META-INF/
org/
org/odpi/
org/odpi/openmetadata/
org/odpi/openmetadata/opentypes/
META-INF/maven/
META-INF/maven/org.odpi.egeria/
META-INF/maven/org.odpi.egeria/open-metadata-types/
[...]

PR #3025, opened 2020-05-04
#2670 remove direct log4J ref in clients

### LINKED ISSUE #2670: Upgrade log4J to v2
Egeria is using  log4j:log4j  which is affected by CVE-2019-17571
Moving to the newer org.apache.logging.log4j:log4j-core >= 2.8.2 should address this

planetf1:

Fixes #2670
(by removal)

* removes log4J from direct usage by client apps
* was already replaced by use of logback
* updated/added logback.xml where required to match original intent of code

See #2670 for more info


planetf1:
Tested the samples. All seem to correctly output their text/results/prompts without debug info, except

Admin Server samples fails with
```
jonesn:admin-services-samples/ (issue2670) $ java -jar target/admin-services-samples-1.8-SNAPSHOT-jar-with-dependencies.jar                         [21:43:52]
Error: Could not find or load main class org.odpi.openmetadata.adminservices.samples.ConfigureMetadataServerSample
Caused by: java.lang.ClassNotFoundException: org.odpi.openmetadata.adminservices.samples.ConfigureMetadataServerSample
```
Propose this will be collected up with any required fixes to dependencies, packaging, entry points in #2693 . It is unrelated to this fix.

Have not tested the json import - this also needs the above work from #2693. Reviewers - suggest you check if it meets your needs when the utilities are run within the side

Conformance suite client looks ok

PR #3031, opened 2020-05-05
Data Platform OMAS

mandy-chessell:
Refactor of the Data Platform OMAS API to support database schemas.

Currently the interfaces do not do anything useful as they need the implementation of the RelationalDataHandler to save and retrieve metadata.

The reason to push the function now is to allow review of the new API.


mandy-chessell:
There is a comment on a new Access Service option called publishZones.  This defines the list of zones set in the asset when the "publish" API request is made.  It means that an integration connector can assemble a complex asset/schema in private zone and then publish it to zones where the consumers are listening.

Issue #1403, opened 2019-08-15
[crypto_used_network] use secure protocols for all network communications by default


cmgrote:
(This is required for **gold** status under CII.)

The software produced by the project MUST support secure protocols for all of its network communications, such as SSHv2 or later, TLS1.2 or later (HTTPS), IPsec, SFTP, and SNMPv3. Insecure protocols such as FTP, HTTP, telnet, SSLv3 or earlier, and SSHv1 MUST be disabled by default, and only enabled if the user specifically configures it.

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/other.md#crypto_used_network)

bogdan-sava:
I will enable HTTPS (and disable HTTP) for server-chassis also.
The default used certificate will be by default self signed and included in the source code.

It might affect notebooks and docker images that are using 8080 port and HTTP protocol.

As the UI is using by default 8443 port for HTTPS, in order to be able to run on the same machine my proposal is server-chassis to use by default 9443 port.

mandy-chessell:
Please make sur you include the updates to the tutorials and labs in your PR so there is no disruption to those consuming Egeria.

bogdan-sava:
Of course. Not sure I know all the places where changes are needed, but I'll add you, @planetf1 and @cmgrote as reviewers.

planetf1:
Thanks -- Can I suggest leaving the PR as 'draft' so we don't accidentally merge - as the notebooks etc will definately break without being updated accordingly

bogdan-sava:
Not sure how important is the default self-signed certificate, but I generated default self signed cert that will stay in the resources inside the jar like:

```
 keytool -genkey -alias tomcat -storetype PKCS12 -keyalg RSA -keysize 2048 -keystore keystore.p12 -validity 3650

Enter keystore password:  
Re-enter new password: 
What is your first and last name?
  [Unknown]:  Egeria ODPi
What is the name of your organizational unit?
  [Unknown]:  Egeria
What is the name of your organization?
  [Unknown]:  ODPi
What is the name of your City or Locality?
  [Unknown]:  San Francisco               
What is the name of your State or Province?
  [Unknown]:  CA
What is the two-letter country code for this unit?
  [Unknown]:  US
Is CN=Egeria ODPi, OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US correct?
```

planetf1:
A very good question.. I'd be inclined to go with your suggestion.

[...]

Issue #2378, opened 2020-01-08
Document how run egeria from official packages efficiently


planetf1:
The official build of egeria is effectively the files we lay down in /Users/jonesn/src/egeria/open-metadata-distribution/open-metadata-assemblies/target

See also #2375

As we move to more dynamic loading, the target tree will diverge more from the source tree - for example in terms of where data files are put, where dynamically libraries are loaded (perhaps the biggest impact)

Some developers may already make use of the assemblies, whilst others (including me) tend to run the jars directly at the command line, and others run the code from within intellij

The simplest, easy way of running egeria components from both the IntelliJ IDE and command line, making use of the asemblies, should be clearly documented. As part of this we need to ensure as developers modify their classes, they can quickly and simple test their changes with minimal impact - for example this may need additional IntelliJ 'run' configurations.

Primarily documentation but we may identify code improvements

planetf1:
One reported issue has been running the server chassis jar & getting connectors dynamically loaded.

This does work
 1. Create a run configuration as usual for the server chassis (JAR)
 2. set 'VM options' to something like '-Dserver.port=8083 -Dloader.path=/Users/jonesn/src/egeria/open-metadata-resources/open-metadata-deployment/docker/egeria/target/assembly/server/lib/discovery-service-connectors-1.5-SNAPSHOT.jar'

I actually created 4 - with only the server port different, so on ports 8080,8081,8082, 8083. This matches our typical tutorial environment.

With this done I was able to run through the config, start,discovery notebooks without errors.
If the loader.path is omitted, the notebooks fail as expected - being unable to load classes

Note
 - I did not use the 'spring' capability specific in IntelliJ to the paid-for edition
 - I only tried a single dynamically loaded module

One 'tension' between the intellij environment & command line is that at the cli maven will generally build the right things, is coded to build the assemblies, and it's easy to then run the code directly from the assembly output

In the IDE the objective is quick turnaround.

[...]

PR #3056, opened 2020-05-13
use secure protocols for all network communications by default

bogdan-sava:
A first part change for #1403


planetf1:
Thanks for looking at this - it's a super step forward :-)

I see you've touched on many of the areas around the container environments too. I'll add some more detailed comments on the review.

BEFORE we merge though, can we make sure between us that we TEST the k8s deployments and docker environments as we have people using them from master actively.

I wonder if leaving as draft until we're sure could be a plan.

If the PR review is good, I'll try and test - at least initially by the end of the week

developer-resources/Spring.md
```diff
@@ -26,11 +26,11 @@ For example, this is the declaration used for the OMRS REST APIs
 @RequestMapping("/open-metadata/repository-services")
 ```
 
-This URI follows the root URL of the server.  So if the server was using `http://localhost:8080`, the methods are called
+This URI follows the root URL of the server.  So if the server was using `https://localhost:9443`, the methods are called
 using
 
 ```
-http://localhost:8080/open-metadata/repository-services ...
+https://localhost:9443/open-metadata/repository-services ...
```
planetf1, line 33: A big question here is whether we enable https ONLY or if we listen on BOTH ports. 

My initial thought was both - but given all interactions are moving to secured it's probably ok to do https only, as long as it 'just works' with self signed, works fine when developers debugging - in that case simply changing the ports is ok

open-metadata-implementation/admin-services/admin-services-configuration.postman_collection.json
```diff
@@ -202,12 +202,12 @@
 					"raw": ""
 				},
 				"url": {
-					"raw": "http://localhost:8080/open-metadata/admin-services/users/garygeeke/servers/cocoMDS1/server-user-id?id=cocoMDS",
-					"protocol": "http",
+					"raw": "https://localhost:9443/open-metadata/admin-services/users/garygeeke/servers/cocoMDS1/server-user-id?id=cocoMDS",
```
planetf1, line 205: Another key question - are we ok with 9443 as the default. I know you mentioned this in the issue and indeed referred to 8443.

I'd actually be inclined to suggest the chassis (ie core egeria) sticks with 8443 - and we move the UI to 9443.

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/LoginFilter.java
```diff
@@ -35,9 +33,15 @@ protected LoginFilter(String urlMapping, AuthenticationManager authenticationMan
...
-        return getAuthenticationManager()
-                .authenticate(new UsernamePasswordAuthenticationToken(
-                        request.getParameter("username"), request.getParameter("password")));
+        String username = request.getParameter("username");
+        String password = request.getParameter("password");
+        Authentication authentication =  getAuthenticationManager()
+                .authenticate(new UsernamePasswordAuthenticationToken( username, password));
+
+        if(authentication.getAuthorities().isEmpty()){
+            throw new InsufficientAuthenticationException("NO authorities for the user: " + authentication.getPrincipal().toString());
```
planetf1, line 42: Typo? Should we just have 'No' ?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/resources/application.properties
```diff
@@ -6,20 +6,21 @@ server.compression.enabled=true
...
+server.ssl.key-store=classpath:keystore.p12
+server.ssl.key-store-password=egeria
+server.ssl.keyStoreType=PKCS12
+server.ssl.keyAlias=tomcat
 
-# set to false if you want to ignore invalid ssl certificate of a server you want to connect to
+# WARNING! setting 'false' allows java clients to open https connections without checking certificates validity
+# Alternate you can import self signed certificates into java truststore or setup an truststore only for this app
+# by adding the store into server.ssl.trust-store parameter
 strict.ssl=true
```
planetf1, line 17: will it work OOTB with strict.ssl=true? If not how do we ensure it's obvious/simple/clear to get up (less of an issue in the container env where we can easily setup the certs - or disable for tutorial (less preferable))

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
```diff
@@ -15,8 +15,8 @@ metadata:
 spec:
   type: {{ .Values.service.type }}
   ports:
-    - port: 8080
-      targetPort: 8080
+    - port: 9443
+      targetPort: 9443
       nodePort: {{ .Values.service.nodeport.core }}
```
planetf1, line 20: Logically we probably should update the node ports to be suitably numbers - it will still work without, but it seems more complete - ie rather than 30080 maybe it's 30043 (ports need to be 30000-32999 I think) etc

open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
```diff
@@ -23,29 +23,29 @@ LABEL org.opencontainers.image.vendor = "ODPi" \
       org.opencontainers.image.version = "$version" \
       org.opencontainers.image.documentation = "https://egeria.odpi.org/open-metadata-resources/open-metadata-deployment/docker/egeria/" \
       org.opencontainers.image.ext.vcs-date = "$VCS_DATE" \
-      org.opencontainers.image.ext.docker.cmd = "docker run -d -p 8080:8080 odpi/egeria" \
-      org.opencontainers.image.ext.docker.cmd.devel = "docker run -d -p 8080:8080 -p 5005:5005 -e JAVA_DEBUG=true odpi/egeria" \
+      org.opencontainers.image.ext.docker.cmd = "docker run -d -p 9443:9443 odpi/egeria" \
```
planetf1, line 26: The ports are also documented on docker hub so I'll need to update there (easy just shouldn't forget)

open-metadata-resources/open-metadata-deployment/docker/egeria/Dockerfile
```diff
@@ -23,29 +23,29 @@ LABEL org.opencontainers.image.vendor = "ODPi" \
...
 COPY --chown=egeria:egeria target/assembly /opt/egeria
 
 # Expose port 8080 (default) for client access, and allow for 5005 being used for remote java debug
-EXPOSE 8080 5005
+EXPOSE 9443 5005
 
 WORKDIR /opt/egeria
 
 # Require use of numeric userids to meet some container security checks
-USER 8080:8080
```
planetf1, line 48: interesting .. nothing to do with ports we just decided to make the userid the same! I'd probably not update (indeed I wonder if it should just be 1111 or similar! - but that's outside the scope of this change)

developer-resources/Spring.md
planetf1: Interestingly if using a service mesh like istio in k8s there's a certain appeal in allowing individual containers to continue using simple http - without any need to deal with certificates & keystrokes, and using the mesh framework to manage the proxying through https. This is a pretty standard pattern. We may actually be making this a little more complicated by 'doing the right thing' and assuming https. It may not change the fact https is a good default, but I wonder if this is another reason we need to keep http - and figure out how to make configurable.

Will research some more on what is best practice

planetf1:
After our discussion this morning
 - I agree we go with https only
 - there was a suggestion of using a non-standard port to reduce chance of clashes (mostly when run locally/for dev). We didn't agree final port
 - agreed we could likely include the trust store in the omag assembly even though not in jar

planetf1:
I tested this briefly
 -We need the trust store included/ and/or docs before this can be integrated
 - Even with STRICT_SSL=false set, in both standalone & docker-compose the notebooks fail with certificate validation errors. Slightly odd as that should bypass the check
 
I'm going to move the PR to draft -- it's excellent we persue this initiative now but I think it best to guard better against accidental merging as currently lack of the above will break many users and tutorial content.

bogdan-sava:
I have added the trustsores, but there is another issue, if the path to the trustore in wrong it crushes.
If set the path right for intellij it crushes for running from jar and the other way around.

With STRICT_SSL=false is very weird to have certificate issues as it returns true no matter what. 
Can you please paste stack trace ?

bogdan-sava:
A possible solution would be to activate profiling and switch from application.properties to yaml configuration and have default configuration with strict.ssl set to true and false for production profile.

planetf1:
The profile approach would work if this was a build-time decision -- but I think more likely we want a runtime decision - it should be configurable. Generally I would suggest we reserve build-time switches only to support development processes / build pipelines / specific code-owner cases .

I need to focus on 2693 for a little longer, so may take a couple of days to retest this. I didn't get a stack trace with STRICT_SSL, it just reported an error that certificate validation failed - the same as without. Very odd I agree. I may have made an error, or maybe am not understanding something about how the configuration is inherited here -- but we already use this in the UI - and in the docker-compose/k8s environment where the UI works and it's ok there.

We need to ensure we can still
 - allow a dev to easily run egeria standalone at cli
 - allow dev to easily run egeria in intellij etc
^^ on these two we can remember they are devs, so some minor setting adjustment is ok as long as we document it. #2378 also has some relevance here -- we expect people to run egeria from our ASSEMBLIES normally... but also need to ensure they can simply do it in their IDE
 - ensure the docker image can be used standalone
^^ ok I think if we need to modify the docs to suggest setting STRICT_SSL if needed, or to add pointer to truststore
 - ensure the docker compose environment for tutorials works ok without any extra configuration
 - tutorials in k8s
 - vdc chart in k8s (which also pulls in igc/atlas connectors)
^^ and for all of these we should make those changes automatically ie by setting env appropriate - or much better, to include the truststore

Until then we can't make default as critical things will break (including what we need to run education/labs)

Hope to be able to help with this soon but it may not make 1.8

bogdan-sava:
This is what I am thinking also: allow a dev to easily run egeria standalone.

The profiles I am talking about are not maven profiles but springboot profiles, which means using application.yaml instead fo application.properties which allows switch on rintime.

```
environments:
    dev:
        strict.ssl=false
...
    prod:
        strict.ssl=true
...

make the dev env default, but state it clear is not for production (well known for the ops deploying spring-boot apps ) .

Official documentation is here:
https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config-yaml

What do you think?

```

bogdan-sava:
Another solution is to place truststore in the root of the project, so path to the truststore would be simple 
`server.ssl.trust-store=truststore.p12`
It works in intellij as the default working dir is the root of the project. 
But, on the runtime with jar, must be in the working directory of JVM otherwise it will crush with FileNotFoundException, or must be configured properly.

planetf1:
I think that's a good solution @bogdan-sava .

Within the containers we can either add the command line parm, or it can be overriden via 
SPRING_PROFILES_ACTIVE environment variable

I'm still not sure why strictssl didn't work, but this solution is better IMO -- we should add ref in docs. Just need to also check logging configuration can be overriden within the image also

I'm going to return to testing this - when do you think you'd be able to add the profile support?
I can first check our container environments then work ok in dev mode, but ALSO then look at prod mode (which is a better default there -- especially as we can easily insert the trust store in any case)

planetf1:
I think I've identified an approach that easily allows the existing notebook environment to work regardless of whether we are secure or not.

This is only a short term fix, as it injects additional settings into the containers to disable certificate checking (3 different settings needed) - without changing a lot of our python code.

A *subsequent* change would be to deploy the trust-store, and remove these settings, but I think it's a prudent transition step.

I can either
 a) refactor & check in as a pre-req fix
 b) Submit a PR to the @bogdan-sava branch
 c) Directly push to the branch  (if permitted - it's an option when creating the PR)
 d) add the fix after we merge this - less preferable as there'll be a short break

I think c) is the best

Only 2 files so far for compose - and I've tested server config, startup, asset management & CTS (at least starting it), & cohort config

I will continue and check k8s now. I wanted to ensure we have a short term workaround prior to the final change -- in part as the code changes may be more diverse, and it would be easier to work on in a new PR.

planetf1:
permissions -> https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

bogdan-sava:
Yes! Changes are allowed for the maintainers since beginning.

planetf1:
Super. I will push the first two fixes.

I have now made some changes to the lab k8s chart, which again allows this to work.

To repeat though - consider this a 'compatability' fix rather than a 'utilize certs' fix -- that will come in another PR.

I need to next look at the vdc chart and do the same there ...  then we'll be close to a good initial PR to push through.

planetf1:
Pushed a further change to allow the vdc chart to work - again a workaround, as we still need to inject the certificates in a later change or PR.

However now the container environments are working ok with the change to https.

That leaves ensuring they work standalone

However from my experience with the containers -- that means one of the following must be true
 * The jvm under which the user is running egeria must a) have the truststore referenced or b) our spring config needs to default to a profile with strict.ssl=false -- and this may depend on how egeria is launched.....
AND
either
 a)  The user's environment where they run the notebooks from somehow needs to disable cert validation -- this can be via 2 env vars, but they would need to be SET BY THE USER -- so an impact. For developers this MAY be acceptable.
or
b) The notebooks themselves will need updating directly to add `,verify=false' on all of our useages of the python request module -- and similar. Or we work out how to package a workaround directly into the notebook -- since we can't control the environment of a developer so easily..... and jupyter is outside the scope of egeria - it's a client app -- so whilst we can add info about the truststore in the docs I think we WILL need to do this

I'll look more back at the notebooks next.,

For now, as above the fixes so far do allow our containerized notebooks & vdc environment to work properly - albeit skipping cert validation



@bogdan-sava

bogdan-sava:
Thanks @planetf1 .
Sounds very good.
I am looking also to the options. 
Basically certificate is in the keystore.p12 for outbound and in the truststore.p12 for inbound.
keystore is loaded from classpath and is fine.
trustore is simply not working to be loaded with springboot properties even in the docs says could be also in the classpath. Simply not working.

The way is working is setting jvm options like:

```
-Djavax.net.ssl.trustStore=truststore.p12 
-Djavax.net.ssl.trustStorePassword=***
```

bogdan-sava:
Now it works from intellj without any other configuration.
Is loading truststore from root.

In order to override the trustore the user must override the spring-boot properties or set the jvm options.
If jvm options are set, spring-boot properties are ignored.

planetf1:
Good news - will try with the latest change.. though there's a build error:
```
[INFO] --- maven-dependency-plugin:3.1.2:analyze-only (analyze) @ server-chassis-spring ---
[WARNING] Used undeclared dependencies found:
[WARNING]    jakarta.annotation:jakarta.annotation-api:jar:1.3.5:compile
[WARNING]    org.springframework:spring-core:jar:5.2.6.RELEASE:compile
[INFO] Add the following to your pom to correct the missing dependencies: 

```
I'll try a fix for that now

planetf1:
With spring-core added to server chassis I'm getting some challenges in resolving annotations - which were in EE in J8 and below, then in javax.annotation, and most recently moved to jakarta.annotation.  Looks as if there is some conflict also involving tomcat - perhaps caused by the new spring core dependency. Still working through this

planetf1:
Pushed a fix for the build clash which wasn't spring-core, but rather around PostConstruct.

Since server chassis picking up tomcat embedded, this also picks up tomcat annotations - which we're not using (build time). excluded.

Pushed fix to the PR branch

Will test..

planetf1:
Also added fix for user-chassis-spring

planetf1:
Updated notebooks with common code to skip cert validation (for now)

planetf1:
Further changes to fix notebooks. All now tested successfully - one minor issue found in the ui testnotebook @grahamwallis 
```

instanceCreateResponse: {'class': 'RelationshipResponse', 'relatedHTTPCode': 400, 'actionDescription': 'getEntitySummaryFromStore', 'exceptionClassName': 'org.odpi.openmetadata.repositoryservices.ffdc.exception.EntityNotKnownException', 'exceptionErrorMessage': 'OMRS-GRAPH-REPOSITORY-400-004 The attempt to retrieve an entity with GUID 7aa74355-1377-449a-a618-552a62999481 failed in getEntitySummaryFromStore method of class org.odpi.openmetadata.adapters.repositoryservices.graphrepository.repositoryconnector.GraphOMRSMetadataStore to open metadata repository Schema_Server', 'exceptionErrorMessageId': 'OMRS-GRAPH-REPOSITORY-400-004', 'exceptionErrorMessageParameters': ['7aa74355-1377-449a-a618-552a62999481', 'getEntitySummaryFromStore', 'org.odpi.openmetadata.adapters.repositoryservices.graphrepository.repositoryconnector.GraphOMRSMetadataStore', 'Schema_Server'], 'exceptionSystemAction': 'The system was unable to perform the entity retrieval.', 'exceptionUserAction': "Correct the caller's code and retry the entity retrieval request."}
```
but will revisit that in a separate issue if needed

planetf1:
One thought (cc: @mandy-chessell ) - All client interaction with the server now needs to do one of:
 * Utilize the appropriate truststore
 * include the certificate in their own truststore
 * disable cert validation (as I did here in notebooks)

This includes the new utilities easily run with 'java -jar' and certainly warrants updated documentation

planetf1:
Some of the things still left to do
 - improve documentation
 - utilize the certificates properly in the container environments
 - check clients - including FVTs
 - add release note for the next release
 - advise collaborators on slack

My inclination is to move these into a seperate PR/issue and integrate this change

One key point - ANYONE interacting with egeria at the rest level will need to move to https. Are you ok with this @mandy-chessell  ?

Will do a little further testing ....

planetf1:
Updated various docs, clarified requirement to disable SSL certification in postman, curl, http

planetf1:
Further doc updates.

SpringRESTClientConnector.java has now had a 'bypass' function added to skip SSL checks.

This allows existing client library to work as-before - but this is an INTERIM measure. We need to extend the configuration to allow a certificate authority/trust store to be referenced - in a future PR

planetf1:
Admin client sample is fine. However the other samples fail in different ways and need debugging -
In addition:
Subject Area OMAS is not using the common spring rest connector
Also noticed the CTS tests the ‘origin’ API call but directly uses restTemplate - which will also fail
Security officer also uses it’s own rest client

planetf1:
Egeria UI asset search - working ok
Repository explorer - working ok
Type explorer - fails to connect

mandy-chessell:
Hello @planetf1 - thanks for all of your testing - you did the right thing not merging until you are sure - I will keep dojo using unsecured protocols and we can change it after next week when we are sure.

bogdan-sava:
@planetf1 thanks for detailing testing.
Swagger UI should have the same self-signed certificate as it is configured on tomcat not on the app level.
Type explorer is attempting to connect to the server-chassis directly? Because it shouldn't. I'll also  continue investigating.

planetf1:
Public holiday today in the UK so I won't be able to progress this today.

Tomorrow I will aim to add a clearer up-to-date summary of what does and doesn't work, and try and break down the remaining issues we need to clear up prior to merging. 

Swagger UI was a red herring. In fact I subsequently realised I had the same issue with the UI. Both current versions of chrome and safari for me failed- they refuse to permit use of self signed certs - makes things tricky without adding cert to browser. Some products gen a cert on startup dynamically.. but a tricky area.. But I'll return to this tomorrow.

bogdan-sava:
I added also the logging for embedded tomcat.
Normally this will produce 2 lines regarding the protocol, port, and context path:

```
07:20:20.755 [main] INFO  o.s.b.w.e.tomcat.TomcatWebServer - Tomcat initialized with port(s): 9443 (https)
07:20:24.846 [main] INFO  o.s.b.w.e.tomcat.TomcatWebServer - Tomcat started on port(s): 9443 (https) with context path ''
```

Of corse, if tomcat runs into trouble will produce more logs, but I think this is wanted behavior.

planetf1:
I have updated #1403 with a status summary of where we area

planetf1:
When we started this PR it was with the intent of verifying the fixes and integrating into master.

We are now 99 commits later (albeit some merges) and still have a number of areas to look at -- and this is just for interop, let alone making full use of ssl cert checking.

So far the work has been on @bogdan-sava fork. I'm wondering -- if this is ongoing and/or needs wider collaboration -- we may be better off making this a full blown feature branch.

This will allow this PR to be integrated into that feature... and we can then have more independent working on specific areas causing issues.

planetf1:
Asset management samples fixed. Not ssl related -- the wrong users, servers, platforms were being used for the sample - inconsistent with the standard coco environment. Corrected & tested samples working with ssl.

planetf1:
Governance leadership sample mostly works - errors do not appear related in any way to ssl, so raised https://github.com/odpi/egeria/issues/3106

Governance zone creation worked once server name corrected

planetf1:
Subject area sample now works - added ssl workaround to subject area omas client. This is not currently using the spring rest connector, but should be. https://github.com/odpi/egeria/issues/3112 opened to address

planetf1:
* Type explorer working (https://localhost:9444 / cocoMDS1) + glossary authoring UI
* Discovery service working ok (from notebook sample)
* asset consumer fvt is an empty module... so no tests. Will leave in-situ as we hope to expand the variety of tests
 * Fixed subject area FVT to work in std coco config & tested with SSL

planetf1:
The containers are no longer working. Probably after 106d325055004b729569d0d63afebb62629f97ce

Error is:
```
 ODPi Egeria
    ____   __  ___ ___    ______   _____                                 ____   _         _     ___
   / __ \ /  |/  //   |  / ____/  / ___/ ___   ____ _   __ ___   ____   / _  \ / / __    / /  / _ /__   ____ _  _
  / / / // /|_/ // /| | / / __    \__ \ / _ \ / __/| | / // _ \ / __/  / /_/ // //   |  / _\ / /_ /  | /  _// || |
 / /_/ // /  / // ___ |/ /_/ /   ___/ //  __// /   | |/ //  __// /    /  __ // // /  \ / /_ /  _// / // /  / / / /
 \____//_/  /_//_/  |_|\____/   /____/ \___//_/    |___/ \___//_/    /_/    /_/ \__/\//___//_/   \__//_/  /_/ /_/

 :: Powered by Spring Boot (v2.3.0.RELEASE) ::

14:16:57.339 [main] INFO  o.s.b.w.e.tomcat.TomcatWebServer - Tomcat initialized with port(s): 9443 (https)
14:17:19.756 [main] ERROR o.s.boot.SpringApplication - Application run failed
org.springframework.context.ApplicationContextException: Failed to start bean 'webServerStartStop'; nested exception is org.springframework.boot.web.server.WebServerException: Unable to start embedded Tomcat server
	at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:185)
```
...
```
Caused by: org.springframework.boot.web.server.WebServerException: Unable to start embedded Tomcat server
```
...
```
Caused by: java.io.FileNotFoundException: /opt/egeria/truststore.p12 (No such file or directory)
```

planetf1:
We have multiple copies of truststore.p12 checked in.

There should be a single source that we than copy to anywhere else it may be needed - whether in the development tree, or as part of the assemblies.

For now will take /truststore.p12 and add to assembly to get things working - and return to this aspect in a subsequent PR

planetf1:
VDC chart mostly works -- Atlas fails to start, but this is unrelated to the security change. Tracking in #3118

planetf1:
Tested CTS inmem (1 failure) and graph (0 failures.)
Raised https://github.com/odpi/egeria/issues/3121 to track inmem cts failure as this also occurs in master.

No code change required currently due to the way ssl checks are currently disabled

planetf1:
Still looking for possible solution to allow self-signed cert to work in Chrome 80 & above.

So far the only workarounds are
 * start chrome with --ignore-certificate errors
 * type 'thisisunsafe' on the error screen - click the page and type blind - it's a hidden workaround

Neither are really acceptable.

Also we should document cert creation in the build, possibly even allow gen of the cert.

for now I've used
```
 #!/bin/sh

 CERTSTORE=keystore.p12
 TRUSTSTORE=truststore.p12
 PASS=egeria
 #dName="CN=*, OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 dName="CN=localhost OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 san="dns:localhost"
 NAME=tomcat
 VALIDITY=3650
 BASE=~/src/egeria
 cd ${BASE}
 rm -f ${TRUSTSTORE}
 keytool -importkeystore -srckeystore ${JAVA_HOME}/lib/security/cacerts -destkeystore ${TRUSTSTORE} -deststoretype PKCS12 -deststorepass ${PASS} -srcstorepass "changeit"
 # Generating key
 keytool -genkeypair -alias ${NAME} -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore ${TRUSTSTORE} -validity ${VALIDITY} -noprompt -storepass ${PASS} -dName           "${dName}" -ext SAN="${san}"
 cp ${TRUSTSTORE} ${CERTSTORE}
 cp *p12 ${BASE}/open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources
```

Note we use the same store for server & client in this case. It's also based off the jvm truststore as a starting point, as we may need the CAs for other outbound https access. We also have dup files in the build currently - not needed.

Experimenting with the addition of 'subject alternative name' which appears to be wanted by chrome ...

planetf1:
There's reference to this behaviour at https://bugs.chromium.org/p/chromium/issues/detail?id=1075912&q=self-signed&can=2 - note that it differs between platforms, and indeed versions. In part it's related to a MacOS requirement in catalina 10.5.5 - which is where the requirement for subjectAlternativeName seems to come from.


The Apple reference is at https://support.apple.com/en-us/HT210176 which requires:
*  server certs/CAs using RSA must use key size >=2048 bits -> I think we pass on this
* hash algorithm must be SHA-2 -> looks ok on this too as we use SHA-256
* The DNS name must be in subject alternative name - commonname is not trusted. -> I'm trying this, but unsure what validation we may encounter. Let's start with localhost
* TLS server certs need an 'ExtendedKeyUsage' externsion (EKU) containing  id-kp-serverAuth OID. -> WHat is that!
* TLS server certs must be valid for 825 days or less -> changing that!

More explanation at http://blog.nashcom.de/nashcomblog.nsf/dx/more-strict-server-certificate-handling-in-ios-13-macos-10.15.htm?opendocument&comments

However still not working with
```
 #!/bin/sh

 CERTSTORE=keystore.p12
 TRUSTSTORE=truststore.p12
 PASS=egeria
 #dName="CN=*, OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 dName="CN=* OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 san="dns:localhost"
 NAME=tomcat
 VALIDITY=824
 BASE=~/src/egeria
 cd ${BASE}
 rm -f ${TRUSTSTORE}
 keytool -importkeystore -srckeystore ${JAVA_HOME}/lib/security/cacerts -destkeystore ${TRUSTSTORE} -deststoretype PKCS12 -deststorepass ${PASS} -srcstorepass "changeit"
 # Generating key
 keytool -genkeypair -alias ${NAME} -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore ${TRUSTSTORE} -validity ${VALIDITY} -noprompt -storepass ${PASS} -dName           "${dName}" -ext SAN="${san}" -ext EKU=serverAuth
 cp ${TRUSTSTORE} ${CERTSTORE}
 cp *p12 ${BASE}/open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources
```

planetf1:
Opened up an issue on chromium -> https://bugs.chromium.org/p/chromium/issues/detail?id=1090086 to try and get clarity on expected behaviour

planetf1:
With the improved certificate bypass now works (after I built properly!) - so hunch was correct. The issue was indeed the validity of the certs - which chrome/macos seems better at testing?:

Screenshot 2020-06-01 at 22 56 51

planetf1:
The UI cert is also invalid in a similar way, so we'll use it there too.

Once this PR is merged can revisit the correct cert attributes to use & remove duplication

planetf1:
UI now verified working for the cert bypass also using the same cert as above.

At this point we are likely good to go with merging into master.

Discussion planned on team call tomorrow with a view to branch for a 2.0 release and then merging this immediately before continuing with security enhancements.

planetf1:
Final script fragment
```
 #!/bin/sh

 CERTSTORE=keystore.p12
 TRUSTSTORE=truststore.p12
 PASS=egeria
 #dName="CN=*, OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 dName="CN=* OU=Egeria, O=ODPi, L=San Francisco, ST=CA, C=US"
 san="dns:localhost"
 NAME=tomcat
 VALIDITY=824
 BASE=~/src/egeria
 cd ${BASE}
 rm -f ${TRUSTSTORE}
 keytool -importkeystore -srckeystore ${JAVA_HOME}/lib/security/cacerts -destkeystore ${TRUSTSTORE} -deststoretype PKCS12 -deststorepass ${PASS} -srcstorepass "changeit"
 # Generating key
 echo keytool -genkeypair -alias ${NAME} -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore ${TRUSTSTORE} -validity ${VALIDITY} -noprompt -storepass ${PASS} -dName       "${dName}" -ext SubjectAlternativeName="${san}" -ext ExtendedkeyUsage=serverAuth -ext BasicConstraints="ca:false"
 keytool -genkeypair -alias ${NAME} -keyalg RSA -keysize 2048 -storetype PKCS12 -keystore ${TRUSTSTORE} -validity ${VALIDITY} -noprompt -storepass ${PASS} -dName            "${dName}" -ext SubjectAlternativeName="${san}" -ext ExtendedkeyUsage=serverAuth -ext BasicConstraints="ca:false"
 cp ${TRUSTSTORE} ${CERTSTORE}
 cp *p12 ${BASE}/open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources
 cp *p12 ${BASE}/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/resources
```

planetf1:
Merged in latest code from master 
 - Updated dojo related .md & .json (postman) files still referring to http port
 - merged other master updates including switch to 2.0-SNAPSHOT

PR #3071, opened 2020-05-19
Code reformats in subject area and removed unnecessary double "new HashSet"

immmus:
Signed-off-by: immmus <fonto.trg@gmail.com>


immmus:
> I brought in your reformatting changes into my workspace. I then tried to format using the default Intellij setting and it changed the formatted code. I think we need to be able to format files consistently in subject area. I agree the reformatting you did makes it look better, but I want to ensure that all formats in the subject area omas are done consistently and as owner I can format the file -without it changing. Let me know if I am missing anything.

In addition to reformatting, I made some corrections along the way if I noticed any flaws that you might not have noticed at the beginning. You requested some changes, can I clarify what should I fix?

davidradl:
@immmus I have tested , your change does not regress anything - we will need to be aware that if I go into these files , the formatting may change slightly. I think on balance it is better to put this in. I will merge today.

PR #3085, opened 2020-05-22
Initial drop of code for the presentation server

davidradl:


planetf1:
Added some notes to #3098 of some of the questions/todos we'll have based off this PR... which of course we are just using to checkin initially.

Issue #3091, opened 2020-05-23
Create my postcard file for Egeria Dojo


mandy-chessell:
This is a change to Egeria that is part of the Egeria Dojo session
[Making a contribution to Egeria - Step-by-Step](https://egeria.odpi.org/open-metadata-resources/open-metadata-tutorials/egeria-dojo/egeria-dojo-day-2-3-contribution-to-egeria.html).  It will create a new markdown file in `open-metadata-resources/open-metadata-tutorials/egeria-dojo/egeria-dojo-postcards` and link it to the readme.

PR #3108, opened 2020-05-27
CREATING NEW POSTCARD (#3091)

nilmeier:
Part of the dojo 
Signed-off-by: nilmeier <nilmeier@us.ibm.com>


planetf1:
Looks like a zero length file?

open-metadata-resources/open-metadata-tutorials/egeria-dojo-postcards/from-Jerome-Nilmeier.md
```diff
@@ -0,0 +1,12 @@
+Hello!  I am enjoying the dojo.
```
planetf1, line 1: normally we would put the spdx headers at the top then the text further down

planetf1:
Hi @nilmeier if you ping me (@planetf1) on odpi.slack.org I'd be happy to help you through making the change and explain why we use DCO.

Unfortunately without the DCO we can't integrate any change so I'll close this for now -- but we can reopen with a single click when you are ready.

PR #3122, opened 2020-05-30
Updates from Dojo Experience

mandy-chessell:
This PR containes a number of small fixes and doc improvements that were uncovered while running the Egeria Dojo Live.  This includes a better structure for the postman collections around administration, and better descriptions for configuring the server.  This was a section that participants particularly struggled with.

There are also fixes/enhancements to the admin services to prevent bad connectors being passed as configuration for the security connectors, audit log connectors and matadata archive connectors.  The validation has been added to the common error handler so it can be used by other admin calls that pass a connection object.

Finally there is the fix to the Swagger annotation for the platform origin rest call.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,11 +89,11 @@ egeria-lineage-repositories/**
 *-graph-repository/**
 
 # Ignore archives if they are in the top level directory
-./CloudInformationModel.json
-./DataStoreConnectorTypes.json
-./OpenMetadataTypes.json
-./*Archive.json
-./*GUIDMap.json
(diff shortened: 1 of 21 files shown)


mandy-chessell:
I have added the rename of the Data Platform OMAS to Data Manager OMAS.  This had a wide impact on the code - not just the implementation of the access services, admin etc.  The assemblies and packages were effected, configuration in the labs and the admin sample.  Plus numerous hits on the documentation

The services of the new Data Manager OMAS are not working (or if they are, it is only by luck as they are not properly tested :).

Issue #3123, opened 2020-06-01
Glossary Author in presentation manager does not handle refresh well 


davidradl:
Refreshing the browser loses the session and and the React context which means we lose logged in userid; so  when we issue rest calls, the userid is not in the url, leading to the rest calls failing. It is difficult for the user to know how to recover from this situation. 

This change is to make the glossary author more robust to a refresh. The change will be to notice that we have lost the userid, alert the user using a popup indicating what has happened and put up the login page so they can put in their credentials again and set up a new session;  

@grahamwallis I thought I would make this change to Glossary author first - as some stage we should move this and the connect logic to the top level so that all components benefit from this checking and recovery.

grahamwallis:
@davidradl Thanks David - as of Friday I added detection only to RepositoryExplorer - it does not attempt to recover from this situation. It will be much better to do the recovery automatically. Please let me know if I can help or when you have made the change - then I'll modify RepositoryExplorer accordingly.

PR #3124, opened 2020-06-01
#3123 handle browser refresh for the glossary author

davidradl:


davidradl:
This change has pick up some json files that should have been ignored using the .gitignore - looking into it.

PR #3130, opened 2020-06-01
Adjust egeria community bridge insights link

leonardotatsch:
The another its invalid, showing blank dash.


planetf1:
You will need to sign the DCO to make the change

( `git commit --amend -s` should work )

planetf1:
Many thanks for the proposed contribution. However without the DCO we cannot merge this so are closing and replacing.

Replaced by #3161

Issue #3131, opened 2020-06-01
Cannot resolve d3 when building presentation server


davidradl:
Can't resolve 'd3' in '/Users/davidradley/egeria/open-metadata-implementation/user-interfaces/presentation-server/nodejs/src/client/components/RepositoryExplorer/components/diagram'

davidradl:
if I issue npm install d3 it gives errors - but seems to work after that

grahamwallis:
Hi @davidradl @planetf1 - I was going to comment on the PR but switched to here as the PR is closed. For my development environment I used 'npm install d3' and was thinking that when we integrate the presentation-server into the Maven build we would do essentially the same, if a general 'npm install' does not spot the dependency (which I had assumed it would).

planetf1:
I think it's fine to just have a working recipe for now. Perhaps if you could capture that in the README, it would be a good starting point for getting into a maven build. I could try and work with it to put something together...

Makes sense the big moving parts are in there, and you know how you will deploy it -- at least for now - ie is it a jar, a war etc? Or indeed multiple.

davidradl:
@grahamwallis - I forgot the npm install trick. Thanks.

davidradl:
closing as I think this is not an issue if you run npm install

planetf1:
Sounds good - will verify for sure in #3098

PR #3133, opened 2020-06-01
#3131 add d3 dependancy

davidradl:


davidradl:
@planetf1 great comment Nigel.

davidradl:
I think @planetf1  is right this change is not right. After havin done an npm install --save d3;  I reset --hard back to master and now the npm start works . It seems that something is in my environment that is not in git. closing this pr. Will raise a new pr once we understand what is required.  @grahamwallis any thoughts?

planetf1:
As soon as you have code in a reasonable working state & know what artifacts you want to produce, we should probably get the maven build in place asap as this will improve consistency/reproducibiity

Issue #3035, opened 2020-05-06
Make encrypted configuration the default


cmgrote:
We have a request from one of our consumers to make our file-based configuration encrypted by default rather than clear-text -- which is related to #1387 .

I would suggest trying to incorporate this into 1.8 the same as the HTTPS by default.

Issue #1387, opened 2019-08-15
[crypto_password_storage] store passwords securely


cmgrote:
(This is required for lowest level **passing** for CII.)

If the software produced by the project causes the storing of passwords for authentication of external users, the passwords MUST be stored as iterated hashes with a per-user salt by using a key stretching (iterated) algorithm (e.g., PBKDF2, Bcrypt or Scrypt). (N/A allowed.)

This criterion applies only when the software is enforcing authentication of users using passwords, such as server-side web applications. It does not apply in cases where the software stores passwords for authenticating into other systems (e.g., the software implements a client for some other system), since at least parts of that software must have often access to the unhashed password.

This is a bare minimum today when storing passwords. Sometimes software needs to have a credential, such as a password, to authenticate it to other systems; those are intentionally out of scope for this criterion, because in many cases it's not possible to store them as iterated hashes using per-user salt.

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#crypto_password_storage)

Furthermore: The security mechanisms within the software produced by the project MUST generate all cryptographic keys and nonces using a cryptographically secure random number generator, and MUST NOT do so using generators that are cryptographically insecure. (N/A allowed.) [crypto_random]

A cryptographically secure random number generator may be a hardware random number generator, or it may be a cryptographically secure pseudo-random number generator (CSPRNG) using an algorithm such as Hash_DRBG, HMAC_DRBG, CTR_DRBG, Yarrow, or Fortuna. Examples of calls to secure random number generators include Java's java.security.SecureRandom and JavaScript's window.crypto.getRandomValues. Examples of calls to insecure random number generators include Java's java.util.Random and JavaScript's Math.random.

(see: https://github.com/coreinfrastructure/best-practices-badge/blob/master/doc/criteria.md#crypto_random)

cmgrote:
Two pieces here:

1. I don't believe we store any passwords for our own authentication (other than for demo purposes): that is, we are relying on external providers like LDAP in production scenarios?  If so, then for the first point (`[crypto_password_storage]`) we simply have `N/A`.
1. For the second point, so long as we use the `EncryptedFileBasedServerConfigStoreConnector` we are using cryptographically secure random number generator (`java.security.SecureRandom`) and thereby meeting `[crypto_random]` requirement.

PR #3146, opened 2020-06-03
Config encryption fixes and enhancements

cmgrote:
- Fixes overlap with `keystore.p12` file now present for https changes
- Adds environment variable-based option for holding the keyset for encryption (avoiding storing the encryption key in the same system where the encrypted config file itself is stored)
- Still defaults to generating a secure encryption key and storing in a secure location (as it did previously -- so backwards-compatible)
- Also replaces the various error logging and `IllegalStateException` throwing with proper coded errors

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -167,7 +167,7 @@
         <jsonld.version>0.13.0</jsonld.version>
         <commons-logging.version>1.2</commons-logging.version>
         <commons-codec.version>1.14</commons-codec.version>
-        <json-sanitizer.version>1.2.0</json-sanitizer.version>
+        <json-sanitizer.version>1.2.1</json-sanitizer.version>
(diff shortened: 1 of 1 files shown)


planetf1:
A good reminder that as #1403 is implemented that this is an important part of securing the server.

Will hold off from suggesting any further specific changes yet but will add a note to 1403 that we should return to it

cmgrote:
Latest commit updates the following, as discussed on Thursday's calls:

- Makes the encrypted file-based configuration store the default
- Automatically detects any pre-existing clear-text configuration document and encrypts it automatically
- Updates docs to reflect this change in default behaviour

As a result, this PR:

- Resolves #3035 
- Resolves #1387

cmgrote:
Merging based on successful tests thus far from myself and Nigel -- to give maximum time for further testing by others prior to release...

Issue #3147, opened 2020-06-04
do not load trustStore if strict.ssl is false


bogdan-sava:
The configureTrustStore currently is setting the trustStore according to server.ssl.trust-store propertiey if only is not set already on JVM level.
In the case 

If the user/ops engineer deliberately set strict.ssl to false the application still attend to load truststore from default location. For the case file is not present, error occurs.

**Desired behavior:** if strict.ssl is set to false, no truststore will be set as it will be ignored anyway.

PR #3148, opened 2020-06-04
do not load trustStore if strict.ssl is false

bogdan-sava:


planetf1:
Hi,
 Since I have a large set of changes still needed to complete the security updates, can you clarify the reason for this change? Can we consolidate the discussion in the issue #1403 so we can ensure we have a 

Was this change because there was a scenario you hit that I did not test/consider - and this is a fix?
Or is the intent to make an incremental improvement?
Or is it to try and implement part of 1403? Which is very much work in progress  - there's checklists currently but not details.

I would prefer we don't make changes in isolation without trying to tie them together with the issue so that we can both keep things working, ensure we each understand the many scenarios, the testing that is done, and are working to a consistent end goal.

bogdan-sava:
I have already open an issue for this PR: #3147 
**Scenario:**
1. successful build egeria and deply artifacts to the internal artifactory 
2. deploy artifacts from artifactory to an VM
3. as deployment is not included truststore, intended to use bypass strict.ssl and run like:
`jave -jar egeria.jar --strict.ssl=false `

**Result:**
command crushes with file not found truststore.p12 

**Solution:**
if strict.ssl=false should not even try to load truststore because it will be ignored anyway

This is current PR all about

planetf1:
Discussed and agreed good to merge.
Will fix the case where the jar is run directly

PR #3171, opened 2020-06-08
#3101 Ensure connector jars include required dependencies

### LINKED ISSUE #3101: Many missing connectors in OMAG assembly
encrypted connector needs checking , in assembly deployed to docker container

planetf1:

Fixes #3101

This change ensures that all connectors are present in the egeria server distribution, and additionally that each includes all required dependencies.

See discussion in #3101 for some issues with this approach - duplication & size.

Since this enables dynamic loading of connectors I propose we go ahead with this simple change, and once we have dynamic loading as the standard approach we can then look at optimization.


planetf1:
See comments in #3101 - we can either go with this as a first step, or defer this change to likely later release as I think ultimately a new design is needed to be scalable and meeting numerous differing requirements

PR #3176, opened 2020-06-09
#3123 Update the subject area omas to use the rest connector

davidradl:


davidradl:
@mandy-chessell .I updated 2 parts out of my ownership area
1) I enhanced the rest connector to add put and delete
2) I fixed issues in the view admin - as the individual services did not work. I raised #3177 to address other issues I noticed in the view admin.

davidradl:
I plan to merge at the beginning of Wednesday unless there is more feedback

Issue #3135, opened 2020-06-01
Update Presentation server to use secure protocols


davidradl:
The code as is uses unsecured protocol. I will look to change this one the main change has gone in #1403 . @planetf1 @grahamwallis.

I intend to use this issue to introduce configuration into the presentation server. I notice that there is a webpage that details a way of doing this 
https://medium.com/@felipedutratine/manage-config-in-nodejs-to-get-variables-from-file-and-env-variables-87526509aad1

I am thinking that we could use a similar method, with: 
1) we store the config in the config folder where the presentation server runs.
2) Our config files would be named to include the serverName e.g. 
presentation.server.<server-name>.config
I assume we would need to do a 'require' with a wildcard in the code.
3) There would only be one file for a one server in the presentation server config folder . 
4) It would be possible to run a presentation server using environment variables
5) It would be possible to run a presentation server with server1 environment variables. It would also be possible to run a presentation server with server2 environment variables at the same time.
6) In order to allow  5) we would need to ensure there were no clashes or ports urls etc.  
7) For this initial check in all servers would authenticate using the same user store. 
8) for the environment variable I am envisaging at least:

egeria_presentation_server_name = aaaa.    // this is the server name to use      
egeria_presentation_server_remote_servername_for_aaaa = viewserver1 // remote servername
egeria_presentation_server_name_remote_url_root_for_aaaa = https://viewhostName:8443 //remote url
9) Currently the nodejs server code references a self-signed certificate from the file system. 
// ssl self signed certificate and key
var cert = fs.readFileSync(__dirname + '/../../ssl/keys/server.cert');
var key = fs.readFileSync(__dirname + '/../../ssl/keys/server.key');
I assume we should use the same certificate as the the rest of Egeria. This may involve build process copying the cert and key into the presentation server folder tree - so the existing logic could be used. 
10) I assume we would need an environment variable alternative to 9). 
11) We need to amend the express middleware passthrough of the rest calls to augment them with ssl, based on the environment variables.      

Is the above sufficient @planetf1 ? I am not sure whether this code needs to be aware of the trust store.

planetf1:
self signed certs is something I'm working on currently as we have a few challenges. 

As you say, we should probably start by using a common cert - note that currently we have them stored in pkcs12 format, and they are configured by reference to the classpath. That could change, but I would consolidate to as few variations/copies as reasonable. At least by default. 

[...]

PR #3184, opened 2020-06-10
#3135 hard code outbound secure presentation server rest calls

davidradl:

Changed the outbound rest calls to use axios library, ran npm audit fix and hard coded secure outbound calls that are not validated, that use the same certificate as the presentation server.


planetf1:
After a build, can you check if the files look ok under open-metadata-distribution/open-metadata-assemblies/target/egeria-2.0-SNAPSHOT-presentation-server/egeria-presentation-server-2.0-SNAPSHOT
You should be able to run node src/server/index.js to launch the UI server

PR #3189, opened 2020-06-11
#3135 Drive presentation server configuration from environment variables.

davidradl:


davidradl:
@grahamwallis  fyi this introduces a new way to run the presentation server. I have updated the readme to say what to do for developers.

grahamwallis:
Thanks @davidradl - one thought about naming of the env variable - I wonder whether it would be better to rename this (whilst we are still in development mode) to EGERIA_PRESENTATION_SERVER_CONFIG? That way, if we discover additional configuration that we want to add to the environment file/object (in addition to the parameters for the remote server) - then we could add them and the variable name would still be relevant.

davidradl:
> Thanks @davidradl - one thought about naming of the env variable - I wonder whether it would be better to rename this (whilst we are still in development mode) to EGERIA_PRESENTATION_SERVER_CONFIG? That way, if we discover additional configuration that we want to add to the environment file/object (in addition to the parameters for the remote server) - then we could add them and the variable name would still be relevant.

Good idea I will do. that

planetf1:
Just as we're referring to certificates, it's worth noting they come into play in a few ways

server side
 - *certificate* that server offers to client to prove it's authenticity (and contains it's public key)
 - certificate of a client, or more likely the certitication authority to validate the client IF using mutual SSL

client side
 -  certificate of the server stored locally, or more likely the certification authority to validate the server
 - client's OWN certificate, if using mutual SSL, which is then sent back to the server

Within that there's a few controls to consider
 * How much certificate validation is done (most often for self-signed server certs or cas) - this can extend to hostname for example, and is something we need to bypass in a typical dev env.
 * Whether the server enforces mutual SSL

there will be more in #1403

PR #3217, opened 2020-06-18
OLS - Remove edge redirection

lcpopa:


lcpopa:
Also fixed exception in property built up

Issue #1153, opened 2019-06-20
UI sends usage statistics - possible production concern


planetf1:
The Egeria UI currently throws an error that can be seen in 'developer tools' '(index):1 Uncaught SyntaxError: The requested module '../vaadin-usage-statistics/vaadin-usage-statistics.js' does not provide an export named 'usageStatistics'' 

I noticed that the vaadin components perform useage tracking, and our UI is including https://host:port/node_modules/@vaadin/vaadin-usage-statistics/vaadin-usage-statistics-collect.js

The components themselves look ok being Apache 2.0, but the stats gathering is disconcerting. Even the vaadin site suggests that during production useage the stats are disabled - and the .js replaced with empty bodies. We at least need to ensure this happens when we deploy. The docs at https://github.com/vaadin/vaadin-usage-statistics also cover how to disable the stats gathering

It's crucial we are transparent and need to take the utmost care with any form of data gathering

Note - I cannot confirm 100% data is being captured, but this needs to be verified

  This was found when debugging #1129 
cc: @bogdan-sava @davidradl

planetf1:
Note during the build this appears
```
19:26:05 [INFO]     Vaadin collects development time usage statistics to improve this product. To opt-out, either run:
19:26:05 [INFO]     npm explore @vaadin/vaadin-usage-statistics -- npm run disable
19:26:05 [INFO]     to store disable statistics for the machine, or add
19:26:05 [INFO]     "vaadin": { "disableUsageStatistics": true }
19:26:05 [INFO]     to the project package.json and re-run npm install to disable statistics for the project.
19:26:05 [INFO] 
19:26:05 [INFO]     For more details, see https://github.com/vaadin/vaadin-usage-statistics
19:26:05 [INFO]   
```

planetf1:
This should be done before next release/any production useage IMO cc: @mstrelchuk

planetf1:
This is still being reported and could be a concern for any production use case. I would suggest they are disabled

bogdan-sava:
still need to be investigate

planetf1:
@sarbull @bogdan-sava  Is this addressed now? If so can be closed

sarbull:
yes, i saw it some days ago, it can be disabled by a `package.json` property.. `"vaadin": { "disableUsageStatistics": true }`

sarbull:
[...]

PR #3219, opened 2020-06-18
UI fix polymer build for production as maven profile

bogdan-sava:


bogdan-sava:
@planetf1, yes, make sense. 
I made es6 build default.
Regarding #1153 , I'll take a look, Somehow I missed it.

PR #3238, opened 2020-06-22
Continue enabling SSL Security (Draft)

planetf1:
This PR
 * Adds full support for SSL (1 way and 2 way) to the Spring REST Client connector
 * Disentangles the combined client/server ssl configuration used previously
 * removes http helper and moves more complex ssl configuration to spring rest client connector
 * Allows egeria client to be different to other REST usage in the same app (for user's application clients as we can't mandate what else they may connect to)
 * supports TLS v1.2 by default (and minimum), TLS v1.3 for Java 11 & above
 * Resolves configuration ENVIRONMENT > egeria properties > JDK properties - aiming to support current known usage, standard java parms, as well as egeria specific config)
 * Deprecates 'strict.ssl' spring property - though this continues to work for now, and As a last resort reads key/truststore from the spring server config (may be removed in a later PR before 2.0 ships)
 * Will report any exceptions in the spring rest client constructor as audit events

This PR does not
 * Use connector properties to configure SSL - if needed this will be added in a later PR
 * override configuration of SSL crypt algorithms
 * override the X509 certificate provider to allow specification of which identity cert a client should present in 2 way SSL
 * remove strict ssl/use of spring server config
 * use audit events to debug configuration - currently this is in a static initializer & only debug logging is implemented
 * add additional unit tests (as tests assumptions, and they are still being worked through in understanding different configurations)
* support JKS format keystore or truststore. Only PKCS#12 is supported

Subsequent update will add doc & may consider refactoring

Configuration

* Allows SSL checks to be skipped by setting verification to false (default=true) by setting
  - EGERIA_SSL_CLIENT_VERIFY(environment var)
  - egeria.ssl.client.verify (system property)

* If SSL checks are allowed, Configure the Truststore (used to validate server certs/CAs when making a client request) using:

  - EGERIA_SSL_CLIENT_TRUSTSTORE (environment variable)
  - egeria.ssl.client.truststore (system property)
  - javax.net.ssl.trustStore j(ava system property)

  And an associated password using:

  - EGERIA_SSL_CLIENT_TRUSTSTOREPASS
  - egeria.ssl.client.truststorepass
  - javax.net.ssl.trustStorePassword

* If SSL checks are allowed support mutual SSL (client & server) by configuring the keystore used to store the client
  identity certificate. Note the first valid cert is presented to the server. Not setting will mean 1 way SSL is used.

  - EGERIA_SSL_CLIENT_KEYSTORE
  - egeria.client.keystore
  - javax.net.ssl.trustStore

  And an associated password

  - EGERIA_SSL_CLIENT_KEYSTOREPASS
  - egeria.ssl.client.keystorepass
  - javax.net.ssl.keyStorePassword

For filenames, 'classpath:' can be added as a prefix to lookup in classpath. Spring supports this construct - evaluating if this should remain. In the code here the prefix is permitted however the key/trust store is referred to

**Opening for visibility - NOT READY TO MERGE YET**

Testing 
  - Happy path (only) using existing cert works fine
  - further testing underway - consider this more a 'it compiles, sharing for info stage'. LOTS MORE needed!


open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+    // These are distinct from SERVER (in the context of a socket) settings
+    // As we may want to make distinction (for mutual ssl). here is only client
+    private static final String ENV_SSL_VERIFY = "EGERIA_SSL_CLIENT_VERIFY";
+    private static final String SYSPROP_SSL_VERIFY = "egeria.ssl.client.verify";
+
+    // TODO: Rename vars
+    private static boolean sslVerify=false;
+
+    private static final String ENV_SSL_TRUSTSTORE = "EGERIA_SSL_CLIENT_TRUSTSTORE";
+    private static final String SYSPROP_SSL_SERVER_TRUSTSTORE = "server.ssl.trust-store";
```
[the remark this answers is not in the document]
planetf1: On a) I went with 'EGERIA_SSL_CLIENT' as the prefix. Had to choose a name but open to ideas. It affects all client usage of egeria - regardless of server. Perhaps should be EGERIA_CLIENT_SSL  instead?

In a later iteration I may look at other config mechanisms (including in particular connector properties), but there are so many combinations - including picking up system properties (which I think ING are using) that I wanted to get config in for a first pass

SYSPROP_ are for system properties, yes I did introduce - same comment as above on a naming strategy?  The main reason for adopting this here is that it seems very common in industry to use system properties for configuring SSL. Often they are the normal java ones, but this causes a problem if using multiple clients. Take for example a proxy between something IGC and egeria -- what if the required settings differ for each? Hence allowing overriding with an egeria specific property. 

My thought is to prove the basics all ok then we can loop through on a later iteration and refine this approach - though renames are easy to do now :-)

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+        if (StringUtils.isNotBlank(sslVerifySetting))
+        {
+            log.debug("egeria-ssl: ssl verification set from environment " + ENV_SSL_VERIFY);
+        }
+        else
+        {
+             sslVerifySetting = System.getProperty(SYSPROP_SSL_VERIFY);
+                    if (StringUtils.isNotBlank(sslVerifySetting))
+                    {
+                        log.debug("egeria-ssl: ssl verification set from system property " + SYSPROP_SSL_VERIFY);
```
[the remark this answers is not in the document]
planetf1: I'd already Acknowledged in list of what PR doesn't do (at least yet) as I'm thinking I do in second PR once worked through the config & actionable entries. Not sure I need to create an audit log entry to report config, but I do need it to support debugging for me as I develop - hence debug for now ...

open-metadata-implementation/adapters/authentication-plugins/http-helper/src/main/java/org/odpi/openmetadata/http/HttpHelper.java
```diff
@@ -1,52 +0,0 @@
-/* SPDX-License-Identifier: Apache-2.0 */
```
bogdan-sava, line 1: This needs to stay as it handles the skip client checks on JVM level as suppose to be. At least for UI.

open-metadata-implementation/adapters/authentication-plugins/http-helper/src/test/java/org/odpi/openmetadata/http/HttpHelperTest.java
```diff
@@ -1,31 +0,0 @@
-/* SPDX-License-Identifier: Apache-2.0 */
```
bogdan-sava, line 1: same here

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+        sslVerify = !StringUtils.equalsIgnoreCase(sslVerifySetting, "false");
+        log.debug("egeria-ssl: ssl verification SET to: " + sslVerify);
+
+
+        // Decide which truststore to use - once we decide we'll use the corresponding password if found
+        // We will use Environment -> Egeria Client Property -> JVM property -> server config (if found)
+        // If none found we'll remain set at null
+        // NOte that a null password is valid - so if the truststore is set we'll take the associated pass even if null
+        truststoreSetting = System.getenv(ENV_SSL_TRUSTSTORE);
+        if (StringUtils.isNotBlank(truststoreSetting)) {
```
bogdan-sava, line 129: why to have so many properties meaning the same ?

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
-         * name_0007, etc) would be sent to the server as "name_0 7".
-         */
-        DefaultUriBuilderFactory builderFactory = new DefaultUriBuilderFactory();
-        builderFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.VALUES_ONLY);
+        final String methodName = "SpringRESTClientConnector";
+        /* TODO: Disable SSL cert verification -- for now */
 
+        // Numerous opportunities for error. We'll propogate the base exception to caller, and it will appear in audit log
+        try {
+            SSLContext sc = SSLContext.getInstance(protocol);
```
bogdan-sava, line 233: This constructor is used to create multiple instances per java application, and making SSLContext per client is not efficient and useless.

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+                if (StringUtils.isNotBlank(keystore)) {
+                    InputStream ksFile;
+
+                    KeyStore kStore = null;
+                    KeyManagerFactory kmFactory = null;
+                    if (StringUtils.isNotBlank(keystore)) {
 
-        restTemplate = new RestTemplate();
+                        //TODO: Default KeyManagers chooses a cert to use ie by alpha sort -- Implement own X509KeyManager if required to choose by alias - the default
+                        kStore = KeyStore.getInstance("PKCS12");
```
bogdan-sava, line 273: why to have such a limitation?

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+             * name_0007, etc) would be sent to the server as "name_0 7".
+             */
+            DefaultUriBuilderFactory builderFactory = new DefaultUriBuilderFactory();
+            builderFactory.setEncodingMode(DefaultUriBuilderFactory.EncodingMode.VALUES_ONLY);
+            restTemplate.setUriTemplateHandler(builderFactory);
+
+            /* Ensure that the REST template always uses UTF-8 */
+            List<HttpMessageConverter<?>> converters = restTemplate.getMessageConverters();
+            converters.removeIf(httpMessageConverter -> httpMessageConverter instanceof StringHttpMessageConverter);
+            converters.add(0, new StringHttpMessageConverter(StandardCharsets.UTF_8));
```
bogdan-sava, line 338: Why to limit?

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -41,49 +49,317 @@
...
+        try {
+            SSLContext sc = SSLContext.getInstance(protocol);
 
-        /* TODO: Disable SSL cert verification -- for now */
-        HttpsURLConnection.setDefaultHostnameVerifier(bypassVerifier);
-        SSLContext sc = SSLContext.getInstance("SSL");
-        sc.init(null, INSECURE_MANAGER, null);
-        HttpsURLConnection.setDefaultSSLSocketFactory(sc.getSocketFactory());
+            TrustManager[] trustManagers = null;
+            KeyManager[] keyManagers = null;
```
bogdan-sava, line 236: TrustManager and KeyManager lists per client? I think is not a good idea

open-metadata-implementation/server-chassis/server-chassis-spring/src/main/java/org/odpi/openmetadata/serverchassis/springboot/OMAGServerPlatform.java
```diff
@@ -98,10 +97,11 @@ public static void main(String[] args) {
     public InitializingBean getInitialize()
     {
         return () -> {
+            // TODO: This section will be removed once the new implementation is complete and tested
             if (!strictSSL)
             {
-                log.warn("strict.ssl is set to false! Invalid certificates will be accepted for connection!");
-                HttpHelper.noStrictSSL();
+                log.warn("strict.ssl is *DEPRECATED*! Please use egeria.ssl.client.verify system property or environment variable. If already set they will take precedence over strict.ssl. Note also that strict.ssl=true is the same as egeria.ssl.client.verify=false, and in the new implementation only affects client interaction");
```
bogdan-sava, line 103: If you reach this point it is already set.
In the current implementation also affects only client interaction but on the JVM level which is more efficient

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/EgeriaUIPlatform.java
```diff
@@ -46,9 +45,10 @@ public static void main(String[] args) {
     public InitializingBean getInitialize()
     {
         return () -> {
-            if (!strictSSL)
-            {
-                HttpHelper.noStrictSSL();
+            // TODO: This section will be removed once the new implementation is complete and tested
+            if (!strictSSL) {
+                LOG.warn("strict.ssl is *DEPRECATED*! Please use egeria.ssl.client.verify system property or environment variable. If already set they will take precedence over strict.ssl. Note also that strict.ssl=true is the same as egeria.ssl.client.verify=false, and in the new implementation only affects client interaction");
```
bogdan-sava, line 50: If you reach this point it is already set.
In the current implementation also affects only client interaction but on the JVM level which is more efficient

bogdan-sava:
`Adds full support for SSL (1 way and 2 way) to the Spring REST Client connector`
There is no need to have 2 way SSL support. It is useless for applications that expose APIs. 

`Disentangles the combined client/server ssl configuration used previously`
It was not combined.

`removes http helper and moves more complex ssl configuration to spring rest client connector`
The purpose of it is to disable checks on JVM level. I need this to stay.

`Deprecates 'strict.ssl' spring property - though this continues to work for now, and As a last resort reads key/truststore from the spring server config (may be removed in a later PR before 2.0 ships)`
Deprecate it to introduce another one that is doing the same thing on every instance of connector. Why?

bogdan-sava:
What we need to achieve is:
**1. server side:** java application is using HTTPS only by default and certificate for it is loaded in the following order:
- from path specified in the java ops `-Djavax.net.ssl.keyStore`
- spring-boot specific property (if first option missing)
- from inside the jar where is included (if first 2 are not specified)

**2. client-side:** Here we are talking about java clients that Egeria provides. The client checks the server certificate against a trustStore. Egeria provides a trustStore that contains already last option from the server side. However, if the user wants to use the server with own certificate, so needs to configure also it's own trustStore. The way to configure the trust store is using java ops:
**javax.net.ssl.trustStore**.  Alternatively may choose on his own risk to use strict.ssl option to false in order to disable all the checks.

Both were are already done.
What was not done, was to set trustStore all the places java clients were used.

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
planetf1: pkcs12 is standard across environments, and the default from java 9 & above. JKS is java specific
For the intial implementation I simply didn't spend the time to allow jks, but this would be easy if needed. 
Do you use jks or pkcs12?

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
planetf1: This has been highlighted after some code moving around in the method - but it wasn't introduced by the PR. It came in   #1872

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
planetf1: Not so much the lists, but instances per client I'd agree -  this is indeed inefficient. 

I felt it important to avoid changing the default context to avoid interaction with rest requests for other client libraries. By ensuring we use our own context we are isolated. This is especially important for 3rd party applications that are making use of our libraries. At some point we might expect the client API and/or connector properties to support specification of the configuration, for now they are at least scoped to egeria

open-metadata-implementation/server-chassis/server-chassis-spring/src/main/java/org/odpi/openmetadata/serverchassis/springboot/OMAGServerPlatform.java
planetf1: Yes - noticed this last night.

JVM level didn't allow for the isolation (see above comment). Additionally I think it's important to try and keep our variables scoped clearly to egeria - hence the prefix 'egeria.xxxx' so that we don't get into any clashes with other applications . 'strict_ssl/strict-ssl' etc or a variant is used elsewhere. Same reason as above. Keep clarity for the egeria configuration.

I left support in in the short term to ensure things kept working

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/EgeriaUIPlatform.java
planetf1: yes as per above

planetf1:
Thanks for the feedback.

* Two-way SSL is potentially important on some of our interactions -- anytime we make a rest API call that may change state. This is probably particular relevant with OMRS.  I think it's important to have the capability to secure that interaction. With 1 way SSL the 'client' in any interaction isn't trusted, only potentially the server. By keeping the client and server roles very distinct we allow for mutual SSL to be used

* By client/server being entangled, what I mean is that in any SSL interaction there are two roles
 - Client : this initiates a request, and validates the server is trusted. It may also offer it's own certificate
 - Server : this responds to a request, provides it's cert, and optionally will validate the client's cert

But our 'server' has both of these roles -- as a client, and as a server. The certificates used may differ, as may the configuration. Any JVM params will affect both equally. By having egeria specific properties we have more control over keystore, truststore, ssl cert check skipping independently by role -- whilst also supporting the simpler case where they are the same.

On configuration, also note that other client using restTemplate will not be using spring boot. So they will need another way to control behaviour. Whilst the current jvm params do allow for some config they may conflict when mixing client libs in the same jvm. The proposed implementation addresses that. We have a lot of clients - so having that function in the connector itself seemed to make a lot of sense.

planetf1:
To clarify - my hope is that
 - Your current config will continue to work with no changes. It is one way of configuring that should work. 
 - We can discuss about the current strict.ssl setting and whether to rename/remove
 - The new configuration adds more flexibility especially for clients that keeps concerns seperated- and allows extension to use connector properties/new methods in future - indeed it is clients that were the most painful here, as spring is managing the server.
 - Mutual SSL is easily possible
 
Agree that the creation of contexts/builders needs some rework/checking.

bogdan-sava:
To secure the interaction over https we need to use authentication and authorization mechanism, not 2 way SSL over HTTP.  
2 way SSL is a proper solution for SSH not for REST over HTTPS. What are we gonna do when a new client want to consume API? Are we gonna restart the server to load the new certificate?

bogdan-sava:
nevertheless unnecessary complexity in order to use different server-side HTTP servlet container connectors for the OMRS and non OMRS. Because with normal web server you only have one, so will behave the same.

And why to have server side SSL configuration in a class that is client only such SpringRESTClientConnector? This class might be used in a totally different java application that is consuming the API.

planetf1:
2 way SSL provides transport level security. There are other ways needed too -including authentication higher up the stack. usually a combo is needed. 

Adding a new client is absolutely find if you use a certificate authority (as will be the case for our self signed OOTB setup). The 'client' role only needs to have the CA in it's truststore. No restart needed

planetf1:
The SpringRESTClientConnector does not do configuration of the 'server' as in tomcat. It does impact the egeria chassis where it keeps quite distinct separation between it's client and server role -- intentionally.

planetf1:
Agreed in status call 23 Jun to generate some pictures showing the flow & covering different use cases

planetf1:
Refer back to #1403 . I am going to revisit the implementation of this PR through a few separate issues/PR
The main intent will be to complete the docs, refine strictssl processing (configurable), but to not -- at his point -- add egeria specific config, and instead just use default behaviour to the extent possible through the server.ssl spring settings & specifically the javax.net.ssl settings for the client. Whilst not allow specific config just for egeria, or against specific servers, this will simplify hugely. INstead minimize custom code.

Issue #3144, opened 2020-06-03
Split single cohort topic into multiple


cmgrote:
@grahamwallis @mandy-chessell and I had some email-based discussion on the potential of the cohort topic being "flooded" by events in certain scenarios, which given our single-topic approach to date would mean that some events would end up queued up behind a long list of other events and could therefore face an unnecessarily significant delay before being processed.

From our discussion, we'd like to suggest splitting the current single cohort topic into at least 2 (probably 3) topics, as follows:

1. Instance events
1. TypeDef events
1. Registry events (may be combined with (2) for simplicity, though given increasing volume of (2) probably makes sense to already separate out?)

cmgrote:
Suggesting that we try to get this into v2.0, as I believe it would be a breaking change (not backwards-compatible), at least not without some overhead in various places catering for both the single-topic and multi-topic setups (?)

grahamwallis:
Agreed.
As an adjunct to this we should augment our KafkaOpenMetadataEventConsumer so that if it gets an exception from commitSync() it should at least log it - and preferably take some action on it if appropriate. Currently it seems to implicitly throw the exception to the rebalance operation (in Kafka) and I think Kafka throws another exception back in response to our poll of the topic. By logging we would at least find out why the commitSync has failed.

planetf1:
Could another option be to create additional topics for let's say 'high' priority
* In our topic connector we could always listen to both regular & high, and aim to process any on high first
* we could have a map of events to priority
* A configuration could control if we SEND any messages on the high priority queue - we would only tell the admin to do this if all cohort members are at a suitable version (ie 2.0 or above) - we might be able to do this dynamically

This would work with explicit instance/typedef/registry also

ie.. I think we could make it non breaking if we wanted?

One concern is we have a lot of suggestions for 2.0 and may have too many to get in at once?

cmgrote:
Approach discussed on Thursday call:

- Leave existing cohort topic there for backwards compatibility -- with all events
- Add 3 new topics: one for instances, one for typedefs, one for registry events
- Rework event managers to work across both sets of topics (ie. cover new topics as well as original)
- Enterprise event manager should only reconfigure the enterprise connectors when a registration event indicates something has changed

We could dynamically decide whether we need to put events onto the old topic or not, based on the version of the members of the cohort that are registering.

mandy-chessell:
Still needed

lpalashevski:
Still relevant as discussed on the last community call

[...]

Issue #2991, opened 2020-04-29
Kafka - WakeupException after temporary deactivate the services for a server 


danielaotelea:
After calling DELETE {{url-server}}/open-metadata/admin-services/users/{{user-id}}/servers/{{server-id}}/instance, I got the following exception: 

Mon Apr 27 15:51:09 EEST 2020 platform-2 Shutdown OMRS-AUDIT-0011 The Open Metadata Repository Services (OMRS) has disconnected from the open metadata repositories
Mon Apr 27 15:51:09 EEST 2020 platform-2 Shutdown OMAG-ADMIN-0006 The platform-2 server has completed shutdown
Mon Apr 27 15:51:09 EEST 2020 platform-2 Shutdown OMRS-AUDIT-0016 The listener thread for the OMRS Topic Connector for topic OMRSTopic.openmetadata.repositoryservices.cohort.cohort1.OMRSTopic has shutdown
Mon Apr 27 15:51:09 EEST 2020 platform-2 Shutdown OMRS-AUDIT-0016 The listener thread for the OMRS Topic Connector for topic platform-2.openmetadata.repositoryservices.enterprise.platform-2.OMRSTopic has shutdown
Exception in thread "Kafka-Consumer-OMRSTopic.openmetadata.repositoryservices.cohort.cohort1.OMRSTopic" org.apache.kafka.common.errors.WakeupException
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.maybeTriggerWakeup(ConsumerNetworkClient.java:511)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:275)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:233)
	at org.apache.kafka.clients.consumer.internals.ConsumerNetworkClient.poll(ConsumerNetworkClient.java:212)
	at org.apache.kafka.clients.consumer.internals.ConsumerCoordinator.commitOffsetsSync(ConsumerCoordinator.java:977)
	at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:1511)
	at org.apache.kafka.clients.consumer.KafkaConsumer.commitSync(KafkaConsumer.java:1459)
[...]

PR #3247, opened 2020-06-23
Close consumer gracefully #2991

wbittles:
This PR address both issues raised in #2991 and #3144

The First commit is the fix that means the TopicConnector now waits for both client threads to close, and removes an uncalled method "stop".
The second commit add exception handling to the calls to the consumer.commitSync().

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/asset-lineage/asset-lineage-view.js
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/asset-lineage/asset-lineage-view.js
@@ -124,7 +124,6 @@ class AssetLineageView extends mixinBehaviors([ItemViewBehavior], PolymerElement
             graphInteraction: {
                 type: Object,
                 value: {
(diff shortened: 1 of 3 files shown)


planetf1:
If we fail to update the sync point I am presuming that what will happen is that when next started (which is likely) we will end up potentially re-processing some messages. Most of the time we sync quickly in any case, and we should be able to handle a duplicate message, so I don't think this adds undue concern. If we are concerned we probably need to unwrap the next level of analysis and testing (beyond the high level architecture) in terms of everything around transactionality...

grahamwallis:
@planetf1 - Agreed - the fact that this is consumer side, as you say, means we would duplicate not drop a message.

planetf1:
I think we're good here then and should merge once the build is up to date...
(Note that it will be going into 2.1)

wbittles:
@grahamwallis  and @planetf1 In this case I believe the root cause is that there isn't actually a message consumed, we've opened a transacted poll, no message has arrived and then we issued a WakeUP and tried to commit an empty transaction. I don't think there is any underlying problem , (except that the consumer doesn't track if there is anything to actually commit , before calling commitSync )

Issue #3236, opened 2020-06-22
Add customisation to search results in glossary author


davidradl:
This will add a multi select dropdown to allow the user to customise what fields to be returned on a search.

PR #3250, opened 2020-06-24
#3236 add filtering to search results

davidradl:


davidradl:
I also added more standard users to the presentation server in this pr.

PR #3252, opened 2020-06-24
UI logging sanitizing

bogdan-sava:
Signed-off-by: bogdan-sava <bogdan.sava@gmail.com>


gurshafriri:
👋  @bogdan-sava 
would you say this is concerning a security issue or just "best-practice" sort of fix?

bogdan-sava:
It is a best-practice, because by having root level on DEBUG some libraries are logging the whole request with headers and body.
This will avoid logging password in logs even debug level is set.

Issue #940, opened 2019-04-18
Add useful version info to egeria jars (and other relevant 'assemblies')


planetf1:
When debugging a recent issue with atlas & egeria it would have been helpful to be able to easily check the git commit of the code running/installed 

#931 suggests allowing the server to report this as an API call

This issue suggests we could also add it into our jars so that they can be easily inspects on the filesystem.

The pom properties file already contains a timestamp, but git commit/branch/tag etc info would be helpful.

planetf1:
still valid

planetf1:
As per https://github.com/odpi/egeria/pull/3267 @bogdan-sava comment
 - would be useful so we could remove use in code like https://github.com/odpi/egeria/pull/3267/files/a6f07963691335df42a21ef67c12eaf2c34f3170#diff-bb336b19f4b3932ebb3dc61065e48203

Specifically we'd need to add  Implementation-Version: 2.1-SNAPSHOT or similar into the manifest (if we're ok with the maven versioning)

There may be the opportunity to add other tags -- for example when building the docker images we now add the build date, last git update time, git version.

planetf1:
Closing as not planned

Issue #426, opened 2018-12-04
Automate the FVT tests


davidradl:
use a container approach similar to what is proposed for the proposed Conformance test automation

planetf1:
We should look at automating the deployment of a k8s environment based on egeria that we can run automated tests against, configured from within the build pipeline

First step:
* Identify/fund a k8s environment/cluster

planetf1:
Results of fvt test runs should be stored as some kind of build output artifact so that it can be referred to both by developers (to see how the last test run went) and for consuming organisations wanting to see the results from the current release.

planetf1:
Looking at what we have under FVT currently:
 - open types gets built and run & is now quick and reliable
 - subject area fvt can be run  ( see also #2693) in part -- the single server solution, The multi server test requires infra (see #747) but this could be satisfied by the typical coco environment
 - asset consumer fvt could be targetted at a single server

One we have a deployed environment and pipeline we can return and link these with the runtime

planetf1:
In #3245 we had some confusion over which tests to run, and in #3254 around the environment. 

It's excellent we have some FVT tests, it's clear a little time invested in running these tests automatically (ie via surefire) with the likely needed refactoring will pay off, and we can extend the model to other tests.

I think this will supercede #747 so will close that

Aiming to 2.1 timeframe.

planetf1:
Our current unit tests use the 'surefire' plugin. If a test failure occurs the build immediately aborts.
Whilst the surefire plugin could be used to execute any kind of test, it's not well suited to having pre and post environment setup steps prior to the actual test. For example launching the egeria chassis or tearing it down.

Failsafe (an anagram of surefire) allows this, and the maven faq positions failsafe as more appropriate for integration testing. 
 https://maven.apache.org/surefire/maven-failsafe-plugin/
https://www.baeldung.com/maven-failsafe-plugin
https://stackoverflow.com/questions/28986005/what-is-the-difference-between-the-maven-surefire-and-maven-failsafe-plugins


It can still run junit, testng, pojo etc tests , and just as with surefire, coverage data can be captured to be used with Jacoco & potentially onwards to sonar.

Therefore I am proposing we gear functional tests around failsafe. These will be tests that require some running components to execute - so are more complex and may take longer than unit tests.

Tests will not be run by default (mvn clean install) but will be run with 'mvn verify' . This can be added to the merge build, but most likely not the PR build due to time taken (we can check and add after if wanted).

failsafe's convention is that tests are named:
[...]

PR #3267, opened 2020-06-26
Release 2.0

### LINKED ISSUE #3254: 2.0 release
Create 2.0 release

- [X] ✅ Create 1.x branch (PR not needed)
- [x] ✅ Update master to **2.0-SNAPSHOT** 
- [x] ✅ Update branch from 1.x-SNAPSHOT to 1.x 
- [x] ✅ Remove release notes from branch 
- [x] ✅ update release notes & rename (inc. for 2.x)
- [x] ✅ Verify notebooks
- [x] ✅ Check UI 
- [x] ✅  Run CTS 
- [x] ✅  Run FVTs (this is only FVTs automated in the build process)
- [x] ⚠️  Check samples (only samples in the build)
- [x] ✅ Check swagger doc
- [x] ✅ Verify docker image/charts/compose
- [x] ✅ Launch release pipeline 
- [x] ✅ Approve release
- [x] Verify artifacts all on maven central (check for JCenter & other errors in release logs)
- [x] Create GitHub release
- [x] ensure any fixes ported to master

### LINKED ISSUE #3261: containers broken after transition from 2.0-SNAPSHOT to 2.0
The helm lab chart fails on the 2.0 branch (test PR not yet integrated) following the switch from 2.0-SNAPSHOT to 2.0

In the helm chart the version number is possibly not a quoted string in all places, leading to some auto type conversion, which, with just '2.0' is getting converted to the integer 2, and hence the wrong image being picked.

This didn't occur when -SNAPSHOT was part of the name......

However 1.8 seemed ok....  tbc..

For example
```
 Normal   Scheduled  <unknown>            default-scheduler       Successfully assigned default/lab-odpi-egeria-lab-ui-d9dd97d78-66cnm to 10.72.127.182
  Normal   Pulling    78s (x4 over 2m58s)  kubelet, 10.72.127.182  Pulling image "docker.io/planetf1/egeria:2"
  Warning  Failed     77s (x4 over 2m56s)  kubelet, 10.72.127.182  Failed to pull image "docker.io/planetf1/egeria:2": rpc error: code = Unknown desc = Error reading manifest 2 in docker.io/planetf1/egeria: manifest unknown: manifest unknown
  Warning  Failed     77s (x4 over 2m56s)  kubelet, 10.72.127.182  Error: ErrImagePull
  Normal   BackOff    53s (x6 over 2m56s)  kubelet, 10.72.127.182  Back-off pulling image "docker.io/planetf1/egeria:2"
  Warning  Failed     39s (x7 over 2m56s)  kubelet, 10.72.127.182  Error: ImagePullBackOff
```
Relates to: #3254

planetf1:
Final Release changes for 2.0
 - replaces release notes with master
 - updates version to 2.0
 - Fixes helm charts (bad version for docker image)
 - ensures FVTs run

Replaces #3256 

Partially fixes #3254
Fixes #3261

### DIFF
--- a/open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/FVTConstants.java
+++ b/open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/FVTConstants.java
@@ -8,7 +8,7 @@ package org.odpi.openmetadata.accessservices.subjectarea.fvt;
 public class FVTConstants
 {
     public static final String USERID = "erinoverview";
-    public static final String DEFAULT_URL = "https://localhost:9443";
+    public static final String DEFAULT_URL = "https://localhost:9444";
     public static final String SERVER_NAME1  = "cocoMDS4";
     public static final String SERVER_NAME2  = "cocoMDS1";
 }


planetf1:
This is now ready to be merged to create the 2.0 release

The issue has more details on progress and outstanding tasks (including updating the 2.0 release notes in master)

open-metadata-implementation/platform-services/platform-services-server/src/main/java/org/odpi/openmetadata/platformservices/server/OMAGServerPlatformOriginServices.java
```diff
@@ -8,7 +8,7 @@
  */
 public class OMAGServerPlatformOriginServices
 {
-    final String   implementationOrigin = "ODPi Egeria OMAG Server Platform (version 2.0-SNAPSHOT)\n";
+    final String   implementationOrigin = "ODPi Egeria OMAG Server Platform (version 2.0)\n";
```
bogdan-sava, line 11: Don't like this hardcoding in the java file.
Maybe use in the future getClass().getPackage().getImplementationVersion(), of course we need to have the version in the package also something like maven-jar-plugin need to be used.

However, this is not a stopper for release.

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/FVTConstants.java
```diff
@@ -8,7 +8,7 @@
 public class FVTConstants
 {
     public static final String USERID = "erinoverview";
-    public static final String DEFAULT_URL = "https://localhost:9443";
+    public static final String DEFAULT_URL = "https://localhost:9444";
```
davidradl, line 11: Why are we changing the port here as it then does not match the ports we set up in postman collections which set up  for cocoMDS1 and 4?

release-notes/README.md
```diff
@@ -3,113 +3,7 @@
...
-    * Project management of IT change
-    * IoT and large scale cohorts
-
-
-## Reporting issues
-
-If you discover an issue in the release you are using, we recommend
-first upgrading to the latest available release.  If this does not
-resolve the problem, please raise a new
-[git issue](https://github.com/odpi/egeria).
```
davidradl, line 112: I am wondering why the release notes are not appended to this readme, as before. Are we moving the release notes out of github?

open-metadata-implementation/platform-services/platform-services-server/src/main/java/org/odpi/openmetadata/platformservices/server/OMAGServerPlatformOriginServices.java
planetf1: Yes I'd agree with that sentiment. 

We want to report the maven /project version here, and currently we don't add it in to the manifest.
```
 Manifest-Version: 1.0
 Created-By: Maven Jar Plugin 3.2.0
 Build-Jdk-Spec: 14
```
we'd need to add Implementation-Version

I opened up #940 ages ago to look at adding this/other info -- so let's continue the discussion there & if we can think how to get rid of it in more places the better --

Indeed not specific to this release

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/FVTConstants.java
planetf1: * I agree with your point about consistency
* In our samples/tests we tend not to document the 'required' configuration to run them. The best we tend to have (IMO) is the 'coco pharma' environment which a) is coherent and b) I know is being used by others even to test clients against and c) is how I test much of the release currently .The master template I used for this was generally the coco pharma notebook diagrams.  I also accept that a test is quite different to a tutorial - the requirement is more just 'I can run it'.
* In the notebooks, both cocoMDS1 and cocoMDS4 are on the 'data lake' omag server platform. Under 'common' the default URL for the data lake platform is defined as `dataLakePlatformURL  = os.environ.get('dataLakePlatformURL', 'https://localhost:9444')` - hence why the default setting.
* So this means at least if running a comparable coco setup locally, the FVT's will 'just work' - a placeholder for further automation
* From a search I couldn't find explicit use of the coco servers/platforms in the postman collections/environment for the dojo. Geared towards a developer role it's a more manual approach. The reason for changing the test above is if it looks like coco, and smells like coco, it is coco ;-)
* In #426 I've been looking at options to automate - I plan to create a new PR (on 2.1+) for that, which will require changing the parm handling here to accept input parms (in addition) - for a single server environment, which is IMO quite different to coco pharma. Along with this we'll add docs, and this would superceded any change similar to this

release-notes/README.md
planetf1: When we release we remove all of the notes (including the main readme) from the branch - it's the same as 1.8, 1.7 etc.

We can open up an issue to discuss a better approach, but the change here is aligned with what we've done previously.

The release notes will be updated in master (this still needs to be done - contributions welcome - seperate PR though as targets master)

planetf1:
@davidradl - have responded to comments above. hope this explains the thinking. I think it's good to go, but if you disagree am open to change also :-)

PR #3044, opened 2020-05-11
initial Cognos OMAS implementation

yevgenmar:
This is initial implementation of the Analytics Modeling OMAS plugin.

1. User documentation *.md files will be added separately.
2. Event listener is left for further event processing.
 
The plugin is based on plugin information-view https://github.com/odpi/egeria/blob/master/open-metadata-implementation/access-services/information-view/README.md

PR #3271, opened 2020-06-29
validate guid rest api parameters

yevgenmar:
As requested : https://github.com/odpi/egeria/pull/3044#discussion_r423503993


planetf1:
We've recently added a gradle build (under development) which needed a small update for this component (dependencies - to mirror the POM)

I'll push the fix so that we can then merge - it doesn't touch the maven build but will get us all the green ticks :-)

PR #3295, opened 2020-07-03
UI disable usage statistics by default and silent npm install

bogdan-sava:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Egeria UI application",
   "author": "Egeria contributors",
   "license": "Apache-2.0",
-  "repository": "odpi/egrias",
+  "repository": "odpi/egria",
```
planetf1, line 7: is this a typo or intentional?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/pom.xml
```diff
@@ -358,7 +358,7 @@
                             <goal>npm</goal>
                         </goals>
                         <configuration>
-                            <arguments>install polymer-cli --prefix .</arguments>
+                            <arguments>install polymer-cli --silent --prefix .</arguments>
```
planetf1, line 361: What will this silent? I would worry if we hide any version/dependency warnings? Do you have a before/after comparison?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/package.json
bogdan-sava: typo! I fixed the only partial. Spotted the final 's' not the missing "e'

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/pom.xml
bogdan-sava: this is silent the install of the polymer-cli. Always latest version.

bogdan-sava: But for dependencies (components) you are right, shouldn't be silent. I'll push the change.

bogdan-sava: During the build, node and npm are installed locally. 
Then polymer-cli (the silent one).
Then the dependencies according to the package.json from the static folder (silent removed).
Then polymer build is called.
Then the polymer build folder is copied to the resources.

planetf1: sounds good

planetf1: will approve with update :-)

bogdan-sava: not sure what 'with update' means :))

planetf1: I just meant when I saw your new commits - I think I replied just before you pushed the latest changes. I was just waiting for them. All looks good thanks

Issue #3213, opened 2020-06-17
'building a data catalog' - rather slow at 2 mins with graph repo


planetf1:
Running through the 'building a data catalog' notebook takes around 2 minutes when using
the local graph repo running locally (MacOS) in a coco pharma configuration, which seems rather slow. We don't expect berkeley+local graph to be particularly scalable - better suited to dev, but it still seems excessive.

Not done any profiling yet.

We have 290 incoming events, 
  60 ClassifiedEntityEvent
  76 NewEntityEvent
  80 NewRelationshipEvent
  36 ReclassifiedEntityEvent
  38 UpdatedEntityEvent

this is across all servers/platforms so some parallelism, but on an individual server (thread) is at around 1 'transaction' every 2-3 seconds

Needs investigating

planetf1:
Repeated... if we start with all the servers configured, started, and stabilized/idle, then with graph around 1:50 is typical, whilst inmemory takes around 22s. This is running locally (MBP 2016) otherwise fairly idle. CPU not maxing, low disk activity

planetf1:
Taking a look with the profiler it's not clear there's any particular hotspots when specifically running this notebook 

This would probably benefit from a collective profiling session @grahamwallis @mandy-chessell and indeed a closer look at the logs/audit messages during the notebook run, and perhaps closer inspection of individual REST API calls

planetf1:
Whilst sampling (regular or async) didn't identify any delays, a full trace does seem to indicate a lot of time (in my case 34s - and the only execution was the noetbook at 1:50 total) in
```
AssetHandler.java:3801 org.odpi.openmetadata.commonservices.repositoryhandler.RepositoryHandler.getEntitiesByName(String, InstanceProperties, String, int, int, String) 34651
```
And indeed digging deeper 25s of this in the following backtrace:

```
org.springframework.web.client.RestTemplate.postForObject(String, Object, Class, Object[]) RestTemplate.java
org.odpi.openmetadata.adapters.connectors.restclients.spring.SpringRESTClientConnector.callPostRESTCall(String, Class, String, Object, Object[]) SpringRESTClientConnector.java:439
org.odpi.openmetadata.repositoryservices.clients.MetadataCollectionServicesClient.callPostRESTCall(String, Class, String, Object, Object[]) MetadataCollectionServicesClient.java:5018
[...]

PR #3297, opened 2020-07-03
Automation of Subject Area FVT

planetf1:
This PR adds support for automation of the Subject Area FVT
 - Adds JUnit5 wrapper around tests, failsafe to orchestrate, launches chassis before/after. Single server
 - Intent is to demonstrate the principle  to allow component owners to extend tests as required

Dependencies
 - JUnit5 for test definition
 - failsafe plugin for test orchestration
 - process-exec-maven-plugin to launch/kill server chassis
 - groovy-maven-plugin to provide configuration for server
 - SA FVTs are now dependent on the full egeria assembly being built

Caveats/Todos - 
 - Whilst the Subject Area Tests are launched, and we check no exceptions occur, the actual results are not fully checked ..! This requires potential changes in the existing FVT code (question: this PR, or perhaps in a new PR, so we can get the base pattern more shared) - which can build on top of the basic framework
 - The groovy configuration script is an 'executable set of notes' - could do with some refactoring for clarity. Once the Admin Client API is sufficient (to launch an instance) this can be used instead of REST. When the security work is complete, or we use a client,  the SSL workaround may be removed. The error checking/reporting is rudimentary. More parameterization may help for other tests. The result may be a trivial groovy script (which is what it's useful for) and could even be inlined in maven pom if desired
 - EffectiveDatesFVT is not run currently as the existing test structure is harder to add a simple JUnit5 test wrapper to
 - Could consider full refactoring of these tests in future to be JUnit5 based - easier to report/identify behaviours
 - we could capture code coverage
 - Add documentation to developer guide, after the next set are integrated (to ensure we have a repeatable pattern)
 - 2-server tests are not included. These may belong in more complete scenario-based testing
 - not yet tested on windows/linux/in build environment
 - build scripts need checking to tune defaults for PR vs merge vs sonar etc
 - there may be cases where if the test is aborted/maven killed, the child chassis process doesn't get killed, though I've not seen this in practice yet
 - though the environment for launching the chassis is parameterized, the actual JUnit5 tests are not. We could inject server name, baseURL as needed. via properties is probably the easiest for failsafe
 - A variety of TODO: are left in place

Despite ALL of these caveats, this moves us a step forward, so if we are happy nothing here is a backward step, or completely the wrong approach, then merging  would make it easier to experiment with this approach in other FVTs and evolve to a better solution.

Recommendation for other FVTs
 - Use JUnit5
 - follow pattern in the subject area POM
 - keep tests in the test folders/phases, but in an integration specific project
 - Refactoring to add more to higher level POMs can follow - all the logic currently is in the subject area FVT pom

See the issue #426 for further discussion about some of the approaches tried to get here.


planetf1:
Interestingly I tested this before creating the PR -- and the tests ran successfully.

After merging from master, the tests then failed -- rather than a broken test I suspect there has been a regression in the subject area code.

I will post the log :

planetf1:
Note, in #3254b (release 2.0) under pr #3267 a small fix was made to the FVTs to allow them to run against coco Pharma.
this hasn't been added into 2.1 since I believe this PR for automated FVT effectively supercedes that way of working.

immmus:
Hi @planetf1 I freshly running tests from master. All test is passed.
Screenshot 2020-07-03 at 18 38 17
 I can connect to the analysis the current task only next week.

planetf1:
Thanks. It will be good to get them running in the build so we can be confident in future. 

The exceptions are both count assertions... more found than expected

I will also try rebuilding the fix or rebase/ff rather than merge on the update from master in the commit I mentioned just in case I hit any merge oddities

immmus:
I also plan to write junit for the Subject Area client, which are used while working in fvt. Thus, I think you can cover all the necessary cases. And then we will no longer need a large part of the integration fvt. What do you think about this?

immmus:
@planetf1 I also was ran tests from your branch. Result can be seen under. 
Screenshot 2020-07-03 at 19 18 15
Why did not pass in build worth understanding...

planetf1:
And specifically why it worked before the master merge... 

If you have the branch you could run

mvn clean verify

(Or install)

Will be interesting to see the build tests fail as well as the standalone fvtrun

It maybe running different tests 

Not at Mac to check currently

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/test/script/configureStartServer.groovy
```diff
@@ -0,0 +1,113 @@
...
+    {
+        // TODO: look at whether some exceptions should be deemed irrecoverable rather than retry
+        i--;
+        Thread.sleep(1000 * delay);
+    }
+}
+
+// --- Configure the platform - any errors here and we exit
+System.out.println("=== Configuring server: " + server + " ===");
+post1 = new URL(baseURL + "/open-metadata/admin-services/users/" + user + "/servers/" + server + "/local-repository/mode/in-memory-repository" ).openConnection()
```
davidradl, line 80: @planetf1 We do have junits that inline the in memory repo to run tests, which means we do not need to have a dependant server. I like being able to run against a real server as you can change the type of repo.  I wonder if we could run the FVTs against an in memory repo and then a graph repo (in case there are quirky differences, for example around search). I raise this as a consideration in case there is something in the current structure you might want to change to accommodate running it against multiple repositories.

planetf1: I did wonder about whether we should run the FVTs twice - once for each repo type (that we ship).

Can you raise a note in the parent issue for that (or a new issue)- as it's just a minor change it will be easy as a new PR? It also means you have a base to work with as you add more tests, or adapt the tests to be more JUnit ish?

As to how that scales.. I think automation is fairly easy, but optimizing performance won't be with all the starts/stops of platforms/servers -- but if we don't do this we may contaminate test results in any case, so probably worth taking the hit

I've also seen v.slow perf recently esp. in graph. I have #3213 but not looked at yet. That is probably a precursor.

planetf1:
Test run after updating from latest master (more SA changes) [ still fails ] -> https://gist.github.com/d4cd82fbfaaf138eabb098908d115f81

planetf1:
@davidradl @immmus Any idea on the failures?

I've checked server name, URL, userid -- changing these makes no difference as expected. I've also inspected my code and changed the order (in case the test was dependent on ordering). 

In all cases I consistently get a failure which seems to have followed the large update to subject area, YET when I run the tests directly (old way) I don't see a failure.

The failures all relate to getting too many results back ie:
```
org.opentest4j.AssertionFailedError: Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
Caused by: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
org.opentest4j.AssertionFailedError: Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
Caused by: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
[ERROR]   CategoryIT.testCategory:13 Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
[ERROR]   TermIT.testTerm:14 Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
```

Looking at the errors, I wondered if the issue could be delete/purge.

I randomized the tests and there were varying failures 

I then added Thread.sleep(2500) into EACH test case to space out the tests. 

ZERO ERRORS.

This could just be subtle timing issue OR it could be a regression in master at the end of last week -- since the tests failed then. We have seen issues in this area before with the notebooks.

It may be the FVTs may be more selective. I'll take a look at the notebooks/conformance test too.

davidradl:
> @davidradl @immmus Any idea on the failures?
> 
> I've checked server name, URL, userid -- changing these makes no difference as expected. I've also inspected my code and changed the order (in case the test was dependent on ordering).
> 
> In all cases I consistently get a failure which seems to have followed the large update to subject area, YET when I run the tests directly (old way) I don't see a failure.
> 
> The failures all relate to getting too many results back ie:
> 
> ```
> org.opentest4j.AssertionFailedError: Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
> Caused by: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
> org.opentest4j.AssertionFailedError: Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
> Caused by: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
> [ERROR]   CategoryIT.testCategory:13 Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 51
> [ERROR]   TermIT.testTerm:14 Unexpected exception thrown: org.odpi.openmetadata.accessservices.subjectarea.fvt.SubjectAreaFVTCheckedException: ERROR: Expected 6 back on the find got 12
> ```
> 
> Looking at the errors, I wondered if the issue could be delete/purge.
> 
> I randomized the tests and there were varying failures

@planetf1 I could be wrong, but I guess it will only work if all the tests are run in order. I suspect we need to amend each of the tests so they can run in isolation by checking how many results are all ready there, then look for additions and deletions due to the test - rather than assuming an initial content. I suggest we do not run the tests in parallel - or we may need to rethink how we check result numbers.

planetf1:
There may be a regression here.

Generally each test (category, term etc) seems to be sufficiently robust to run in different orders.

If we go back to d7cc1465273de6580eb49881bd49ad555e271457 (my base for developing the FVT automation), the tests CONSISTENTLY run & pass. I can even add `<runOrder>random</runOrder>` to the failsafe plugin to mix things up, and so far they continue to consistently work. Time after time ...  I cannot prompt any failure

Once d7cc1465273de6580eb49881bd49ad555e271457 was merged in (which has a lot of subject area changes) I see issues. Of course I agree the manually run fvt works... but perhaps it's not picking up a subtle change of behaviour that has been introduced. Note that there have been ZERo changes to the pom/tests in that change, it is only the master merge.

We can see the changes at https://github.com/odpi/egeria/commit/9d17cbd9e4c35329892bf034fb5ac24e0bb42e44   
243 changed files  with 9,501 additions and 25,419 deletions.

davidradl:
> There may be a regression here.
> 
> Generally each test (category, term etc) seems to be sufficiently robust to run in different orders.
> 
> If we go back to [d7cc146](https://github.com/odpi/egeria/commit/d7cc1465273de6580eb49881bd49ad555e271457) (my base for developing the FVT automation), the tests CONSISTENTLY run & pass. I can even add `<runOrder>random</runOrder>` to the failsafe plugin to mix things up, and so far they continue to consistently work. Time after time ... I cannot prompt any failure
> 
> Once master was merged in (which has a lot of subject area changes) I see issues. Of course I agree the manually run fvt works... but perhaps it's not picking up a subtle change of behaviour that has been introduced. Note that there have been ZERo changes to the pom/tests in that change, it is only the master merge.
> 
> We can see the changes at [9d17cbd](https://github.com/odpi/egeria/commit/9d17cbd9e4c35329892bf034fb5ac24e0bb42e44)
> 243 changed files with 9,501 additions and 25,419 deletions.

@planetf1 Ok so the big refactor may have introduced a regression where by the tests can no longer be run out of order and rely on prior state in some way. For example TermFVT line 149 creates some terms one of which is called zzz, we then do a search for zzz and expect 1. We did not check that there was no zzz terms prior to the find. Some of the FVTs leave stuff around and some issue deletes. We could change the FVTs to fully delete and purge everything they create.

planetf1:
It does look like an ordering issue, however 

I created a new test class, where I run the tests in the same order as in the RunAllFVTNow class, but still get the same issue.

Hence why it may *also* be related to timing around delete/purge, or could be some other regression.

My proposal is to:
 * Leave the one-test-class per fvt class (better for refactoring in future)
 *  disable the tests that fail 
 * I can Optionally add my 'AllIIT.java https://gist.github.com/ab7cb264d5530e34825ef5993b4a4557 if it's helpful - but will assume not for now
* Raise issue(s) to correct the SA FVT ordering
* Suggest we merge

The main intent of this work was to prove a mechanism for running FVTs. There are other tests that we can add (@mandy-chessell ) to prove the pattern & you can look through the test failures at your leisure.

davidradl:
> It does look like an ordering issue, however
> 
> I created a new test class, where I run the tests in the same order as in the RunAllFVTNow class, but still get the same issue.
> 
> Hence why it may _also_ be related to timing around delete/purge, or could be some other regression.
> 
> My proposal is to:
> 
> * Leave the one-test-class per fvt class (better for refactoring in future)
> * disable the tests that fail
> * I can Optionally add my 'AllIIT.java https://gist.github.com/ab7cb264d5530e34825ef5993b4a4557 if it's helpful - but will assume not for now
> * Raise issue(s) to correct the SA FVT ordering
> * Suggest we merge
> 
> The main intent of this work was to prove a mechanism for running FVTs. There are other tests that we can add (@mandy-chessell ) to prove the pattern & you can look through the test failures at your leisure.

@planetf1 I am ok for you to merge as-is, but subsequently I would like to have some documentation on  how to run it and any caveats as a base line on which I could refactor the FVT code.

planetf1:
I'm going to pick on another area with Mandy first I think so we can establish a good pattern. 

I'd like to temporarily draw a line on the SA fvt whichprobably weren't ideal as they have had a lot of effort put into them based on a different model, and also there is refactoring ongoing of the code itself. A number of things were learnt, and I think refactoring of SA fvt may be better once we've got a little more experience with another suite of tests (and docs/generalization/running on both repos will follow from this)

To run the tests 'maven clean install' will do this already (or maven clean verify)

immmus:
@planetf1 Check please my comment's in the -> https://github.com/odpi/egeria/issues/3302

planetf1:
> @planetf1 Check please my comment's in the -> #3302

Thanks. That's very helpful. Will revisit once completed my current change.

planetf1:
@immmus @davidradl I reviewed the comments in #3302 . There are still ongoing discussions about behaviour, and test interaction. 

My intent was to provide a working example to get us started with FVTs. I'm therefore going to to suggest that we merge this PR once the next merge update (I will check it again with master) then the both of you can refactor as you see fit, as component experts, in #3302 based on what's been learnt. I do think longer term more fine grained JUNits are probably the way to do - but even with a single AllFVT method we still get the value of running tests as part of a normal build.

When I looked through the tests I couldn't see evidence of tests running in parallel. I reviewed the docs for the failsafe plugin and whilst it has a lot of capability in this area it's not the default. We just fork a single jvm and run a single thread as best as I can see - though I may have missed something!

planetf1:
I merged from master

The working tests seem to be reliable
 Different errors (counts) are introduced in the other tests depending on sequencing. The tests need refactoring to fix this
 
I'm going to merge once built

immmus:
> @immmus @davidradl I reviewed the comments in #3302 . There are still ongoing discussions about behaviour, and test interaction.
> 
> My intent was to provide a working example to get us started with FVTs. I'm therefore going to to suggest that we merge this PR once the next merge update (I will check it again with master) then the both of you can refactor as you see fit, as component experts, in #3302 based on what's been learnt. I do think longer term more fine grained JUNits are probably the way to do - but even with a single AllFVT method we still get the value of running tests as part of a normal build.
> 
> When I looked through the tests I couldn't see evidence of tests running in parallel. I reviewed the docs for the failsafe plugin and whilst it has a lot of capability in this area it's not the default. We just fork a single jvm and run a single thread as best as I can see - though I may have missed something!

I think I can write juints for the new subject-area-client that is used by fvt, thereby we will cover the fvt part with regular juint

immmus:
> > @immmus @davidradl I reviewed the comments in #3302 . There are still ongoing discussions about behaviour, and test interaction.
> > My intent was to provide a working example to get us started with FVTs. I'm therefore going to to suggest that we merge this PR once the next merge update (I will check it again with master) then the both of you can refactor as you see fit, as component experts, in #3302 based on what's been learnt. I do think longer term more fine grained JUNits are probably the way to do - but even with a single AllFVT method we still get the value of running tests as part of a normal build.
> > When I looked through the tests I couldn't see evidence of tests running in parallel. I reviewed the docs for the failsafe plugin and whilst it has a lot of capability in this area it's not the default. We just fork a single jvm and run a single thread as best as I can see - though I may have missed something!
> 
> I think I can write juints for the new subject-area-client that is used by fvt, thereby we will cover the fvt part with regular juint

We can also write integration tests using spring tests.

planetf1:
I think the pattern here will work for a variety of areas - we hit a flaw in the subject area tests, but the same approach will work for other areas we want to add FVT to.

One of the objectives to the approach taken was to try and run the egeria server chassis in as close to a real environment as possible.

For this reason
 - We run the application out of the assembly - since that is what we expect people to use to run the server
 - We are gradually removing hard dependencies (ie for connectors/plugins) and replacing them with presence in LOADER_PATH 
 - We are trying to limit the use of Spring, which is why we have -spring modules server-side. And we do still use spring's rest template in the clients. In future we might hope to have, for example, microprofile server-side too - indeed we know of one project that are taking egeria and doing something similar. That being said I agree the current chassis is spring dependent.
 - I do think it's good to try different ideas - I wonder if the spring based testing will expose too much in the way of server/spring implementation details (from a quick glance at the docs) and be less representative as a test of what we ship. But then there's always scope for intepretation of 'fvt' - I'm thinking of them more as integration tests

I decided to merge what we have now anyway so we don't spin and make a step forward, and hope we can try the approach covered here with some other components @mandy-chessell

planetf1:
Confirming that the MERGE build (which gets triggered AFTER a PR gets merged) works perfectly, including executing the FVTs (as they are here). Starting a server, running tests, stopping. That's great news.

PR #3313, opened 2020-07-08
Enable FVT tests for PR build

planetf1:

Now that we have more FVTs in the build, enable execution as part of a PR verification.

This will extend the build time a little (1-1.5 minutes currently) but will maintain code quality.
As more FVTs get added we can review, and adapt as needed by defining short/long tests, or trying to optimize.
It will stop us adding bad tests, or break code that is being tested.
Also it will stop us accidentally breaking the merge build which drives docker images and snapshot publishing.

The 'skipFVT' option was originally added when the types test took many many minutes (5-8) but the underlaying issue there was fixed a year or so ago, whilst the latest tests including server start/shutdown take only around 40s

* We can undo this at any time
* We may wish to look again at enabling some kind of auto-merge so that we only need to tag a build and at some point it will be merged. This was troublesome last time around and may require writing a github action, but should be possible if needed.

CC: @mandy-chessell

### DIFF
--- a/.azure-pipelines/verify.yml
+++ b/.azure-pipelines/verify.yml
@@ -38,7 +38,6 @@ stages:
                   --batch-mode
                   -e
                   -s .m2/settings.xml
-                  -DskipFVT
                   -Dstream=$(System.PullRequest.TargetBranch)
                   -DaltDeploymentRepository=staging::default::file:$(Agent.BuildDirectory)/local-repo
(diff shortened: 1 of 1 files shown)


planetf1:
Note : this change is intentionally made only for master. It will not affect current release or feature branches unless the relevant master fixes are merged in to those branches. (likely for features, unlikely for release)

Issue #3312, opened 2020-07-08
Improve Subject Area FVTs


davidradl:
Improve Subject Area FVTs so that each FVT deletes everything it creates. So one FVT (e.g.glossary one)  cannot effect another (e.g. the Term one) .

planetf1:
Note also #3304 reported this issue - prob want to close either as a dup

Issue #3304, opened 2020-07-06
Fix automated FVTs for Subject Area


planetf1:
In #3297 a PR was developed that automated the subject area tests.

However 9d17cbd9e4c35329892bf034fb5ac24e0bb42e44 appears to have broken some tests when run under the failsafe/JUnit framework . In the PR above two tests were disabled - Category & Term. This needs further investigation and resolution. See the PR for a more extensive discussion on the issues noted.

In addition the FVTs could be tidied up
 * Use of hardcoded strings duplicated in the test classes should be replaced by the use of java properties from failsafe. A subsequent FVT set will likely adopt this approach
 * The tests could be refactored to be more JUnit centric - finer grained tests

planetf1:
Note also EffectiveDates needs FVT adding (none was added as it's a slightly different pattern to the other tests )

planetf1:
See also #3302

davidradl:
@planetf1 are we ok to close this one ? Or do you want to leave it open to track the other potential improvements we could make.

planetf1:
I think we're good to close this one.

#426 is open with some general notes on fvt testing (which might improve cross-component -- for example ensuring we capture jacoco coverage data)- we'll learn more implementing on another component. So I'll handle it via that issue

Issue #3302, opened 2020-07-06
Subject Area API subtle change in delete processing after big refactor


davidradl:
Prior to the big refactor a delete of a term in the FVT was interpreted as a a soft delete, the response contained the deleted Term (and I assume the other nodes)  
After the refactor the delete term does not contain the deleted term. 

The API needs to be re instated and the FVTs amended to test this, the glossary author view service needs amending as well

davidradl:
@immmus is this something you want to look at?

immmus:
@davidradl Hi, after refactoring, the removal is still soft. There is no remote term in the answer just for the reason that the remove method in RepositoryHadler is void. But the deleteEntity method from OMRSMetadataCollection is still used. And before they sent it, I scratched it all out, the method did not change in any way.

immmus:
@planetf1 @davidradl  I've just ran test and we are can see the results. All test's without errors.
<img width="1341" alt="Screenshot 2020-07-07 at 11 17 21" src="https://user-images.githubusercontent.com/35165540/86746222-94340380-c043-11ea-90c5-08ebf73bfb5d.png">

https://gist.github.com/immmus/e87b80b697e4208a9aa9779fecf8260b

And can see soft deleted.
<img width="1198" alt="Screenshot 2020-07-07 at 10 43 25" src="https://user-images.githubusercontent.com/35165540/86741687-3fdb5480-c040-11ea-9e7d-a46cd1ca8625.png">

The problem most likely is that by running tests from different methods, they run in parallel and therefore we always get different search results.

immmus:
@planetf1 @davidradl 
In this form, tests pass in parallel.

<img width="1271" alt="Screenshot 2020-07-07 at 11 50 09" src="https://user-images.githubusercontent.com/35165540/86752799-4a015100-c048-11ea-9e02-eb76fd6dd383.png">

It's results:
https://gist.github.com/immmus/fbfc75da6a5bfd44a8c53a977ab506f9

In the future, there is an intersection of the logic of tests! As can be seen on the next slide.
<img width="1044" alt="Screenshot 2020-07-07 at 11 53 34" src="https://user-images.githubusercontent.com/35165540/86753218-977dbe00-c048-11ea-8a5c-c15cfb4e4a87.png">

So friends, let's objectively look at the situation and not come up with theories about some kind of regressions. In this case, the problem is that the logic of the tests is not isolated from each other and when they are run in parallel, they simply break.

immmus:
[...]

PR #3314, opened 2020-07-08
# Resolve fvt problems

immmus:
@planetf1 Hi, having same read a documentation of junit 5 I found how disable parallel and enable order. Now all test works and passed.

https://gist.github.com/immmus/585ba8fd9ff6ab9013d695e3d0b5c8c5

### DIFF
--- a/.azure-pipelines/verify.yml
+++ b/.azure-pipelines/verify.yml
@@ -38,7 +38,6 @@ stages:
                   --batch-mode
                   -e
                   -s .m2/settings.xml
-                  -DskipFVT
                   -Dstream=$(System.PullRequest.TargetBranch)
                   -DaltDeploymentRepository=staging::default::file:$(Agent.BuildDirectory)/local-repo
                   -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn


open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/test/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/junit/AllIT.java
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.jupiter.api.Order;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestMethodOrder;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+import org.odpi.openmetadata.accessservices.subjectarea.fvt.*;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
+
+@Execution(ExecutionMode.SAME_THREAD)
```
planetf1, line 16: Thanks - I missed this when looking. Though I know @davidradl is looking at fixing the underlying issue, so will let him decide whether you want to go with this fix or not 

It's good to know how to get them working, so thanks for sharing the annotation.

I think the other aspect is which way @davidradl thinks makes sense to structure the subject area FVTs in future. I was envisioning that perhaps you would gradually move some of the tests into the JUnit classes themselves, which is why I created one test class in parallel with the FVT classes. Of course you can't order that way so understand why you suggest the change (I tried the same but missed that extra annotation!)

It probably depends whether the other fixes are near term or not, but I think we'd all agree we don't want to depend on single thread execution and ordering to allow them to work in future

Will leave to you both to agree. I'm ok with either . See #3312 #3304 #3302  which all seem to have some overlap

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/test/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/junit/AllIT.java
```diff
@@ -0,0 +1,61 @@
...
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
planetf1, line 17: and this annotation. I misread as being default. :-)

immmus:
@planetf1 I think that this solution will look good as temporary, until all subsequent changes are made. Change no longer make any labor. We will get all work test during maven build now.

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/test/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/junit/AllIT.java
immmus: I initially thought so too. It is strange that this is not so)

planetf1:
@immmus @davidradl I think we should close this PR now. No need to merge since #3304 has fixed the underlying issues, so no need to enforce ordering. It was good to learn this tip though for the future so thanks

immmus:
Yes, I close this.

Issue #3370, opened 2020-07-22
Gradle Build


planetf1:
Ignore the 'current status' here & refer to update towards the end of this issue posted 9 June 2021

**What is Gradle?** 

* Gradle is an alternative build tool to maven, which shares many of it's characteristics but with some key differences
* Gradle is used for large projects like android, and by companies including Netflix, LinkedIn, Adobe, Pivotal (spring)
* IDEs including Visual Studio Code, IntelliJ idea, eclipse are supported as well as tools we use in egeria including snyk, dependabot, sonar
* Gradle can still consume, and produce maven artifacts - meaning no change for consumers
* More about Gradle -> https://gradle.org

**Specific pain points / Issues with Maven**
 * Maven is very focussed on building java projects. It's default project type is jar. This works well for building jars, but other languages/artifact types tend to be more problematic. In egeria we're becoming more diverse. Python scripts. notebooks, docker images, golang (for containers), multiple UIs including node. Whilst there are plugins for most of these integration tends to require much experimentation especially for multi platform. Gradle sets out to work with many languages
* Customization - maven has very strict rules around lifecycle, and customization is dependent on plugins and their customization. Ad-hoc tasks are tricky to add. For example launching a server to run tests was somewhat problematic.
 * Maven activity has slowed. The benefit is stability, but there's less active community to discuss issues with. Mailing list posts, issues rely on very small number of contributors. Documentation is tough to work through, and much of it is not current
* As maven has evolved 'hacks' have been added to many plugins to support multi-module builds. With 320 modules and growing this makes some aggregate tasks very tricky to handle including java doc, site reporting 
* The maven assembly plugin seems to not be able to layout the file structure we want - ie with server, server/lib, clients (inc. dependencies) - for example in maven trying to pull in dependencies with a moduleSet actually pulls in dependencies for ALL modulesets, making it tricky to get, for example, client only libraries laid out in the right place. - this was the 'final straw' that led me to investigate an alternative, though tasks prior have taken longer than one would hope, and there are further changes whilst possibly with maven are probably non trivial
[...]

PR #3371, opened 2020-07-22
Intial gradle build (prototype)

planetf1:
See #3370 

* Adds reference to gradle in .gitignore, rat rules, codeowners, developer guide
* Adds initial gradle configuration to build most components - see issue for areas still needing to work

The maven build is not affected as different build scripts are used. (and the rat rules/gitignore etc account for these)

Can be run via `./gradlew build` from top level (or 'gradle build' if you want to run local gradle install)

### DIFF
--- a/developer-resources/License-Example-Files/License_for_POM_Files.xml
+++ b/developer-resources/License-Example-Files/License_for_POM_Files.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>2.1-SNAPSHOT</version>
+        <version>2.2-SNAPSHOT</version>
     </parent>
 
     <name>Add module name here</name>
--- a/developer-resources/Release-Process.md
+++ b/developer-resources/Release-Process.md
@@ -64,7 +64,7 @@ are GPG signed and distributed to [Maven Central](https://oss.sonatype.org).
 #### 4. Update master version & cleanup notes
 * `git checkout master`
 * `git pull upstream master`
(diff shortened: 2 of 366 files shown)


planetf1:
To give an idea of the work expended so far -- about one day effort for initial build, a further one day for a few refinements around dependency management, getting server chassis to work, checking in/docs

planetf1:
This is very early days - we may decided to move at some point, but only with comparable function, after review. The intent here was to seriously start that investigation and have open discussion.

Also note that in any case gradle can consume AND PROVIDE maven pom projects ie on maven central or locally, so it is interoperable.

CDaRip2U:
We have been using gradle for Java/other projects for a while.  the only issues i have ever had have been addressed in later gradle updates/fixes these were all related to ~/.gradle/cache issues in local develop use cases.  We have almost never seen issues with Jenkins build nodes or other production build systems with gradle usage.    build.gradle/gradle.properties allows for mostly clean management.  For example here is a snippet from our build.gradle for our egeria connector:

```
dependencies {
    compile 'org.odpi.egeria:open-connector-framework:1.+'
    compile 'org.odpi.egeria:repository-services-connectors:1.+'
    compile 'org.odpi.egeria:server-chassis-spring:1.+'
    extraLibs 'org.slf4j:slf4j-api'
    // https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient
    extraLibs group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.12'
    extraLibs 'com.google.code.gson:gson:2.8.6'
    configurations.compile.extendsFrom(configurations.extraLibs)

    testCompile(
            'org.spockframework:spock-core:1.1-groovy-2.4-rc-2',
            'org.odpi.egeria:open-connector-framework:1.+',
            'org.odpi.egeria:repository-services-connectors:1.+'
    )
}
```
This allows us to pickup minor changes with each build from 1.X branch.

Note: ensure when you add "gradle" script for unix/mac that it has the +x attribute and git maintains that setting.  Had issues with projects created on windows not setting that and we end up adding/updating that after the fact.

Hope my opinions help.  Any other questions, please let me know.

PR #3382, opened 2020-07-24
Server author

dkneisly:
@davidradl This PR adds a simple three step form to configure a new OMAG server. Still have a lot of questions for you about the different flavors of server and have ideas for improving the steps to this form depending on those answers.

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/beans/Edge.java
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/beans/Edge.java
@@ -38,4 +38,12 @@ public class Edge {
         this.label = label;
     }
 
+    @Override
+    public String toString() {
+        return "Edge{" +
+                "from='" + from + '\'' +
+                ", to='" + to + '\'' +
+                ", label='" + label + '\'' +
+                '}';
+    }
 }
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/beans/Node.java
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/beans/Node.java
@@ -10,7 +10,7 @@ public class Node {
     private String label;
     private String group;
     private Map<String, String> properties;
-    private Integer order = 0;
+    private Integer level = 0;
 
     public Node(String id, String label) {
         this.id = id;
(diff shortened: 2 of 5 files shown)


davidradl:
@dkneisly some observations:
- It would be good to have info icons next to each field pointing into the Egeria website for the definitions
- Local URL root, I suggest defaulting to the standard non-Docker one https://localhost:9443 would be better.
- I am thinking local server type would be a dropdown. 
- E.g., does not need the last comma. In fact I am not sure I would put e.g. I have seen the label repeated here; I think that makes more sense. 
- It would be good if the wizard widget was at the top or it can scroll off the bottom 
- The current content is Just over a page - so I had to scroll. It would be better to have the basic content show in a page. It would be good that the advanced one shows in a page maybe with accordions to hide sections of content. 
- Instead of first step , second step, third step, I suggest basic , advanced, preview . The text underneath can be more detail. 
- I suggest setting the organisation name up front (prior to the wizard) so all server definitions use it. This ui would be setting up servers for an organisation. We may want to extend the env to supply this so the ui does not need to enter it. 
- If I fill in all the fields and press the button is hangs . Is that was you expect at this stage?  

If it helps you collaborate with Jacob, I can merge as is, if you would prefer to amend parts let me know.

davidradl:
Errors I see in the log: 
Screenshot 2020-07-24 at 10 17 22

davidradl:
@dkneisly I have been talking with @planetf1 another maintainer. We suggest : Platform URL root instead of local URL root. We also think we should remove local from all the fields in the UI - we don't think that word helps. We could have Server Userid and Server password. The server userid would be that of the Server not GaryGeeke.

dkneisly:
@davidradl Thanks for all the input. I agree with all of your comments and suggestions, just want to work out the details for a few and provide a few more thoughts on where I was headed.

For the _server type_:
You mention a dropdown, which isn't a bad option if all of the server configs are homogenous. I was thinking about pulling this out to be the first step of the form, and rather than a dropdown box, using [Radio Tiles](https://react.carbondesignsystem.com/?path=/story/tile--radio) for the various options. I'm assuming configuration might be slightly different depending on the server type selection which is why choosing a type first would help inform which fields to show the user next.

For _access services_:
I need to learn more about the different types, especially if any type requires additional user input. I'd like to move this selection to it's own step in the overall multi-step form and use [Multi-Select Tiles](https://react.carbondesignsystem.com/?path=/story/tile--multi-select) to allow the user to select as many as they want. We could add touches like a select all and deselect all button to help the user, but I'd like to know more about why a user would choose each and what each adds to the config file. That was my only motivation for leaving the list to just one for now. Would any of the options eventually require additional input?

Overall form flow:
To summarize the overall flow after the two changes above, we would end up with four steps:
1. Choose a Server Type
2. Basic Configuration: I would remove the organization name field and move the repo type selection and max page size fields from the existing Advanced Config section to this section as they don't require additional information. This could also be dynamic depending upon the selection in the previous step, if that's needed.
3. Add Access Services
4. Preview and Deploy: I want to print the JSON in a better format.
Alternatively, we could have a separate step to configure the repository if more advanced config is necessary since that is an additional API call behind the scenes.

Environment variables:
As you guys mentioned, I think the platform URL (eventually view server URL in phase 2) and organization name belong outside of this form. I pass them from the form here for testing convenience, but the user should be able to set a real local root URL for the new server and shouldn't have to set the orgId by this stage so I want to take that field out.

PR #3400, opened 2020-07-29
Initial gradeline pipeline for azure #3370

planetf1:

Followon work for gradle prrof of concept

- Ensures gradle wrapper checked in (standard gradle practice, required by azure)
- adds verify job for azure pipelines which just does a regular build. Artifacts not published. (this will be made optional)
- adds merge job for azure pipelines

These jobs will ensure that ongoing changes in egeria don't break the gradle prototype so we'll keep in sync. No blocking checks are done, but it will be easy to refer to azure to see when builds are broken, capture logs, assess performance, investigate artifactory options etc. 

Already tested as ok on fork (but can be tricky to get build tests done correctly in a diff environment)


planetf1:
Since these changes don't affect the maven build (!) and no mandatory checks are added, and general consensus was received over prototyping this area, I will merge once verified.

PR #3401, opened 2020-07-29
# Remove unnecessary code in EntityDetailMapper

immmus:
Noticed that EntityDetailMapper adds unnecessary line to additionalProperties.
Screenshot 2020-07-29 at 14 54 22
After remove:
Screenshot 2020-07-29 at 14 56 51


immmus:
@planetf1 @davidradl Can we include this fix in 2.1?

planetf1:
We've already branched for 2.1 . We would only normally add any critical /big impact fixes / customer regressions at this stage. Though this is a bug, it's an evolving area for subject area which is in development, so I would suggest we just fix in master, and it will go out with 2.2

What is your reason for needing it in 2.1?

cc: @davidradl

davidradl:
@planetf1 I think master and 2.2 is fine

immmus:
> @immmus @planetf1 I think master and 2.2 is fine. This API is still evolving. When we have addressed the open subject area API issues , we should think about assessing and releasing the subject Area API. At that time breaking API changes would not be possible without an API version change. As we are not a released API I do not see any urgency for this - we release every month so it will be in a release in a month or so's time.

Okay

open-metadata-implementation/access-services/subject-area/subject-area-server/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/server/mappers/entities/EntityDetailMapper.java
```diff
@@ -125,15 +125,14 @@ private void mapEntityDetailPropertiesToNode(N node, InstanceProperties omrsEnti
...
-                            String mapvalue =  primitivePropertyMapValue.getPrimitiveValue().toString();
+                            String mapvalue = primitivePropertyMapValue.getPrimitiveValue().toString();
                             actualMap.put(mapkey, mapvalue);
                         }
                         node.setAdditionalProperties(actualMap);
                     } if (!mapMapToNode(node, propertyName, instancePropertyForMap)) {
                         if (null==node.getAdditionalProperties())  {
                             node.setAdditionalProperties(new HashMap<String, String>());
                         }
-                        node.getAdditionalProperties().put(propertyName,mapPropertyValue.valueAsString());
```
davidradl, line 136: I think we are saying with this change that Additional properties are supported using a property called additionalProperties that is of type map and will be stored as a String string map. If this isthe case this change should remove the previous lines if, and the enclosing if. Also the same code needs to be removed from the enum case in this method.

immmus: @davidradl Yes, I removed these.

immmus:
@davidradl It is okay or no?

PR #3427, opened 2020-08-03
# Added the ability to create a client with userId and password.

immmus:
Signed-off-by: immmus <fonto.trg@gmail.com>


open-metadata-implementation/access-services/subject-area/subject-area-client/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/client/SubjectAreaImpl.java
```diff
@@ -42,39 +41,51 @@
...
+        this.graphAPI = subjectAreaGraph;
+        this.projectAPI = subjectAreaNode;
+        this.serverName = client.getServerName();
+        this.omasServerUrl = client.getServerPlatformURLRoot();
+    }
+
+    private static SubjectAreaRestClient initRestClient(String serverName,
+                                                        String omasServerURL,
+                                                        String userId,
+                                                        String password) throws InvalidParameterException {
```
davidradl, line 78: I am uncomfortable with this change . It seems to me that userid and password is basic authentication, I think we need a more considered view across the different forms to authentication.

immmus: @davidradl  I wrote this based on super class:
Screenshot 2020-08-03 at 22 25 18

immmus: If you have any suggestions. Please suggest)

davidradl: @planetf1 Please could you look at this change, if you think it is ok - then I will merge.

planetf1: I've not looked at our use of http auth at all -- we've recently done some work on TLS security, but that is of course transport level, whilst the http authentication is another layer above. We do have a user id in the requests we pass for authorization, but that's another level up again ... I'd be inclined to put on hold without taking a broader view of how we handle http auth across all APIs. Unfortunately I don't have the bandwidth to take a look at this in the very near term (weeks)

planetf1: Just to clarify - this isn't mean to say it was wrong to propose this -- as pointed out, the superclass refers to use of the headers, so following that pattern it's very reasonable. It's just I think the truth that we put it in place initially, didn't persue, and know we need to go back to it at some point ...

davidradl: @planetf1 I think you are saying this is ok to merge but we will need to review as this area changes ? Can you confirm?

planetf1: I meant to persue http auth - and to pick up from the code that is already in place but not exploited. I don't feel the change proposed here makes sense by itself - it's only subject area, and we really need more of that architectural level review to consider how we want to support http auth in general? Plus we would want tests and docs alongside -- similar to what we did adding the TLS support (and haven't completed yet)

davidradl: @planetf1 thanks for the confirmation. @immmus please could you close out this pr and associated issue (you could leave the issue open if you want to change the words to be more of a review/investigation of the whole area).

PR #3430, opened 2020-08-03
added filter to add request information in MDC for logging

pmarius22:

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/index.html
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/index.html
@@ -80,7 +80,8 @@
         --iron-overlay-backdrop-opacity: 0.2;
       }
       :root{
-        --egeria-primary-color: #71ccdc;         /* aqua   */
+        --egeria-primary-color-rgb: 113,204,220; /* value only used for the case of background with alpha */
+        --egeria-primary-color: rgb(var(--egeria-primary-color-rgb));         /* aqua   */
         /*--egeria-primary-color: #FF6200;           !* orange *!*/
         /* --egeria-primary-color: #CC00FF; */     /* purple */
 
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/package.json
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/package.json
@@ -22,6 +22,7 @@
     "@polymer/app-storage": "^3.0.2",
     "@polymer/iron-a11y-keys": "^3.0.1",
     "@polymer/iron-ajax": "^3.0.1",
+    "@polymer/iron-fit-behavior": "^3.0.2",
     "@polymer/iron-flex-layout": "^3.0.1",
     "@polymer/iron-form": "^3.0.1",
     "@polymer/iron-icons": "^3.0.1",
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/asset-lineage/asset-lineage-view.js
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/asset-lineage/asset-lineage-view.js
@@ -136,7 +136,7 @@ class AssetLineageView extends mixinBehaviors([ItemViewBehavior], PolymerElement
                 value: {
                     hierarchical: {
                         enabled: true,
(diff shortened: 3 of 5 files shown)


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/LoggingRequestFilter.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain) throws IOException, ServletException {
+        HttpServletRequest httpRequest = (HttpServletRequest) servletRequest;
+        String requestURI = httpRequest.getRequestURI();
+        if (!path.equals(requestURI)) {
+            filterChain.doFilter(servletRequest, servletResponse);
+            return;
+        }
+        MDC.put("remoteAddress", httpRequest.getRemoteAddr());
```
planetf1, line 37: Do we have a concern here that this is potentially personally identifiable information -- the combination of host, username, session is quite specific. 

Can you elaborate on what is recorded for the logging. how this information is used and how it's control is documented?

bogdan-sava: @planetf1 the purpose is to have audit trail logging to supervise all actions from all users.
It is security related.
There is no personal information involved.

planetf1:
According to our ICO (uk) - https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/key-definitions/what-is-personal-data/ ip address is regarded as personal data, also session identifier - I don't think we explicitly log all of this elsewhere so probably need to figure out what we can log or how we mitigate in terms of defaults or documentation? Hopefully I'm being extra careful.

planetf1:
Here's one commentary article.. but we probably need to look for a little more authoritative advice (has some references to the legislation) - https://www.ctrl.blog/entry/gdpr-web-server-logs.html

Wtat drove the need for the logging? Do we have an issue collating relevant points?

Might be worth a discussion on our next call ?

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/LoggingRequestFilter.java
planetf1: (oops skipped the thread - added comments in main body)

bogdan-sava:
Hmmm...
Good to know that IP is a personal information. :)
As I can see there is a debate if dynamically IP it is or not.

However, the current PR doesn't change the logging default behavior.
The need is that this kind of systems most probably require audit.  This audit is not active by default.
It is pretty much equivalent with enable tomcat access logs.If you do it, you get same information logged for sever chassis, especially as the username is part of the request. Most probably who is do it, needs to ask for consent, but this is another story.

lpalashevski:
This change is driven form IT risk and operation compliancy requirements and ability to configure the system to produce audit log messages for the authentication process of the UI. For some organisation this is in the minimum standards.

About privacy and sensitivity, it really depends on the context. I agree ip address can be considered as private/personal identification data especially is in case the software is publicly available and accessed form addresses that are private. It depends on the environment and the purpose of the system. Each organisation will have its own policies and procedures how personal data is treated of course copying with regulations such as GRDP for example. 

I do not think what we are doing is the code wrong from privacy perspective, the fact that we are associating user access info with the auth event occurred which is quite common practice. How these events will be treated for is subject of operational compliancy and it is responsibility of the organisation operating the software. 
I think we should have it as an option (can be off by default) and if needed operator can switch it on.

Let's discuss this next week on the functionality call.

planetf1:
@lpalashevski @bogdan-sava I agree. the two main points are a) that it is off by default, so someone has to be explicit to capture this information and b) that it is the organization that deploys the stack that has to make decisions about how to use it. That does give on onus on us (the developers) to ensure we provide suitable information and defaults. 

In summary I now agree with the proposal. We don't have to document as it's non-default behaviour, but if we do, let's just be clear and transparent on what is logged (we don't have to get into gdpr just be clear on the info IMO)

PR #3386, opened 2020-07-27
Egeria feature openlineage

lpalashevski:
Open lineage feature, alpha version

PR #3432, opened 2020-08-03
required fix for #3386 (openlineage) for gradle build

planetf1:
I've added a commit which address gradle build failures (dependencies for junit5, egeria ffdc, & janusgraph inmem for UT)

See e740253

I noticed in creating this PR that the openlineage PR is a fair way behind master, which is why as I write, this PR contains many file changes and fixes. If however openlineage is rebased/merged with master, it should just be that commit and a merge commit.


lpalashevski:
Let me get open-lineage to more recent state, it is at least week old form when the PR was created and the review started. Then I will get back and merge this fix.

Issue #3423, opened 2020-07-31
Create React App for Node.js UI


jmmadsen:
Currently, the UI experience for developers is not ideal. There is no way to run the UI in "development" mode - every time a developer makes changes, the static files must be rebuilt. This is time consuming and not best practice. We are going to shift the current UI to use Create React App for their optimized webpack and developer experience. The first step in this will be to add a CRA folder next to the existing UI folder, and get them to match up.

planetf1:
Sounds good! A later step would be to ensure that is also driven from maven -- or it may be by that time we are using gradle. That will then need to be added to assembly potentially, and into a docker image. However these are further down the line and I'm also happy to help when the time is appropriate.

jmmadsen:
That makes sense to me. My plan is to very much take this iteratively, and open a new issue for each step. I think trying to do it all in one PR would be a big lift, and it would be tough for Dave Kneisly and I to sync up our code.

davidradl:
@jmmadsen 
I notice the proposal is to move to Yarn. I am not an expert on NPM vs Yarn. I found this https://www.reddit.com/r/reactjs/comments/9ypfd8/why_does_createreactapp_use_yarn/
It seems that there are people who think npm later releases is preferable. I suggest you get this working with npm as well as Yarn,  so we can compare and contrast ease of use, stability and performance before committing to one of them in an informed way. What do you think @planetf1 @grahamwallis ?

planetf1:
A few considerations

* Currently the maven build mechanism for invoking the UI build is through using the maven-frontend-plugin. This is working now for npm, but from the docs it looks as if it is fine for yarn too -- but if a switch is proposed this needs to be checked too. Specifically this plugin downloads the tool as part of the build process, ensuring a consistent configuration and not having any local install dependencies, including of platform. That will be essential.
* In future we may switch the build to gradle. I've not looked at npm vs yarn in that regard, though I suspect both are viable
* I've not used either -- I see many yarn users are very happy having switched from npm but more probably use npm. In part this may be a bit like JUnit4 and TestNG - the latter was created to address deficiencies in JUnit4 afaik, yet the next gen solution JUnit5 pretty much fixed those issues. I do wonder if perhaps yarn is the same - it fulfilled a real need at the time, but as npm has evolved it has addressed many of those deficiencies
* we currently use npm for the existing UI
* we should always be thinking about better way to do things if the tools address pain points that make us less effective. As david said perhaps providing a compare-contrast view would be useful. 
[...]

PR #3437, opened 2020-08-03
#3423 CRA foundation

jmmadsen:
Signed-off-by: JMadsen <jacob.madsen@ibm.com>


davidradl:
I checked that the existing ui still works.I will tryout the CRA function - when the readme is updated with instructions. I suggest another readme to orientate the Egeria devloper in addition to the CRA readme.

Issue #3442, opened 2020-08-05
[UI][asset-search] remove classifications column from asset search 


bogdan-sava:
The classifications are visible in asset-details-view after click on the displayName in the table

PR #3443, opened 2020-08-05
[UI][asset-search] remove classifications column from asset search

bogdan-sava:
issue #3442


planetf1:
reopening to kick off fresh build after azure network issues

PR #3462, opened 2020-08-06
Fix problem with type resolving

immmus:


open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -75,7 +75,15 @@ public SpringRESTClientConnector() throws NoSuchAlgorithmException, KeyStoreExce
         /* Ensure that the REST template always uses UTF-8 */
         List<HttpMessageConverter<?>> converters = restTemplate.getMessageConverters();
         converters.removeIf(httpMessageConverter -> httpMessageConverter instanceof StringHttpMessageConverter);
-        converters.add(0, new StringHttpMessageConverter(StandardCharsets.UTF_8));
+        converters.removeIf(httpMessageConverter -> httpMessageConverter instanceof MappingJackson2HttpMessageConverter);
```
davidradl, line 78: I am unsure of the implications of this change - as it effects all Spring clients. Please could you detail in the issue and in the comments the need for this change. Does this work if the client is talking to a back level server -if the client is using a newer way to serialise Jackson. I think this change needs wider review - I will include other

davidradl:
I am reviewing the change that @immmus has made, it works.  I would like some feedback . We have a  java generics approach as the client java API for subject area. We have an issue that when we get the relationships associated with a term, which the Subject Area OMAS calls Lines , the lines come back but do not have the child attributes. The majority of the change is as expected using @JsonSubTypes annotation. 

Two parts of the change, I would like you thoughts on @bogdan-sava @planetf1 @grahamwallis @mandy-chessell 

- There is a new org.springframework.core.ResolvableType on which he calls ResolvableType.forClassWithGenerics. I am thinkingthat this could be done using Java reflection without the new to have an increased dependancy on Spring; what do you think?  
- This pr also changes  SpringRestClientConnector to add MappingJackson2HttpMessageConverter. I am not sure or the impact / side effects of this change, as it will effect every rest spring connector. So I would like wider review of this. As per my comment @immmus  please could you detail in the issue why this change was made.

davidradl:
@immmus  please could you update the FVT to test the new child fields exist. For the child elements of Term, Glossary, Project and Category and all the specialised Lines that add extra attributes; I am happy if you want to do this in a follow on pr / issue.

immmus:
> I am reviewing the change that @immmus has made, it works. I would like some feedback . We have a java generics approach as the client java API for subject area. We have an issue that when we get the relationships associated with a term, which the Subject Area OMAS calls Lines , the lines come back but do not have the child attributes. The majority of the change is as expected using @JsonSubTypes annotation.
> 
> Two parts of the change, I would like you thoughts on @bogdan-sava @planetf1 @grahamwallis @mandy-chessell
> 
> * There is a new org.springframework.core.ResolvableType on which he calls ResolvableType.forClassWithGenerics. I am thinkingthat this could be done using Java reflection without the new to have an increased dependancy on Spring; what do you think?
> * This pr also changes  SpringRestClientConnector to add MappingJackson2HttpMessageConverter. I am not sure or the impact / side effects of this change, as it will effect every rest spring connector. So I would like wider review of this. As per my comment @immmus  please could you detail in the issue why this change was made.

MappingJackson2HttpMessageConverter is default RestTemplate deserializer so you can see on that screen:
Screenshot 2020-08-07 at 14 47 15

But it have defaultCharset is null. Because I set UTF-8 value in accordance with StringHttpMessageConverter.
It's the only thing that changes. 

ResolvableType does not carry any additional dependencies in itself. dependence _spring-core_ already added. And it is much safer to use this code, because it has long been covered with tests and was written just to work with c in such cases.

immmus:
> @immmus please could you update the FVT to test the new child fields exist. For the child elements of Term, Glossary, Project and Category and all the specialised Lines that add extra attributes; I am happy if you want to do this in a follow on pr / issue.

Yes I can will do this in next pr.

davidradl:
> > @immmus please could you update the FVT to test the new child fields exist. For the child elements of Term, Glossary, Project and Category and all the specialised Lines that add extra attributes; I am happy if you want to do this in a follow on pr / issue.
> 
> Yes I can will do this in next pr. After all, this change is critical now.

Thanks @immmus. I am just waiting hopefully for approval from other maintainers then I can merge.

davidradl:
Hi @immmus The bit I was worried about was the introduction of   MappingJackson2HttpMessageConverter into the Spring rest client - that is widely used outside of Subject Area. Please could you detail why this was introduced and what impact it might have compared to what was there before,   many thanks , David.

immmus:
> Hi @immmus The bit I was worried about was the introduction of MappingJackson2HttpMessageConverter into the Spring rest client - that is widely used outside of Subject Area. Please could you detail why this was introduced and what impact it might have compared to what was there before, many thanks , David.

I already answering for your question higher. I will attach another screenshot for understanding. 
Screenshot 2020-08-07 at 17 07 53

On this screen we can see how MappingJackson2HttpMessageConverter created from new operator
 in RestTemplate constructor I just added defaultCharset = "UTF-8";

davidradl:
@immmus Thanks for your explanations - I have asked some other maintainers to check the Spring rest client changes, as I am not very familiar with the Spring framework. Apologies for the delay in merging. fyi I am out Monday Tuesday -so it may be Wednesday when I am able to look at this again. Hopefully by then others will have cast their eyes over this and I am merge.

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
```diff
@@ -119,11 +127,7 @@ public void checkServerTrusted(X509Certificate[] certs, String authType) {
      * @param sslSession ssl ession
      * @return boolean result
      */
-    private static final HostnameVerifier bypassVerifier = new HostnameVerifier() {
-        public boolean verify(String hostname, SSLSession sslSession) {
-            return true;
-        }
-    };
+    private static final HostnameVerifier bypassVerifier = (hostname, sslSession) -> true;
```
mandy-chessell, line 130: This looks like just a style change and unnecessary.  Should follow the style of the rest of the module.

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
davidradl: @immmus Another thought - what is the implications of not have the spring rest client change. Is this required to fix the reported problem? If the Spring client changes are removed from the pr and the problem is still fixed I will merge. If this is the case, you can then submit these changes and justifications separately in another issue.

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
immmus: Okay I can return this change if u want:)

open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java
immmus: > @immmus Another thought - what is the implications of not have the spring rest client change. Is this required to fix the reported problem? If the Spring client changes are removed from the pr and the problem is still fixed I will merge. If this is the case, you can then submit these changes and justifications separately in another issue.

@davidradl Yes, this is not important change. I can rollback this tomorrow. Comment with confirmation if you want it

bogdan-sava: @immmus not entirely.
It is very good that you removed never thrown exception.
But, the reason why StringHttpMessageConverter is replaced (removed and added) with another instance using UTF-8 encoding instead of default ISO-8859-1 it is explained in the comment of the method.
Shortly, because of using regex as query param. Without it will break the regex. 
That line needs to stay.
Also cannot see the reason why you have removed MappingJackson2HttpMessageConverter instance from the list.

immmus: > @immmus not entirely.
> It is very good that you removed never thrown exception.
> But, the reason why StringHttpMessageConverter is replaced (removed and added) with another instance using UTF-8 encoding instead of default ISO-8859-1 it is explained in the comment of the method.
> Shortly, because of using regex as query param. Without it will break the regex.
> That line needs to stay.
> Also cannot see the reason why you have removed MappingJackson2HttpMessageConverter instance from the list.

I noticed that a comment was left to make sure the Rest Template always uses UTF-8. But since I knew that MappingJackson2HttpMessageConverter, since the default charset is not set, I decided to set it in accordance with this statement. I wrote about it in my comment -> https://github.com/odpi/egeria/pull/3462#issuecomment-670481314

grahamwallis:
I believe that the default character set for the MappingJackson2HttpMessageConverter is UTF-8. From the Spring javadoc "By default, this converter supports application/json and application/*+json with UTF-8 character set.". If that is correct, I doubt that the explicit specification of the default character set makes any difference. 
Since we are relying on UTF-8 it may be as well to perform the explicit setting, but I would have preferred that this change be made in a separate PR with a corresponding issue and discussion so that it could be agreed and independently rolled back if it is found to cause an issue. That would have been preferable to wrapping it up with the other Subject Area changes in this PR.

immmus:
Colleagues @grahamwallis , @davidradl , @bogdan-sava I made rollback SpringRESTClientConnector changes.

grahamwallis:
Thank you @immmus. If we find we do need the UTF-8 change then it will be good that we use a separate PR for it. 

Thanks for all your efforts.

davidradl:
Thanks @grahamwallis @mandy-chessell @bogdan-sava for your feedback.

@immmus the conclusion appears to be that we should make this pr change only code in the subject area OMAS; the impact is then only to that code area; I can then review and merge as I own this area - I confirm that you can remove the code outside of subject area omas for this PR. If you would like to make changes in the Spring rest connector, please raise a separate issue and justify the change . This keeps everything clean.

immmus:
> Please remove open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java from the pr as discussed.

I already made this)

davidradl:
> > Please remove open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java from the pr as discussed.
> 
> I already made this)

@immmus Ok I see the remaining change is to remove an unused exception from the constructor. I think this is benign, I am just updating the code and will look to merge

immmus:
> > > Please remove open-metadata-implementation/adapters/open-connectors/rest-client-connectors/spring-rest-client-connector/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/spring/SpringRESTClientConnector.java from the pr as discussed.
> > 
> > 
> > I already made this)
> 
> @immmus Ok I see the remaining change is to remove an unused exception from the constructor. I think this is benign, I am just updating the code and will look to merge

Yes I so think too:)

PR #3478, opened 2020-08-12
Ui add node type info on click

pmarius22:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/static/src/token-ajax.js
```diff
@@ -129,7 +129,8 @@ class TokenAjax extends PolymerElement {
                     composed: true,
                     detail: {greeted: "Bye!", status: status}}));
             }else{
-                console.log('Error Response:'+ resp);
+                console.log('Error Response:');
```
bogdan-sava, line 132: Let's get rid of this console log. 
Response it is present in the network tab of the develop tools in browser anyway . There is no need to print it in console.

PR #3498, opened 2020-08-18
Asset Lineage OMAS -  Vertical Lineage for Glossary Terms

danielaotelea:
1. publish the glossary context: this event contains glossary term details and the full context of the schema elements involved in a semantic assignment. Lineage classifications are included in the glossary context
 2. event renamed
 3. no need to publish the New Entity Event for GT and schema elements
 4. AL OMAS endpoint used for initial load returns the entity guid only if the entity has a context and it is published on the AL OMAS out topic


lcpopa (approved):

popa-raluca:
Please add java doc to the new public methods

popa-raluca (approved):

PR #3513, opened 2020-08-24
Vertical lineage - OMRS Events Processing

danielaotelea:
1. Update, Delete Entity, (Re/De) Classify Entity Events are send out for a Glossary Term/Glossary Category if the entity is involved in Semantic Assignment and/or Term Categorization relationships
2. Term Categorization Relationship type added for processing
3. Creation of a SemanticAssigment/TermCategorization Relationships is used to publish out the glossary term context
4. The Glossary Categories are included in the Glossary Term context

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/docker/pom.xml
+++ b/open-metadata-resources/open-metadata-deployment/docker/pom.xml
@@ -100,7 +100,7 @@
             </activation>
             <modules>
                 <module>configure</module>
-                <module>presentation-server</module>
+                <!--module>presentation-server</module-->
                 <module>egeria</module>
                 <module>jupyter</module>
             </modules>


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
```diff
@@ -209,6 +209,10 @@ private void processClassifiedEntityEvent(EntityDetail entityDetail) throws OCFC
         if (!anyLineageClassificationsLeft(entityDetail))
             return;
 
+        if (!publisher.isEntityEligibleForPublishing(entityDetail)) {
```
davidradl, line 212: Would it not be better to have a positive if and  put the subsequent 2 lines inthe body of the if. So you do not need to issue the return?

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/GlossaryContextHandler.java
```diff
@@ -101,6 +112,17 @@ private void addSchemaElementsContext(String userId, EntityDetail glossaryTerm,
     }
 
 
+    private void addGlossaryCategories(String userId,
+                                       EntityDetail glossaryTerm,
+                                       AssetContext glossaryContext,
+                                       List<Relationship> termCategorization) throws OCFCheckedExceptionBase {
+        if (CollectionUtils.isEmpty(termCategorization)) {
```
davidradl, line 119: Would it not be better to have a positive if and put the subsequent line in the body of the if. So you do not need to issue the return?

danielaotelea: I consider an early return makes code more readable and clean.

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
danielaotelea: I consider an early return makes code more readable and clean.

popa-raluca (approved):

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/GlossaryContextHandler.java
immmus: @danielaotelea in this case I agree with @davidradl 
It is better to change the condition and do nothing if it is not met then do "return;"

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/listeners/AssetLineageOMRSTopicListener.java
davidradl: @danielaotelea  For me this is a suggestion, I do not want to battle over style; so for me this can stay as is,  if you do not want to change. I am happy for you to resolve this conversation.

PR #3522, opened 2020-08-25
Release 2.2 - update version & adjust release notes

planetf1:
Drops -SNAPSHOT from version string


planetf1:
The gradle fail can be skipped -- this is caused by the version change and a dependence of the build currently (short term measure) on the main egeria pom (from artifactory) to define the set of dependency versions we use -- prior to porting that list to native gradle definition.

Issue #3525, opened 2020-08-25
Fix gradle build after version bump


planetf1:
TBD ...

planetf1:
The gradle build is failing with:
```

Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
Execution failed for task ':open-metadata-implementation:frameworks:audit-log-framework:compileJava'.
> Could not resolve all dependencies for configuration ':open-metadata-implementation:frameworks:audit-log-framework:detachedConfiguration1'.
   > Could not find org.odpi.egeria:egeria:2.3-SNAPSHOT.
     Searched in the following locations:
       - https://jcenter.bintray.com/org/odpi/egeria/egeria/2.3-SNAPSHOT/maven-metadata.xml
       - https://jcenter.bintray.com/org/odpi/egeria/egeria/2.3-SNAPSHOT/egeria-2.3-SNAPSHOT.pom
       - https://odpi.jfrog.io/odpi/egeria-snapshot/org/odpi/egeria/egeria/2.3-SNAPSHOT/maven-metadata.xml
       - https://odpi.jfrog.io/odpi/egeria-snapshot/org/odpi/egeria/egeria/2.3-SNAPSHOT/egeria-2.3-SNAPSHOT.pom
     Required by:
         project :open-metadata-implementation:frameworks:audit-log-framework

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org

BUILD FAILED in 44s

```

The cause of this is due to the way the current gradle build uses the maven pom last published to artifactory as the 'platform' specification - ie a list of all the dependency versions.

This should be done natively in gradle, but didn't make the first cut of the gradle build - in part as there are multiple ways of implementing, be it via the springboot plugin, or using gradle's native depedency constraints. The latter is in many way preferable, but not supported seemingly by dependabot...

planetf1:
Interestingly one of the sources for the pom is https://odpi.jfrog.io/odpi/egeria-snapshot/org/odpi/egeria/egeria/2.2-SNAPSHOT/ - this has had no 2.2 builds after 14 Aug yet we only branched for 2.3 on 25 Aug

The root cause is as found in #3526 - without that we would have only noticed the gradle issue on the very first build of a new release (even if it should work)

PR #3530, opened 2020-08-26
Asset Lineage OMAS - Audit Log 

danielaotelea:
- Removed depricated code for OMRS Audit Log
- Error code class should implement ExceptionMessageSet
- Lineage Exception should extend OCFCheckedExceptionBase and use ExceptionMessageDefinition 
- Removed unused exception classes


planetf1:
Ignore the gradle issue - will be addressed in #3525

popa-raluca (approved):

planetf1:
@danielaotelea - the main issue with gradle failing was fixed this morning (no builds would work).

However in this PR there's a genuine failure as we need to match the pom.xml update with a change to build.gradle for the modified dependency. I can push a fix

Issue #3503, opened 2020-08-21
Correct presentation server license (apache 2.0 ?)


planetf1:
Picked up in a scan:

```
Priority: Medium
Some package.json files incorrectly list the project's license as "ISC". These should be corrected to "Apache-2.0"
1 file (hide files)
egeria/open-metadata-implementation/user-interfaces/presentation-server/nodejs/package.json
```

planetf1:
i think this is done now?

davidradl:
yes

PR #3535, opened 2020-08-27
#3503 refactor Subject area line ends.

davidradl:


davidradl:
@immmus It would be great if you could review this change, as you have been working in this area.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
```diff
@@ -75,16 +79,48 @@ public Line(Line template) {
...
+
+        String end1TypeName = endDef1.getEntityType().getName();
+        String end2TypeName = endDef2.getEntityType().getName();
+        // Maybe consider subtypes here? We cannot use the Repositoryhelper isTypeOf because Line is used client side
+        // which does not have access to the repository helper.
+        if (end1TypeName.equals("GlossaryTerm")) {
+            end1TypeName = "Term";
+        }
+        if (end2TypeName.equals("GlossaryTerm")) {
+            end2TypeName = "Term";
```
immmus, line 109: I would prefer use "GlossaryTerm" and remove these conditionals.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
```diff
@@ -75,16 +79,48 @@ public Line(Line template) {
...
+        this(name,LineType.Unknown);
     }
 
     public Line(String name, LineType lineType) {
         this.name = name;
         this.lineType = lineType;
+        this.setLineEnds();
+    }
+    protected void setLineEnds() {
+        OpenMetadataTypesArchiveAccessor archiveAccessor = OpenMetadataTypesArchiveAccessor.getInstance();
```
immmus, line 95: I am alarmed by the use of this service here...

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: The Subject Area API does not expose a concept called GlossaryTerm, only Term. I know it is easier internally as it matches the types.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: > I am alarmed by the use of this service here...

yes good spot - I guess this implies that we need access to the archive in the client libraries. That is a shame as it saves a lot of code.  I might try to generate some code from the archive so we do not have this dependancy.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
immmus: I think if we leave GlossaryTerm here, it won't hurt anyone) or it is worth considering revising domain names to map to internal types, as is done with Line
And that help to us in future.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
immmus: @davidradl I really dont know how better resolve this problem but it is problem...

planetf1:
The general gradle issue is addressed.

However this PR is failing since the change to use the open types archiver introduces a change in the pom.xml and we need to make the corresponding change in build.gradle to fix:

```
/home/vsts/work/1/s/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java:15: error: package org.odpi.openmetadata.opentypes does not exist
import org.odpi.openmetadata.opentypes.OpenMetadataTypesArchiveAccessor;
                                      ^
/home/vsts/work/1/s/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java:95: error: cannot find symbol
        OpenMetadataTypesArchiveAccessor archiveAccessor = OpenMetadataTypesArchiveAccessor.getInstance();
        ^
  symbol:   class OpenMetadataTypesArchiveAccessor
  location: class Line
/home/vsts/work/1/s/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java:95: error: cannot find symbol
        OpenMetadataTypesArchiveAccessor archiveAccessor = OpenMetadataTypesArchiveAccessor.getInstance();
                                                           ^
  symbol:   variable OpenMetadataTypesArchiveAccessor
  location: class Line
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
3 errors

FAILURE: Build failed with an exception.

```

Would you like to do this ? Or I can push a fix (later today) if you like

That being said I share @immmus 's concern that this new dependency affects the client - which could be problematic, so it maybe something to address after you've made any decision on refactoring that piece?

open-metadata-implementation/access-services/subject-area/subject-area-api/pom.xml
```diff
@@ -30,8 +30,8 @@
     <dependencies>
 
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
```
planetf1, line 33: Just as a FYI I checked the dependency tree and it's not too bad as this is plain java code:
```
org.odpi.egeria:subject-area-api:jar:2.3-SNAPSHOT
 [INFO] +- org.odpi.egeria:open-metadata-types:jar:2.3-SNAPSHOT:compile
 [INFO] |  \- org.odpi.egeria:repository-services-archive-utilities:jar:2.3-SNAPSHOT:compile
 [INFO] |     +- org.odpi.egeria:open-metadata-archive-file-connector:jar:2.3-SNAPSHOT:compile
 [INFO] |     \- commons-io:commons-io:jar:2.7:compile
```
It's a 200k jar though it's dependenceis (esp commons-io) adds more - but probably typical for any of our clients.

Which really leaves any broader design discussions .. I'm less concerned with the inclusion per se?

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: @immmus I have resolved the problem by hard coding the values in the Line children. The code is ready for a new review now.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/HasA.java
```diff
@@ -3,34 +3,37 @@
...
 @JsonInclude(JsonInclude.Include.NON_NULL)
-@JsonIgnoreProperties(ignoreUnknown=true)
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class HasA extends Line {
     private static final Logger log = LoggerFactory.getLogger(HasA.class);
     private static final String className = HasA.class.getName();
 
-    private static final String[] PROPERTY_NAMES_SET_VALUES = new String[] {
+
+    private static final String[] PROPERTY_NAMES_SET_VALUES = new String[]{
```
immmus, line 36: I would like to raise the issue of removing these unused fields from all classes.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/HasA.java
```diff
@@ -62,72 +65,69 @@
     private static final java.util.Set<String> ATTRIBUTE_NAMES_SET = new HashSet<>(Arrays.asList(ATTRIBUTE_NAMES_SET_VALUES));
```
immmus, line 65: And it.

immmus (changes requested):

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
```diff
@@ -92,35 +85,44 @@ public Line(String name, LineType lineType) {
...
+//        LineEnd lineEnd1 = new LineEnd(getEnd1TypeName(),
+//                                       getEnd1AttributeName(),
+//                                       getEnd1AttributeDescription(),
+//                                       endDef1.getAttributeCardinality()
+//                                       );
+//        LineEnd lineEnd2 = new LineEnd(end2TypeName,
+//                                       endDef2.getAttributeName(),
+//                                       endDef2.getAttributeDescription(),
+//                                       endDef2.getAttributeCardinality()
+//                                       );
```
immmus, line 116: Should we leave this commented-out code?

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
```diff
@@ -92,35 +85,44 @@ public Line(String name, LineType lineType) {
...
-                                       );
-        LineEnd lineEnd2 = new LineEnd(end2TypeName,
-                                       endDef2.getAttributeName(),
-                                       endDef2.getAttributeDescription(),
-                                       endDef2.getAttributeCardinality()
-                                       );
-        setEnd1(lineEnd1);
-        setEnd2(lineEnd2);
+    protected LineEnd getLineEnd1() {
+       return null;
```
immmus, line 122: May be make Line abstract and make these methods are abstract?

immmus (changes requested):

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: @immmus good spot :-).

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/HasA.java
davidradl: I think we should do this separately - please raise an issue for that. This code was originally generated so has some left over code from that time, the generation ended up in the FVT for open types. It would be good to clean it up.

immmus: Well. https://github.com/odpi/egeria/issues/3541

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: > May be make Line as abstract and make these methods abstract?

I had a play with this idea, the OmrsLine extends from Line and would need implementations of these methods that does not make sense for it. I have left this as is - as I do not want to change this without thinking it through in more detail what is required for OMRSLine.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: > @immmus good spot :-).

fixed

davidradl:
@immmus shall I merge?

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
immmus: When do you want to think about it? I'm afraid this will remain so and in the future you can bring us trouble.

immmus: > > May be make Line as abstract and make these methods abstract?
> 
> I had a play with this idea, the OmrsLine extends from Line and would need implementations of these methods that does not make sense for it. I have left this as is - as I do not want to change this without thinking it through in more detail what is required for OMRSLine.

When do you want to think about it? I'm afraid this will remain so and in the future you can bring us trouble.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: this code no longer exists

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/HasA.java
davidradl: I assume this refers to unused code o to be removed in another issue

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
davidradl: > OMRSLine

It turned out to be straight forward, this OMRSLine was not used in the code so I have deleted and made the methods abstract. Thanks @immmus

immmus:
@davidradl Thank you, you can merge.

PR #3545, opened 2020-08-29
# Remove unused fields and resolve NPE

immmus:
I've removed a lot of unused code since 2018. I also moved the repetitive logic to the main class. Resolved NPE issue after last update.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -176,7 +176,7 @@
         <antlr.version>3.5.2</antlr.version>
         <snappy.version>1.1.7.7</snappy.version>
         <snakeyaml.version>1.26</snakeyaml.version>
-        <springdoc.version>1.4.5</springdoc.version>
+        <springdoc.version>1.4.6</springdoc.version>
         <classgraph.version>4.8.89</classgraph.version>
         <swagger.version>2.1.4</swagger.version>
         <groovy.version>3.0.5</groovy.version>


open-metadata-implementation/access-services/subject-area/subject-area-api/pom.xml
```diff
@@ -6,7 +6,6 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <groupId>org.odpi.egeria</groupId>
```
davidradl, line 9: I think it best o leave this line in - so it is explicit unless @planetf1 thinks it is ok.

davidradl (commented):
Looks good to me -  I just wanted to check the pom change with Nigel @planetf1

open-metadata-implementation/access-services/subject-area/subject-area-api/pom.xml
immmus: This is a line and so is explicitly specified in the parent tag;)
Screenshot 2020-09-01 at 11 15 56

planetf1: I agree with @immmus - there is no need to have this groupId in (though should be harmless). Our other poms typically do not include it.

planetf1 (approved):

davidradl (approved):

Issue #3543, opened 2020-08-28
Created New Server for Improved Development in OMAG Configuration UI


jmmadsen:
Currently, the UI is being served in production mode for local development, making it very inefficient. I have created two new directories `cra-client` and `cra-server` that have the same functionality as the original code, but are optimized for development. I also refactored the server code to be more organized, as opposed to in one large file.

@davidradl - we discussed you testing the UI with a configured server to ensure I did not break anything during this process. I have updated the README.md, but to start you should:

1. Navigate into `cra-client`, and run `npm install`.
2. Navigate into `cra-server`, run `npm install`, then run `npm start`.
3. UI should be accessible at `https://localhost:3000/aaa` - I could only get it working with Firefox.

jmmadsen:
More on my Chrome issues. I do not think this is a new issue I introduced in CRA, I experience the same issue with the original code. My version of Chrome is `84.0.4147.105`. 

![image](https://user-images.githubusercontent.com/57506409/92259609-4f8ee200-eea5-11ea-9bfa-c2bff0839015.png)

planetf1:
Looks like the same issue I came across in #1403 -- In the draft PR #3238 I created new certs which def avoid this issue -- but I haven't yet had a chance to go back and complete the work.

In the PR are the commands to generate a compliant cert. Additionally the cert already checked in for the server should be fine.

I'd be pretty confident that is the only reason it doesn't work in chrome - simply because later browsers are being more strict with their security validation.

Either
 a) If @davidradl agrees get this PR in & then raise an issue to correct the certs (I can help)
 b) Wait until #3238 is completed
 c) let me know when the PR is nearly done (now?) and I can push an update -- but it may take me a day or three due to a blocking release issue

jmmadsen:
Hi Nigel, thank you for the information.

I would prefer that this PR be merged and we address as we continue development focused on the UI. I have spoken with @dkneisly and others who are working on different pieces of the UI, and it would be helpful for them to have a working development environment to dynamically view changes. Let me know what you all think.

planetf1:
I'm ok with any of the above  - just probably best to ensure there is an active issue referencing the change we need to do to security (and perhaps any of the other issues that you agree with we identified in the PR review - note that you can use the menu in github web ui to create an issue from a comment, though I think where possible group related ones together). It's good to ensure each comment is responded too/closed with an action ie not doing it/disagree / fixed, or will be addressed in later issue (add note). ok?

[...]

PR #3552, opened 2020-09-01
#3543 New Directory for Create React App UI

jmmadsen:
Currently, the UI is being served in production mode for local development, making it very efficient. I have created two new directories cra-client and cra-server that have the same functionality as the original code, but are optimized for development. I also refactored the server code to be more organized, as opposed to in one large file.
@davidradl - we discussed you testing the UI with a configured server to ensure I did not break anything during this process. I have updated the README.md, but to start you should:
Navigate into cra-client, and run npm install. Navigate into cra-server, run npm install, then run npm start. UI should be accessible at https://localhost:3000/aaa - I could only get it working with Firefox.


davidradl (changes requested):
1) This pr has conflicting changes. Please could you resolve the conflicts, so I can see a clean diff. Currently is shows files have changed outside of the presentation server, which I was not expecting. 
 
2) I looked at the code and I wonder 
a) we could leave as is and rename the cra-client and cra-server to client and server as they are in the cra folder
b) move the cra client and cra server and cra-ssl up to the same level as the nodejs. 
I am think option b would remove a level of nesting that we do not need.  

3) The .gitignore needs to be amendments to the existing .gitignore at the top level (not in the sub folders); as this is where all .gitignore information is specified for Egeria.   

4) It looks strange to me that the npm package is in a src folder. I would expect the src folder to be a level below the package.json, as package.json is a build file and  not source.

PR #3555, opened 2020-09-02
fix vendor data type, foreign key catalog/schema, tests

yevgenmar:


planetf1 (approved):

davidradl (commented):
I am not sure about the context of this change. I do not see an associated issue explaining the need for the change. Looking at the actual code, it seems that the word module is changed to physicalDefinition, which seems to be a module. It seems that if physicalDefinition is a module, it would be more explicit to use the name physicalDefinitionModule.I am not that familiar with the code so am not seeing how the pr title links to the code change, maybe some comments using words from the pr title could help link the code change to the pr title.

planetf1:
Suggest adding a link to the issue once created (as we discussed today).
Happy to merge -- but let me know if you want to make any changes based on david's comment before I click the button :-)

yevgenmar:
I've created https://github.com/odpi/egeria/issues/3581 to describe nature and goals of the plugin.
JSON element 'module' was used by mistake. The term physicalDefinition is used by CA as first consumer of the OMAS.
I would keep this change as is.

planetf1:
I think we should merge this as it moves the analytics modeling omas forward, even though it is still in development

davidradl:
> I think we should merge this as it moves the analytics modeling omas forward, even though it is still in development
makes sense.

planetf1:
we agree, so merging :-)

PR #3561, opened 2020-09-02
#3543 Presentation-Server UI Dev Refactoring

jmmadsen:
Currently, the UI is being served in production mode for local development, making it very efficient. I have created two new directories cra-client and cra-server that have the same functionality as the original code, but are optimized for development. I also refactored the server code to be more organized, as opposed to in one large file.

@davidradl addressed your comments from previous PR in this commit

Signed-off-by: JMadsen <jacob.madsen@ibm.com>


planetf1 (changes requested):
I've not reviewed in detail (yet) but it seems we have a lot of varied filetypes being added in this PR, including many files which one would normally expect to be generated, or the output of the build process.

There are 2586 files in total which seems quite a lot

If so I think we should not merge as it pollutes the git history, and could also have some license implications

I think we have lots of node packages for example, which I would not expect in the source tree, but rather be retrieved (as per package.json) during the build process

I note we have a few areas where we should be more standard too - we use .md instead of markdown, and not usually .files for licenses - though if they are genuine contributions then for those they can be handled in another PR.

But I think we should check and cleanup prior to merge. Given this I recommend moving to draft status (anyone can move back at any point) to reduce the chance of an accidental merge

davidradl (changes requested):
Hi Jacob, I have had a quick look. It is saying you have changed over 2500 files. I think you need /** at the end of the folder lines in the top level .gitignore. Also you need to remove the .gitignores from the cra client and cra-server folders.

planetf1:
As well as changes to the gitignore, files that were included when the change originally made need removing, and the commit to add then remove needs squashing

jmmadsen:
I adjusted the top level .gitignore and deleted the lower level ones. I am confused as to why the 2000+ node_modules are appearing in the PR. When I check my forked branch on GH, I do not see any `node_modules` that have been comitted.

@planetf1 do you mind explaining your squashing comment a little further, so that I can clean this draft up for acceptance?

planetf1:
Screenshot 2020-09-03 at 21 28 03

planetf1:
Most of those files are .js - I assume they were from node modules though not necessarily.

'Squashing' a commit is basically merging multiple commits together

So let's say you had a commit that added loads of files, then another that deleted them, (2000+), since it was an accident it would be best not to include all those extra files in the repository -- which they will be even if immediately deleted.

In those cases it would be better to merge those two commits together - so the add and the delete will cancel out, and you'll just have a commit with the changes you intended -- this is assuming that huge number of file changes is a mistake? We really don't want those to go into the repository

To do this, first fix the issue -- remove any files that shouldn't be checked in for example.

Once that's done you can do `git rebase -i HEAD~3` for example to go back 3 commits. When the editor shows you change all except the first commit to 'squash' (instead of 'pick') then save... once done you would need to do a `git push --force` to update your PR branch.

I know this is a little complex so happy to help if you ping on slack

Another alternative could be to start back with your code before the changes and just apply manually if it is just a small proportion of the 2000+ checked in. A completely new commit - a new branch if you like, even a new PR. could be simpler.

Generally I think it'a good not to squash commits so we can all track the full history, but when it's in a PR prior to merge and correcting an error like which files to checkin (if it is - perhaps I'm wrong) then I think it is the best route to follow

planetf1:
In terms of build process, note that it's best, if we can, to ensure any created files go under 'target' (for maven - or 'build' for gradle) as they are automatically excluded. Elsewhere we have to be very careful not only with the .gitignore, but also in the way maven artifacts are  built. (but could be done in a later PR) - as well as what then happens in getting them into our distribution packages and/or docker images. I think once the basic checkin is sorted, anytime after that (doesn't have to be immediately) I'd be happy to get on a call to discuss? I know little about node.js so will be a learning experience for me too - would be looking to determine what the best practice for building these assets is, or we can work together from basic principles.

davidradl:
> In terms of build process, note that it's best, if we can, to ensure any created files go under 'target' (for maven - or 'build' for gradle) as they are automatically excluded. Elsewhere we have to be very careful not only with the .gitignore, but also in the way maven artifacts are built. (but could be done in a later PR) - as well as what then happens in getting them into our distribution packages and/or docker images. I think once the basic checkin is sorted, anytime after that (doesn't have to be immediately) I'd be happy to get on a call to discuss? I know little about node.js so will be a learning experience for me too - would be looking to determine what the best practice for building these assets is, or we can work together from basic principles.

The current presentation server build puts all the built artefacts in dist. For a production build I think this would continue to make sense, as we are creating the distribution. @jmmadsen I do not think there is an intermediate 'target' folder like with java & Maven, that is this assembled to the distribution. For npm, the distribution is created directly into the dist folder (currently). For dev mode I assume that the source code is run in place and not from any built file; @jmmadsen is this correct?

.gitignore
```diff
@@ -115,9 +115,9 @@ open-metadata-implementation/user-interfaces/presentation-server/nodejs/yarn.loc
 open-metadata-implementation/user-interfaces/presentation-server/nodejs/node/**
 open-metadata-implementation/user-interfaces/presentation-server/nodejs/.env
 
-open-metadata-implementation/user-interfaces/presentation-server/cra-client/node_modules
+open-metadata-implementation/user-interfaces/presentation-server/cra-client/node_modules/**
```
davidradl, line 118: You need to create equivalents of the other excludes from the existing presentation server .gitignore otherwise these files will end up getting into git. 
dist/**        if it has its own output folder.
package-lock.json
.vscode
.log
yarn.lock

If any other files could be generated in the new build, then these need to be ignored as well.

davidradl (changes requested):
I have made a comment on the git ignore. Once the change is minimal and only contains what you are changing I will be able to give this a more thorough review. I agree with Nigel, either squash the commits as recommended or start a fresh with a clean set of commits with a correct .gitignore so no unnecessary generated files are present.

jmmadsen:
Hi all, lots of discussion here. I will try to address everything, but I agree with Nigel that at this point it may be more productive to set up a call. That way, he can learn more about Node and I can learn more about Maven and the team's best practices for PRs.

@planetf1 I do not have any experience with Maven yet, so look forward to learning more.

@davidradl You are correct that in dev mode, no static files are built. Using CRA, they are only built after running `npm run build` inside of `cra-client`. It seems the distribution directory is a relic from the old code then, because using CRA the static files are built under a directory called `build`.

Depending on your all's schedules, I am free to discuss before 1pm EST or anytime next week. Thanks for the feedback!

davidradl:
> Hi all, lots of discussion here. I will try to address everything, but I agree with Nigel that at this point it may be more productive to set up a call. That way, he can learn more about Node and I can learn more about Maven and the team's best practices for PRs.
> 
> @planetf1 I do not have any experience with Maven yet, so look forward to learning more.
> 
> @davidradl You are correct that in dev mode, no static files are built. Using CRA, they are only built after running `npm run build` inside of `cra-client`. It seems the distribution directory is a relic from the old code then, because using CRA the static files are built under a directory called `build`.
> 
> Depending on your all's schedules, I am free to discuss before 1pm EST or anytime next week. Thanks for the feedback!

Hi @jmmadsen, I am happy to have a call. It think it is likely going to be Monday for me as I have other commitments at 1pm EST - we can talk at the Monday meeting we have scheduled. I do not think that understanding more about Maven should be a blocker, I suspect the integration with Maven which is producing the distribution will be the same/ very similar as is it today; unless you think otherwise. The current Maven piece is just a wrapper around npm. 

Thanks for the information about  `build` folder, that is fine if the build folder is an output only folder. As we want to be able to easily identify generated artefacts from the build , so we do not check them into git and can easily delete them.

planetf1:
Quick note. Don't use build as the gradle build assumes that. Can discuss Monday . 

We should look at common patterns for building node projects in a maven and gradle environment.

PR #3571, opened 2020-09-04
# Fixed some code smells from sonar

immmus:
I fixed some code smells from sonar. If this is normal, I propose to make it a practice for all contributors so that over time it will be completely eradicated from these shortcomings. This task is not for one person.
I think if we solve the problems in sonar it will definitely be for the better for the project.


immmus:
@planetf1 Can you look it is?;)

Issue #3568, opened 2020-09-04
Unable to create schemas as part of Asset Owner OMAS


cmgrote:
It does not appear to be possible at the moment to create (and attach) schemas to Assets as part of the Asset Owner OMAS.

For example, against an existing Asset, if you attempt to POST the following to `/servers/omas/open-metadata/access-services/asset-owner/users/{{user}}/assets/{{assetGUID}}/schemas/top-level-schema-type`:

```json
{
    "class": "SchemaTypeRequestBody",
    "schemaTypeProperties": {
        "class": "SchemaTypeProperties",
        "typeName": "RelationalDBSchemaType",
        "qualifiedName": "host=my.system.com::database=MYDBNAME::schema_type=MYSCHEMA",
        "displayName": "MYSCHEMA"
    }
}
```

You receive a response of:

```json
{
    "class": "GUIDResponse",
    "relatedHTTPCode": 400,
    "exceptionClassName": "org.odpi.openmetadata.frameworks.connectors.ffdc.PropertyServerException",
    "actionDescription": "addSchemaType",
    "exceptionErrorMessage": "OMAG-REPOSITORY-HANDLER-400-003 An unexpected error was returned by the metadata server during addSchemaType request for open metadata access service Asset Owner OMAS on server omas; message was OMRS-REPOSITORY-400-022 A null unique identifier (guid) for a TypeDef object has been passed as the entityTypeGUID parameter on a addEntity request to open metadata repository omas",
    "exceptionErrorMessageId": "OMAG-REPOSITORY-HANDLER-400-003",
    "exceptionErrorMessageParameters": [
        "OMRS-REPOSITORY-400-022 A null unique identifier (guid) for a TypeDef object has been passed as the entityTypeGUID parameter on a addEntity request to open metadata repository omas",
        "addSchemaType",
        "Asset Owner OMAS",
        "omas"
    ],
    "exceptionSystemAction": "The system is unable to process the request because of an internal error.",
    "exceptionUserAction": "Verify the sanity of the server.  This is probably a logic error.  If you can not work out what happened, ask the ODPi Egeria community for help."
}
```

The exception stack trace on the console of the OMAG platform running the OMAS indicates the following:

```
12:06:46.118 [REST:Asset Owner OMAS:addSchemaTypeToAsset] ERROR o.o.o.c.ffdc.RESTExceptionHandler - Exception from addSchemaTypeToAsset being packaged for return on REST call
[...]

PR #3574, opened 2020-09-07
Asset Owner OMAS fixes

cmgrote:
Fixes issues identified in Asset Owner OMAS while trying to make use of the APIs directly, including #3568

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
+++ b/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-core.yaml
@@ -18,6 +18,12 @@ spec:
     - port: 9443
       targetPort: 9443
       nodePort: {{ .Values.service.nodeport.core }}
+  {{ if .Values.debug.egeriaJVM }}
+    - name: debug
+      port: 5005
+      targetPort: 5005
+      nodePort: {{ .Values.service.nodeport.debug.core }}
+  {{ end }}
   selector:
     app.kubernetes.io/name: {{ include "myapp.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
@@ -62,8 +68,16 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "myapp.fullname" . }}-configmap
+          env:
+            {{ if .Values.debug.egeriaJVM }}
+            - name: JAVA_DEBUG
+              value:  "true"
+            {{ end }}
           ports:
             - containerPort: 9443
+          {{ if .Values.debug.egeriaJVM }}
+            - containerPort: 5005
+          {{ end }}
           readinessProbe:
             tcpSocket:
               port: 9443
--- a/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-datalake.yaml
+++ b/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-datalake.yaml
@@ -18,6 +18,12 @@ spec:
     - port: 9443
       targetPort: 9443
       nodePort: {{ .Values.service.nodeport.datalake }}
+    {{ if .Values.debug.egeriaJVM }}
+    - name: debug
+      port: 5005
+      targetPort: 5005
+      nodePort: {{ .Values.service.nodeport.debug.datalake }}
+    {{ end }}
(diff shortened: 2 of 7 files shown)


open-metadata-implementation/common-services/ocf-metadata-management/ocf-metadata-handlers/src/main/java/org/odpi/openmetadata/commonservices/ocf/metadatamanagement/handlers/SchemaTypeHandler.java
```diff
@@ -1059,7 +1059,7 @@ else if (instanceElementType.getElementOrigin() == ElementOrigin.LOCAL_COHORT)
             return true;
         }
 
-        return false;
+        return instanceElementType.getElementOrigin() == ElementOrigin.CONFIGURATION;
```
mandy-chessell, line 1062: Configuration is not local cohort

open-metadata-implementation/common-services/ocf-metadata-management/ocf-metadata-handlers/src/main/java/org/odpi/openmetadata/commonservices/ocf/metadatamanagement/handlers/SchemaTypeHandler.java
```diff
@@ -1395,13 +1395,25 @@ public SchemaAttributeBuilder  getSchemaAttributeBuilder(SchemaAttribute schemaA
      *
      * @param schemaAttribute bean to interrogate
      * @return guid of the type
+     * @throws InvalidParameterException if the provided schema attribute type cannot be resolved
      */
-    private String getSchemaAttributeTypeGUID(SchemaAttribute   schemaAttribute)
+    private String getSchemaAttributeTypeGUID(SchemaAttribute   schemaAttribute) throws InvalidParameterException
```
mandy-chessell, line 1400: The typeGUID should be set where the SchemaAttribute object is created.  Isn't there a method like createEmptySchemaAtttibute?  This seems too late.

mandy-chessell (changes requested):

open-metadata-implementation/common-services/ocf-metadata-management/ocf-metadata-handlers/src/main/java/org/odpi/openmetadata/commonservices/ocf/metadatamanagement/handlers/SchemaTypeHandler.java
cmgrote: Configuration is the default value when no other value is provided -- and unfortunately without this an external sourced entity creation is attempted (without any external metadata collection ID)...  For some reason this logic differentiating between external and local cohort is only present when trying to create schema attributes, and not schema types (not sure why they're different in this regard)?

open-metadata-implementation/common-services/ocf-metadata-management/ocf-metadata-handlers/src/main/java/org/odpi/openmetadata/commonservices/ocf/metadatamanagement/handlers/SchemaTypeHandler.java
cmgrote: The `SchemaAttribute` object is created by the `clone()` method of the `SchemaAttributeProperties` bean (see line 238 of `AssetOwnerRESTServices`)...  Not sure how I could do this other than extending the bean itself (to include the typedef GUID), or passing through the repository helper to the bean to be able to lookup the typedef GUID (?)

The actual persistence of the `SchemaAttribute` object isn't until after the current point anyway (line 248 of `AssetOwnerRESTServices`), so any exception should break us out before persistence of either the schema type or its attributes...

planetf1:
@cmgrote is the intent still to merge, I notice it's been idle for 6 weeks now. Some remerging will be needed with the recent changes

cmgrote:
I think the intention was always that some parallel changes @mandy-chessell was making would replace the need for this PR, but have lost track of what the current state is on that / those (?)

planetf1:
@mandy-chessell  ?

mandy-chessell:
I believe it is fixed

planetf1:
@cmgrote Agree? Can we close this now?

Issue #2120, opened 2019-11-28
Build project reports for Egeria


planetf1:
'mvn site' can l build some standard reports including test coverage, dependencies, licensing etc.
It can also build javadoc, and a cross ref with the code for browsing
It may be possible to add pmd, spotbug reports

This could replace the build plugin approach currently used for enabling CVE (-PCVE) and licensing (-PLicensing) in a more integrated, better format way.

Aggregate reports can cause some issues with dependencies -- one set of issues was fixed recently in #2054  - with those done, we can revisit reporting

The output of mvn site can be seen on many projects, for example
https://maven.apache.org/plugins/maven-project-info-reports-plugin/dependencies.html

is part of report for the report plugin itself (!) which nicely lays out licenses used by the project

https://maven.apache.org/plugins/maven-project-info-reports-plugin/surefire-report.html is a summary of test case execution

Many more exist, and we are a much larger project than this example, but auotmatic generation of such information could help accurately describe the project,  if run as part of a release (or a daily build)

cmgrote:
As discussed on today's call, this sounds like an avenue to pursue in order to aggregate the JavaDocs into a single project-wide set that could be published into GitHub pages (for example), and therefore would provide a publicly-accessible, version-controlled (ie. with each release) set of detailed documentation for developers (?)

I've done a little experimentation on my side, and it looks like separating this into the `reporting` phase (so it is only constructed on a `mvn site` rather than the normal build process) ensures that it does not bog down anything else.  (Which would be good, as a `mvn site` with the default settings I'm testing with takes ~45 minutes to complete!)

planetf1:
I agree with the goal of building  more complete aggregated docs

In recent months though, IMO, our use of maven is starting to creak at the schemes. We've hit issues with inflexibility, performance, lack of support in the community, bugs.

I would also want to consider how we might drive the creation via gradle ( I did actually try this briefly and it looks reasonable) as I think it will be better long term.

cmgrote:
Fine for me to do that for the longer-term, but I'm keen even on a quick-fix for the near-term (ie. assuming it could take another several months or more for the gradle build to finalise and be embedded in our release pipelines, etc) (?)

planetf1:
Sure if there's a quick way to do it go for it. It was more an ongoing concern of investing too much in legacy

planetf1:
No easy plugins to do this in gradle. Will open up issues on any particular aspects that need addressing.
Closing

PR #3586, opened 2020-09-08
Adds initial reporting (for mvn site) to aggregate documentation

cmgrote:
Usage: run `mvn site` and (after quite a long while) this will build content into `target/site` including a full set of JavaDocs spanning all sub-modules.

My hope would be that we could run a nightly build (at 45 minutes local runtime, probably far too long for a per-merge build?) that automatically publishes the updated documentation into a sub-directory of this repository, which we can then link to via the website?

This would ensure that we have updated JavaDocs always publicly available, without needing to build locally, and would also capture each release's JavaDocs against each particular release branch / tag.

(See also #2120)

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/audit-log-connectors/audit-log-slf4j-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/auditlogstore/slf4j/SLF4JAuditLogStoreConnector.java
+++ b/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/audit-log-connectors/audit-log-slf4j-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/auditlogstore/slf4j/SLF4JAuditLogStoreConnector.java
@@ -38,7 +38,7 @@ public class SLF4JAuditLogStoreConnector extends OMRSAuditLogStoreConnectorBase
 
         super.validateLogRecord(logRecord, methodName);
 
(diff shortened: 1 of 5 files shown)


planetf1:
I tried the same in gradle and the build appeared to take 57s (I ran the normal gradle build first -- which doesn't do all maven does, but is incremental and 3-4x quicker). Then I ran the 'aggregateJavaDocs' task (after adding plugin) which took 57s.

A quick skim through seems to suggest the build is complete - though I can't be sure of the effect of any caching. It did however seem to report warnings on each area as it went through (very quickly). This was locally on macOS

This is consistent with my maven experience. Maven's determination to stick to the pom structure really slows it down. Gradle is more happy with the filesystem structure -- tends to work loads better for any aggregation work.

We could -- for example -- add a seperate pipeline, and just build docs with gradle, or even with this performance we could  probably include in the main gradle build (even PR?)  & publish as an artifact --perhaps with a chained build to copy that artifact to some hosted website.

Looks more promising than maven IMO which is too slow and clunky?

planetf1:
I will add to a PR so you can try if you like? Unfortunately the gradle build still has some hardcoding in it -- WIP ...

planetf1:
See #3593 for an alternative (quick/dirty -- for review) approach using gradle.

planetf1:
I should add that I see no reason we can't integrate both the maven approach AND the gradle approach if we wish, as long as not default - then can figure out which works better..

planetf1:
Rather than leave this for work around building docs to complete I propose we merge this small incremental update in any case as I don't see downsides

PR #3593, opened 2020-09-09
#2120 add support to Build aggregate javadoc

planetf1:

@cmgrote - this is an example of my thought to approach aggregate javadoc

Rough and ready but raising as a PR so you can see, indeed try? See if you get comparable results?

Go to the top level tree and run './gradlew build' for a normal build or './gradlew aggregateJavadocs' to build the aggregated docs. You don't need gradle installed.

this PR does not build the aggregate docs by default - possible next step...

For me the aggregate build took 57s . Some of that may have been saving from cached results -- not tested.
But at this kind of time there is potential to add not only into merge build, but a PR build, and indeed - if gradle becomes default, a developer build.


planetf1:
Rather than leave this for work around building docs to complete I propose we merge this small incremental update in any case as I don't see downsides - after fixing the conflict

PR #3600, opened 2020-09-10
AL OMAS Client

danielaotelea:
- AL OMAS client added
- unit tests

### DIFF
--- a/open-metadata-implementation/user-interfaces/presentation-server/nodejs/src/client/components/GlossaryAuthor/components/GlossaryAuthorNavigation.js
+++ b/open-metadata-implementation/user-interfaces/presentation-server/nodejs/src/client/components/GlossaryAuthor/components/GlossaryAuthorNavigation.js
@@ -11,13 +11,12 @@ import {
 } from "./GlossaryCard/GlossaryCard";
 import GlossaryImage from "../../../images/Egeria_glossary_32";
 import getNodeType from "./properties/NodeTypes.js";
-import { issueRestGet } from "./RestCaller";
+import { issueRestGet, issueRestDelete } from "./RestCaller";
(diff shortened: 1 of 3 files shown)


open-metadata-implementation/access-services/asset-lineage/asset-lineage-client/pom.xml
```diff
@@ -30,6 +30,37 @@
 
     <dependencies>
 
+        <dependency>
```
planetf1, line 33: We'll need comparable changes in the build.gradle with similar dependencies added (gradle can be more fussy that transitive dependencies are added). The other build.gradle files should show how to handle regular and test dependencies. You should also find examples where mockito, ocf etc are used.

The gradle build doesn't yet create our maven artifacts or run FVT, or indeed build the UI ... and needs hardcoded versions removed but the hope is to do this in future - it's SO much quicker and more flexible.

To add - it does get the dependencyManagement content from the main pom but in a slightly dubious way, again any issues here and I'll take a look

To build you can run './gradlew build' from the top level 

I can pull the PR and make the changes needed if required, or point you in the right direction.

planetf1 (changes requested):

lcpopa (approved):
Code looks good. Only the dependencies need another go, as requested by Nigel

open-metadata-implementation/access-services/asset-lineage/asset-lineage-client/pom.xml
danielaotelea: Hi, Nigel! Can you check if the changes from the latest commit solve the issue, please?

planetf1 (approved):
Thanks for updating @danielaotelea  - the gradle updates look good.

PR #3628, opened 2020-09-17
#3727 default the server type in Server Author

davidradl:


davidradl:
@dkneisly I made this change to the code in order to allow it to work. It might be short term change - but it does allow the client server author to function.

Issue #3629, opened 2020-09-17
View services are not working correctly 


davidradl:
View services need to get the remote server name and platform url from configuration in order to connect to the back end server. Currently there is a bug in the code : 

Post of `https://localhost:9443/open-metadata/admin-services/users/garygeeke/servers/cocoView1/view-services/rex`
with body of `    { 
         "class":"ViewServiceConfig",
         "viewServiceAdminClass":"org.odpi.openmetadata.viewservices.rex.admin.RexViewAdmin",
         "viewServiceFullName":"Repository Explorer OMVS",
         "viewServiceOperationalStatus":"ENABLED",
         "OMAGServerName":"cocoMDS1",
         "OMAGServerPlatformRootURL":"https://localhost:9443"
      }`

Should configure this. What happens is that the OMAGServerName ends up in the ViewOptions. The code in GlossaryAuthorViewAdmin does this:
 `  this.instance = new GlossaryAuthorViewServicesInstance(this.serverName,
                                                                   auditLog,
                                                                   serverUserName,
                                                                   maxPageSize,
                                                                   this.viewServiceConfig.getOMAGServerName(),
                                                                   this.viewServiceConfig.getOMAGServerPlatformRootURL());`

but the OMAGServer Name and PlatformRootURL are null. 

This circumvention works :
`            String remoteServerName = (String) this.viewServiceConfig.getViewServiceOptions().get("OMAGServerName");
            String remoteServerPlatformRoot = (String) this.viewServiceConfig.getViewServiceOptions().get("OMAGServerPlatformRootURL");

            this.instance = new GlossaryAuthorViewServicesInstance(this.serverName,
                                                                   auditLog,
                                                                   serverUserName,
                                                                   maxPageSize,
                                                                   remoteServerName,
                                                                   remoteServerPlatformRoot);`

It seems this content is ending up the viewOptions.

PR #3630, opened 2020-09-17
#3629 Improve and document view service admin requests 

davidradl:
I could not get the view services working with the existing code. I have added this change to now configure a view service by passing a ViewServiceConfig object rather than view options. This new object supplies a OMAG Server name and PlatformRootUrl (which were already there in the code). With this in place, it is possible to enable all view services with one rest call passing values these 2 properties. The viewOptions would then only contain options specific to a particular view.  

As it was not very obvious how to configure the view services, I have added some documentation similar to the access service one, it is not that obvious that a class needs to be supplied in the body so Jackson knows how to deserialise the payload. 

@planetf1 please could you check this does not break any existing  notebooks or Docker images. The body is now mandatory in the configure view service requests.

### DIFF
--- a/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/ViewServiceConfig.java
+++ b/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/ViewServiceConfig.java
@@ -87,6 +87,7 @@ public class ViewServiceConfig extends OMAGServerClientConfig
     {
         this.viewServiceId = viewRegistration.getViewServiceCode();
         this.viewServiceName = viewRegistration.getViewServiceName();
+        this.viewServiceFullName = viewRegistration.getViewServiceFullName();
         this.viewServiceURLMarker = viewRegistration.getViewServiceURLMarker();
         this.viewServiceAdminClass = viewRegistration.getViewServiceAdminClassName();
         this.viewServiceDescription = viewRegistration.getViewServiceDescription();


planetf1:
So far I don't believe we configure view services in the notebooks 

I suspect there won't be any difference between docker and non docker, so you can run the notebooks locally by following

https://egeria.odpi.org/open-metadata-resources/open-metadata-tutorials/lab-infrastructure-guide/running-natively.html

This will ensure you haven't broken the notebooks, and just leaves a small exposure on docker. Just running configure/start notebooks should suffice

After merging the PR, and a new image being built (takes an extra 45m max after merge) you can deploy the lab chart as normal in k8s, or with docker compose & should pick up the new images. A new issue could be raised on master if anything found

This doesn't require anything beyond the docs we currently have really.

I agree it would be useful to either
 a) Document how to build your own docker images, and test them (it is, but needs linking into main test doc) - see https://github.com/odpi/egeria/tree/master/open-metadata-resources/open-metadata-deployment/docker , but should cover how to access a repository etc
OR
 b) Automatically build at least the base docker image for every PR. This actually only takes a few seconds, it's just a case of a repository to use, to avoid cluttering up docker hub, and managing the cleanup.  See #1734

b) is probably is the easiest as it can be done on common infrastructure, and avoids extra load on the developer other than just running them. I'm inclined to add it to my 'nearer term' list as I'm currently working in that area. (1-2 releases)

grahamwallis:
David wrote "With this in place, it is possible to enable all view services with one rest call passing values these 2 properties"

Just an observation - the Rex, Tex and Dino view services either don't or won't use these properties. They are or will be configured with resource endpoints.

davidradl:
> David wrote "With this in place, it is possible to enable all view services with one rest call passing values these 2 properties"
> 
> Just an observation - the Rex, Tex and Dino view services either don't or won't use these properties. They are or will be configured with resource endpoints.

@grahamwallis Hi Graham, that is correct, as coded these 2 values will be available to Tex, Rex and Dino, if you want them. The advantage of this is that  we can activate all the view services using one call (same as access service), as if a view service needs these they will be mandatory.  I suspect most view services will require a back end; my thinking was it best to move these up to the top level.  Otherwise we will need to issue the configuration for each service separately.  I guess the downside of this approach is that there will be configuration that a view service does not need. When you configure rex or tex individually, you do not need to specify these values.

davidradl:
Talking with @grahamwallis  he suggested introducing a solutionViewService that has the 2 top level OMAG server fields. I think this is a good idea. What do you think @mandy-chessell ?

grahamwallis:
When I manage to push the Dino changes to the repo there will be SolutionViewService and IntegrationViewService types as subtypes of ViewService. So this should be a NO OP if it is OK architecturally.

davidradl:
> When I manage to push the Dino changes to the repo there will be SolutionViewService and IntegrationViewService types as subtypes of ViewService. So this should be a NO OP if it is OK architecturally.

As discussed these 2 fields are already defined in OMAGServerClientConfig, which is the super class of ViewServiceConfig

Issue #2739, opened 2020-03-12
New "samples" repository?


cmgrote:
We've discussed on a number of calls for a while now the potential for creating a separate code repository for our "samples", but there did not seem to be an explicit issue for this decision (hence adding this one).

Part of what we need to decide is what should be included under this umbrella term of "samples":

- Existing sample data and metadata (eg. under `open-metadata-resources/open-metadata-deployment/sample-data`)?
- Existing Ansible playbook-based loading / clearing (also under `open-metadata-resources/open-metadata-deployment/sample-data/...`)?
- Existing Helm chart-based deployment (under `open-metadata-resources/open-metadata-deployment/charts`)?
- Existing Compose-based deployment (under `open-metadata-resources/open-metadata-deployment/compose`)?
- Existing Docker image builds (under `open-metadata-resources/open-metadata-deployment/docker`)?
- Existing tutorials (under `open-metadata-resources/open-metadata-tutorials`)?
- Others?

Related: #1643

planetf1:
I thought it worth setting the scene of where we are at with our repos.

It's perhaps worth considering what structure we want moving forward - of course this may change, but deciding at the top level of repos (vs smaller details of builds, artifacts, modules) is significant

My initial reaction is that on a very specific question of whether we should have a repo that aids end users in code samples, I would absolutely agree, as it's unit they can fork, clone, modify, build.

Taking just one example, like sample metadata or helm charts there are many levels
 - data and charts may focus only on core, pure egeria function
 - They may support testing
 - They may relate to external components (including connectors)

However the list above goes a lot beyond this, so I think we need to take a step back:

**Current Repos**

We have a number of project for or used by egeria

* egeria
* egeria-connector-apache-atlas
* egeria-connector-ibm-information-server
* data-governance
* egeria-dev-projects
* egeria-palisade

Some have a different scope within ODPi but have some relevance to us

* tsc
* artwork

Whilst others relate to other activities within the ODPi - either past or present:

* website
* ci-management
* OpenDS4All
* bi-ai
* specs
* bigtop
* self-certification-reports
* security-guide
* ambari
* hive
* hadoop

For the purpose of this discussion I suggest focussing on the first set only.

It's worth noting we have numerous other outlets (and arguably) repositories for our work including
 * Blog sites (including odpi)
 * Azure dev pipelines
 * maven central
 * Artifactory, (+ bintray, JCenter)
 * docker.io
 * Slack

But for this discussion will focus on the core artifacts under source code management

[...]

PR #3631, opened 2020-09-17
Migrates initial set of materials to new samples repository

cmgrote:
Per #2739 this removes the following from this repository (already present on the new https://github.com/odpi/egeria-samples repository):

- open-metadata-demos
    - in-memory-repository-demo (links to it elsewhere in docs are updated to point to new repo)
    - virtual-data-connector
- VDC helm chart
- Ansible roles
- Minimal sample data (and embedded Ansible deployment scripts)

### DIFF
--- a/Content-Organization.md
+++ b/Content-Organization.md
@@ -1,10 +1,13 @@
 <!-- SPDX-License-Identifier: CC-BY-4.0 -->
 <!-- Copyright Contributors to the ODPi Egeria project. -->
 
-
 ## ODPi Egeria Site organization
-  
-The Egeria content is organized into the following modules:
+
+The Egeria content is organized across a number of repositories:
+
+### Egeria Core repository
+
+The Egeria core repository contains the core Egeria functionality, and is organized into the following modules:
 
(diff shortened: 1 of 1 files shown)


open-metadata-resources/open-metadata-deployment/README.md
```diff
@@ -3,8 +3,6 @@
```
planetf1, line 3: This file probably needs rewriting (can be in a separate issue!). Much of this content is fine in the new helm/vdc tree on our samples repo.

This doesn't really add value here and I would probably replace with a simple description of the subdirectories
I would also note that additional samples are now to be found in the new repo
The actual charts directory probably then needs an improved readme which refers to the existing tutorial information for setting up the lab chart (as that is all that is left) - similar to what we have in the compose/tutorials directory

As above - happy if you want to give this a go, or otherwise ok to merge and raise a new issue which I can address

planetf1 (commented):
Agree with the proposal.
I added some notes about cleaning up the docs on the deployment directory - but could be in a followon issue or PR (feel free to assign to me if needed. ok either way)

PR #3640, opened 2020-09-21
Vertical Lineage backend queries

lcpopa:


lcpopa:
Also, removed a check in condensation that was taking glossary terms into account.

popa-raluca (approved):

Issue #3664, opened 2020-09-23
Issue trying to edit an glossary that is in an archive  


davidradl:
If there is a glossary in an archive and I attempt to issue it in the UI. I get this message in the response:

{"class":"SubjectAreaOMASAPIResponse","relatedHTTPCode":400,"exceptionClassName":"org.odpi.openmetadata.viewservices.glossaryauthor.services.GlossaryAuthorViewGlossaryRESTServices","actionDescription":"update(isReplace=false) for Glossary","exceptionErrorMessage":"OMAG-COMMON-400-016 An unexpected org.odpi.openmetadata.accessservices.subjectarea.handlers.SubjectAreaGlossaryHandler exception was caught by update(isReplace=false) for Glossary; error message was OMAG-REPOSITORY-HANDLER-400-003 An unexpected error was returned by the metadata server during updateGlossary request for open metadata access service Subject Area OMAS on server cocoMDS1; message was OMRS-REPOSITORY-400-056 The OMRS repository connector operation updateEntityProperties from the OMRS Enterprise Repository Services can not locate the home repository connector for instance 8763209e-5133-4630-970f-d696ba3030c3 located in metadata collection dd9e1327-3120-46b3-a4de-8d42ac2b6515","exceptionSystemAction":"The system is unable to process the request.","exceptionUserAction":"Review the error message and other diagnostics created at the same time.","result":[]}

- The OMRS issues this message OMRS-REPOSITORY-400-056. I think this message could be improved. Presumably it could tell this is from an archive and give a better message. @mandy-chessell what do you think? 
- in this case there is an omrs error - which ends up as an Exception that comes though to the OMAs, which with handle as an unexpected error. An example of an unexpected error is:
`  UNEXPECTED_EXCEPTION(400, "OMAG-COMMON-400-016",
              "An unexpected {0} exception was caught by {1}; error message was {2}",
              "The system is unable to process the request.",
              "Review the error message and other diagnostics created at the same time."),`  
subject area has another version of this error. It seems that status code 400 means there is a bad request, which does not seem great - though i am not sure if there is a better status code.   I would like to look at the status code in the ui and put up something meaningful. I am thinking of saying that the operation failed.   

  
- The UI currently does not display anything indicating there has been an error, it needs to display a reasonable message (not more intelligible than, a view error , wrapping an omas error wrapping an omrs error). 

I am going to use this issue to address the UI error part.

immmus:
[...]

PR #3666, opened 2020-09-23
#3664 Add in error messages to glossary navigation

davidradl:


planetf1 (approved):
This looks like an improvement, and it's still an area in development, so I think it worth merging

PR #3676, opened 2020-09-24
Adds escaping of Lucene extended grammar extension characters

### LINKED ISSUE #3563: qualifiedName issues with graph repository?
In testing some changes for lineage, I've noticed some differences in the graph repository vs the in-memory repository when it comes to certain qualifiedNames.

Specifically, qualifiedNames that start with (and elsewhere contain) the following characters are problematic when running with a graph repository configured (knock-on effect higher up the stack -- I don't see an exception from the graph repository itself, but the entities that have such qualifiedNames do not seem to be stored in the graph repository as the exception I am able to see from the OMAS is that these entities do not exist):

- The hash symbol `#`
- The ampersand `&`

Running precisely the same code, OMASes, and data set with the in-memory repository configured in place of the graph repository shows no issues (all stored and retrievable as expected).

Similarly, if I change the qualifiedNames not to start with (or contain) these characters (for example, to use an underscore (`_`) instead) then the code, OMASes, and data set all work fine with both the graph repository and the in-memory repository.

So for now we can workaround this; however, I suspect this could cause broader issues if there are specific characters that cannot be included at the start or elsewhere within a qualifiedName string in the graph repository (?)

grahamwallis:
Fixes #3563

### DIFF
--- a/open-metadata-implementation/user-interfaces/presentation-server/nodejs/src/client/components/Dino/Dino.js
+++ b/open-metadata-implementation/user-interfaces/presentation-server/nodejs/src/client/components/Dino/Dino.js
@@ -39,6 +39,9 @@ import ReadmeMarkdown                           from './README.md';
 
 import "./dino.scss";
 
+
+
+
 export default function Dino() {
 
   const containerDiv = useRef();
@@ -52,8 +55,6 @@ export default function Dino() {
 
   const [readme, setReadme]             = useState( { markdown : '' } );
   const [readmeStatus, setReadmeStatus] = useState("idle");
-
-  
 
   let workingHeight = cltHeight - 50;
   let workingWidth  = cltWidth - 265;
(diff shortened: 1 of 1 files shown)


open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/graphrepository/repositoryconnector/GraphOMRSMetadataStore.java
```diff
@@ -1852,6 +1852,9 @@ private String convertSearchStringToJanusRegex(String str) {
                         case '|':
                         case '+':
                         case '?':
+                        case '#':
+                        case '&':
+                        case '<':
```
planetf1, line 1857: do we need to handle double quote?

PR #3693, opened 2020-09-26
# Added configuration scan.packages

immmus:
This change will allow to add custom apis to run egeria platform.
I think this is important change and must be introduce as quickly as possible.

For example:
`scan.packages=org.odpi.openmetadata.*,other.scan.packages`


bogdan-sava (approved):

planetf1:
From a technical perspective I understand how this means we can ensure other components are added into the chassis and picked up by spring to offer rest services.

However I wonder if you could elaborate more on the use case -- as we may need to consider how they fit in with the overall egeria architecture, common services etc - or extend these accordingly.

planetf1 (changes requested):
asking for clarification

immmus:
> asking for clarification

Hi! For example I will want to create my custom apis for my own tasks and run and settings on egeria. Now I need create packeges with name org.odpi.openmetadata.* but after this change I can call as I want and just add it to properties or vm option that spring can detect my services.

planetf1:
Thanks @immmus - understand that aspect of the change.

I was wondering if just having extra code picked up by the spring component scan was sufficient, and what type of code was running

For example, writing a new OMAS requires updating some of our code around registration (currently) as well as adding in the REST services. We have an architecture for pluggable connectors, but these are not found by spring, instead the classes are specified in configuration. Most recently @mandy-chessell has been adding the code for the Integration daemons which should make many more general tasks & services easier

So I'm trying to understand the extension model you are trying to use- whether it makes sense generally, whether we need to do more, what we need to document in addition to just allowing code to be run in the chassis.

bogdan-sava:
@planetf1 what updates are you referring?
This change is not affecting the current functionality and makes the app itself more open, in the sense that allow spring scanning more packages with custom code that is not part of Egeria.
I think is a very good idea.

immmus:
> @planetf1 what code updates are you referring to?
> This change is not affecting the current functionality and makes the app itself more open, in the sense that allow spring scanning more packages with custom code that is not part of Egeria.
> I think is a very good idea.

Yes, we already use this functionality in the our project. We were create our own omas apis and run with egeria spring-chasis. But we cannot configure our packages to scan so we named packages org.odpi.openmetadata.* so spring can to detect our service.

planetf1:
I do like the idea of making it easier for OMASs to be added outside the main egeria build, and contributed dynamically..  but..
If the use case is to add a new OMAS, there are additional changes needed if they are to truly act as an omas .

For example currently we have a hardcoded list in AccessServiceDescription.java which is part of our admin services, and this needs to know about the OMAS. We also have OMASServiceInstanceHandler which currently has specific OMASs listed.

I don't know the full extent here, but it is why I am trying to understand the use case.

Whilst adding to the spring component scan helps in location of classes, I'm concerned that we may impact some of the core egeria functionality unless those other areas are aware of this approach .. which we need to work through the design for

Any thoughts @mandy-chessell ?

immmus:
> I do like the idea of making it easier for OMASs to be added outside the main egeria build, and contributed dynamically.. but..
> If the use case is to add a new OMAS, there are additional changes needed if they are to truly act as an omas .
> 
> For example currently we have a hardcoded list in AccessServiceDescription.java which is part of our admin services, and this needs to know about the OMAS. We also have OMASServiceInstanceHandler which currently has specific OMASs listed.
> 
> I don't know the full extent here, but it is why I am trying to understand the use case.
> 
> Whilst adding to the spring component scan helps in location of classes, I'm concerned that we may impact some of the core egeria functionality unless those other areas are aware of this approach .. which we need to work through the design for
> 
> Any thoughts @mandy-chessell ?

In our case, we did not have any problems with this.
We do not use AccessServiceDescription because we use our own description and this constructor:
Screenshot 2020-10-06 at 15 06 34

mandy-chessell:
This needs a broader review since it is moving away from the intended approach from how an OMAS is added to the server.

immmus:
> This needs a broader review since it is moving away from the intended approach from how an OMAS is added to the server. For example, the list above is only used to help the administration services build the configuration document for our built in OMASs. The OMAS does not need to be in the list.

I added a service similarly schema as GlossaryView service with the exception of some of my additions and its work all good.
Screenshot 2020-10-06 at 20 48 06
Screenshot 2020-10-06 at 20 48 27
Screenshot 2020-10-06 at 20 48 38
Screenshot 2020-10-06 at 20 48 52

And get these beans in my service classes. I am running spring-chasis with -Dloader.path="mylibs" option and spring loads my apis to egeria. And I can my own apis with other egeria core stack.

immmus:
In principle, it doesn't matter to me whether this change is made or not, I found a way for me to competently and without unnecessary dependencies to make my own patches without affecting the main logic of egeria, but I think this change will be very useful not only for me, but also for other users. And this change does not have any problems in itself for the current logic.

mandy-chessell:
Sorry @immmus, my comment was misleading - I think your change is good - I was responding to the concern over the enumerations for the inbuilt OMASs.  I think it is fascinating that you are attempting to add a custom OMAS.  We had this as an early requirement but you are the first to try it :)

immmus:
> Sorry @immmus, my comment was misleading - I think your change is good - I was responding to the concern over the enumerations for the inbuilt OMASs. I think it is fascinating that you are attempting to add a custom OMAS. We had this as an early requirement but you are the first to try it :)

No problem)). I'm not just trying to add my own OMAS, I already added it and we are using it:)

bogdan-sava:
Exactly! I have also added an omas without change the code, but the package was named to be scanned (Using constructor instead of enum ).

planetf1 (approved):

PR #3696, opened 2020-09-28
#3675a Update version to 2.4-SNAPSHOT

planetf1:

Updates master to 2.4-SNAPSHOT as per #3675a


planetf1:
Note #3525 applies here - first build after a version bump causes an issue. Will aim to fix for next release

Issue #3703, opened 2020-09-29
Support for generic handlers


mandy-chessell:
OMASs today use common handlers based around OCF beans.  They each map their private beans to the OCF beans and then pass the OCF beans to the common handlers in ocf-metadata-management module.

The downside is that the OCF beans do not cover all of the open metadata types and they are complex structures.

This changes proposes to change the handlers to use generic beans.  This means that the OMASs can work with their own beans directly and no longer need to work with OCF beans.

The style is that instances are created/updated and deleted using the new handler APIs with its standard builder.  Each OMAS has their own converters that are used to initialized the handlers.

mandy-chessell:
There is more work in the generic handlers to create coverage across all of the open metadata types.  However I am going to close this one now and add support for new handlers in other issues.

Issue #3701, opened 2020-09-29
Correct type of Endianess property of the OperatingPlatform entity


mandy-chessell:
The UML model for the `endianess` property of the `OperatingPlatform` entity shows its type as an Endianness enum.

![0030-Hosts-and-Platforms](https://user-images.githubusercontent.com/12354195/94539955-cec1cc80-023d-11eb-9b24-ccb782a1addb.png)

Unfortunately, the code that adds this attribute to the type, set up the type of endianness to a string.  Since we can not update the type of an attribute, we need to deprecate the `endianness` attribute and add a new one.  The new attribute is called `byteOrdering`.

![0030-Hosts-and-Platforms](https://user-images.githubusercontent.com/12354195/94543062-7d1b4100-0241-11eb-895a-328b2d02d4ef.png)

Issue #3505, opened 2020-08-23
Extend schemaType simplification


mandy-chessell:
The original intent of the introduction of the **TypeEmbeddedAttribute** classification was to reduce the size of a relational schema.  This issue proposes to update the **SchemaTypeOption**, **MapToElementType** and **MapFromElementType** relationships to allow them to be connected to both a **SchemaAttribute** and a **SchemaType**.

This change simplifies the create processing of schema attributes because all schema type can now be linked to the schema attribute through the **TypeEmbeddedAttribute** classification.  It also reduces the size of document style schemas.

This has no implication for backward compatibility because all existing relationships of this type will still work and the schema element handlers will still support both patterns on read.

This is the current model for **SchemaTypeOption**:

![0501-Schema-Elements](https://user-images.githubusercontent.com/12354195/90972752-bf02ea00-e513-11ea-9f17-3c05ec69975f.png)

and the new model.

![0501-Schema-Elements](https://user-images.githubusercontent.com/12354195/90972887-ae06a880-e514-11ea-95bb-45d8343b2149.png)

This is the current model for **MapToElementType** and **MapFromElementType**.

![0511-Map-Schema-Elements](https://user-images.githubusercontent.com/12354195/90972786-f7a2c380-e513-11ea-8b5d-4d90a448cab0.png)

and the new model.

![0511-Map-Schema-Elements](https://user-images.githubusercontent.com/12354195/90972880-9deec900-e514-11ea-91d1-32293c03bae7.png)

Issue #3425, opened 2020-07-31
Search Keywords for Referenceables


mandy-chessell:
Many Referenceables - particularly Assets and SchemaElements have very little text in them.  This makes it hard for search/query processing to find them based on their content.  Even where there is text, it may not reflect the true content of the data.  For example, even though InfoSphere MDM can be deployed as a customer domain hub, its schema does not mention customer - it uses the term party which may not be an obvious search term for many users.

Linking glossary terms can help, although often there is only one linked glossary term since it is suppose to reflect the meaning of the content.  End users - once they find a good asset can add Informal tags (assuming they find it in the first place :) ).

This issue proposes a new type of entity called a **SearchKeyword**. The search keywords provide a mechanism to allow the asset owner to tag the asset - and linked elements such as the schema - with a variety of keywords that can be matched during a search.  This helps to boost relevant assets to the top of the search results.

This is what it looks like.  It is modelled on the search keywords attached to academic documents.  

![0012-Search-Keywords](https://user-images.githubusercontent.com/12354195/89053494-30a1ab00-d34f-11ea-9ec5-c3021e838a61.png)

The **SearchKeyword** entity stores the definition of the search keyword.  It is linked to the assets and their associated elements using the **SearchKeywordLink** relationship.  Related keywords can be linked together using the **RelatedKeyword** relationship to allow simple synonym type expansions of the search.

Search keywords can be added manually through the Asset Owner OMAS.   Some files such as documents and photos may
have keywords embedded in them.  These can be automatically discovered through metadata discovery and each stored as a SearchKeyword entity that are linked to their corresponding Asset properties using these same search keyword elements.

PR #3707, opened 2020-09-29
Intregration Daemons and Generic Handlers

mandy-chessell:
This pull request primarily delivers the integration daemon OMAG server and support for the database integrator integration service.  This supports the automatic capture of relational database metadata. (#2512)

Testing this work uncovered a number of issues and missing functions needed to support the expansion out of integration services to supporting asset managers and other third party technologies.

The first was a series of issues in the open types (#3701, #3505, #3332).  This impacted the support for the OCF handlers.  These have been replaced with generic handlers to allow the common functions to be used beyond the types supported by OCF (#3703).

The mechanism for supporting latest changes, templated asset curation, anchor management, external instances and cascaded deletes needed to be improved (#3308).  This has been implemented in the generic handlers.

As the generic handlers were being worked through, a circular dependency between the handlers, metadata security and the OCF occurred because the metadata security module used OCF beans in its API.  This has been address by giving the metadata security module its own tailored beans, simplifying the interface.  I also added support for connection selection to the security module (#3419).

This PR included support for search keywords (#3425).

### DIFF
--- a/open-metadata-test/open-metadata-fvt/access-services-fvt/asset-consumer-fvt/build.gradle
+++ b/open-metadata-test/open-metadata-fvt/access-services-fvt/asset-consumer-fvt/build.gradle
@@ -5,8 +5,9 @@
 
 dependencies {
     implementation project(':open-metadata-implementation:access-services:asset-owner:asset-owner-api')
-    implementation project(':open-metadata-implementation:access-services:asset-consumer:asset-consumer-api')
     implementation project(':open-metadata-implementation:access-services:asset-owner:asset-owner-client')
+    implementation project(':open-metadata-implementation:access-services:asset-consumer:asset-consumer-api')
+    implementation project(':open-metadata-implementation:access-services:asset-consumer:asset-consumer-client')
     implementation project(':open-metadata-implementation:frameworks:audit-log-framework')
     implementation project(':open-metadata-implementation:frameworks:open-connector-framework')
     implementation project(':open-metadata-implementation:repository-services:repository-services-apis')


open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/beans/ElementStatus.java
```diff
@@ -0,0 +1,125 @@
...
+
+    /**
+     * toString() JSON-style
+     *
+     * @return string description
+     */
+    @Override
+    public String toString()
+    {
+        return "InstanceStatus{" +
```
grahamwallis, line 119: ElementStatus

open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/AssetExternalSchemaType.java
```diff
@@ -0,0 +1,219 @@
...
+/* Copyright Contributors to the ODPi Egeria project. */
+package org.odpi.openmetadata.frameworks.connectors.properties;
+
+import org.odpi.openmetadata.frameworks.connectors.properties.beans.ExternalSchemaType;
+import org.odpi.openmetadata.frameworks.connectors.properties.beans.SchemaType;
+
+import java.util.Objects;
+
+/**
+ * SchemaTypeChoice describes a schema element that has a choice of type.  This class returns the list of schema choices.
```
grahamwallis, line 11: AssetExternalSchemaType

open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/AssetExternalSchemaType.java
```diff
@@ -0,0 +1,219 @@
...
+
+    /**
+     * Standard toString method.
+     *
+     * @return print out of variables in a JSON-style
+     */
+    @Override
+    public String toString()
+    {
+        return "AssetSchemaTypeChoice{" +
```
grahamwallis, line 173: AssetExternalSchemaType

grahamwallis (changes requested):
Have skimmed through just over 1000 of the files and have the following comments. Most of them are cosmetic but some affect Javadoc or logging. Some of them may affect de/serialization and be problematic if the serialised JSON is not of the form expected by a REST client such as Postman or Python (notebook) command.

grahamwallis:
In total I had 44 review comments but they do not all seem to have made it through review submission - I still have the review page open and will try to get the rest of them across.

Content-Organization.md
```diff
@@ -61,29 +63,34 @@ The Egeria core repository contains the core Egeria functionality, and is organi
   * **[admin-services](open-metadata-implementation/admin-services)** - supports the configuration of the OMAG Server Platform.  This configuration determines which of the open metadata and governance services are active.
   * **[common-services](open-metadata-implementation/common-services)** - support modules that are reused by other services.
     * **[ffdc-services](open-metadata-implementation/common-services/ffdc-services)** - provides base classes and validation for First Failure Data Capture (FFDC).
-    * **[gaf-metadata-management](open-metadata-implementation/common-services/gaf-metadata-management)** - provides metadata management for the [Governance Action Framework (GAF)](open-metadata-implementation/frameworks/governance-action-framework) properties and APIs.
```
planetf1, line 64: Not an issue with the gradle pom, but this is a useful list of the affected modules. Can you verify that when built, each maven module either has packaging=pom if an aggregator, or packaging=jar, AND contains 3 jars for the a) source b) javadoc c) built jar - otherwise the new modules may not be releaseable. Hopefully should happen by default but worth checking. If neiher a jar or pom we should figure out how to handle on a case by case basis

grahamwallis (commented):
Submitting review again to see if it retrieves the remaining comments.

grahamwallis:
Well I guess that's a fail then. I'm not going to repeat the review so just for the record, my comments were generally around Javadoc header comments referring to a different class name (a simple copy-paste problem), toString methods also naming incorrect class (copy-paste problem), and a few other typos in comments and readme files such as word repeats and grammatical typos. The more important ones though were around serialisation of classes that contain a boolean property field. In some places these fields are named isPublic, isPrivate, etc. and the getter and setter methods for isPublic are called isPublic() and setPublic(). This will serialise and deserialise OK but the property field in the serialised JSON will be called 'public' rather than matching the Java field names. This is because Jackson will strip the "is" prefix. If another type of user (via REST or Python) is parsing the JSON, it will not match what they might expect. OK if we document the serialised format, but possibly subtle otherwise. My preference is to use rather explicit getter and setter methods (always prefixed with get or set) and use a simple property name rather than prefixed with "is". 

I think it would be a lot better to avoid massive PRs, they are much harder to review both in terms of time and the ease for each reviewer to find the changes they are best placed to review. Also it seems that with a PR with so many files (1399) the GitHub tools start to struggle. I spent a whole day reviewing and although I still have the review page open with my comments, there is no easy way to find the comments within the (collapsed) files that I reviewed (1004). It would be far better to break up work into smaller PRs, if only to help the reviewers.

mandy-chessell:
Thank you for the comments.  I did get them in emails so I think they are there.  On the javadoc and toString problems, most of the ones in the first commit are fixed - they were only in the first commit because intelliJ forces a `git add` whenever you rename or delete a file and each file refactor often requires manual update of the Javadoc at the top of the file.  This then comes in a later commit.

On the boolean properties in the Java beans, they are only ever used through the Java bean as we do not support direct calls through the REST APIs.  However, I agree we should be prepared to support the REST APIs directly in the future since we see many architects who are evaluating Egeria try to call it through the REST APIs.  We are also using the REST APIs thrugh the notebooks.   I also favour going with `getIsPublic()` and `setIsPublic()` style because it is clear then how it maps to the open types rather than continuing with intelliJ's approach.  This issue is reported in #2312 but I will pull the fix into this PR since it is reported in the review comments.
 
I agree, in general, on the large PRs - however in this case it was unavoidable because integration daemon (actually what this PR is delivering) needed to use the Data Manager OMAS for testing.  This uncovered issues in our support for schemas that impacted OCF and the open types and I was then changing a middle layer of code to create the new handlers.  It was only by migrating the 5 OMASs to the new handlers that I was able to flush out all of the patterns that the generic handlers had to support.  The testing also revealed other issues in the modules I was changing and in those both above and below where I was working.  For example, there were problems in the the admin services, repository services as well as the documentation since new features had been going in without updating it.

The alternative would be to push untested code into master which would have been fine a year ago but I think is less appropriate now.

mandy-chessell:
I take back the comment I made about the javadoc errors in the beans - I have just gone through changing all of the boolean getters and setters in Asset Consumer OMAS and Asset Owner OMAS and found a number of errors in the class name in the Javadoc and toString

open-metadata-implementation/common-services/generic-handlers/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>basic-file-connector</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.odpi.egeria</groupId>
+            <artifactId>avro-file-connector</artifactId>
```
planetf1, line 101: this will pull avro into many areas. Not necessarily a problem, but is it an intent to have dependencies on connectors - I had assumed we were striving for dynamic loading of all connectors?

planetf1 (commented):
An additional question -- Will any of these changes require updates to the assembly?

I don't see new connectors, and other dependencies should be managed by maven. Nor do I see utilities. So probably not, but a question worth asking

planetf1:
I thought it would be useful to build our docker images and validate the container environment.

However when following the git instructions command provided in the PR to test , I get a compile error:
```
12:33:13,168 [INFO] --- maven-compiler-plugin:3.8.1:compile (default-compile) @ tex-view-server ---
12:33:13,176 [INFO] Changes detected - recompiling the module!
12:33:13,180 [INFO] Compiling 6 source files to /Users/jonesn/IdeaProjects/egeria-maven/open-metadata-implementation/view-services/tex-view/tex-view-server/target/classes
12:33:13,319 [INFO] -------------------------------------------------------------
12:33:13,320 [ERROR] COMPILATION ERROR : 
12:33:13,320 [INFO] -------------------------------------------------------------
12:33:13,320 [ERROR] /Users/jonesn/IdeaProjects/egeria-maven/open-metadata-implementation/view-services/tex-view/tex-view-server/src/main/java/org/odpi/openmetadata/viewservices/tex/handlers/TexViewHandler.java:[187,68] cannot find symbol
  symbol:   variable VIEW_SERVICE_NULL_PLATFORM_NAME
  location: class org.odpi.openmetadata.commonservices.ffdc.OMAGCommonErrorCode
12:33:13,322 [INFO] 1 error
12:33:13,322 [INFO] -------------------------------------------------------------

```

This is likely due to more recent changes in master than you currently have in the PR (the instructions you can copy/paste include merging from master -- on the basis that prior to merging a PR it needs to be up to date in any case).

planetf1:
Just checking out this branch (without merging from master) I get the following error in the config notebook:
```
   ... configuring the dino Integration View Service for this server...
   ...... (POST https://dev:9443/open-metadata/admin-services/users/garygeeke/servers/cocoView1/view-services/dino )
   ...... Response:  {'class': 'VoidResponse', 'relatedHTTPCode': 400, 'exceptionClassName': 'org.odpi.openmetadata.adminservices.ffdc.exception.OMAGInvalidParameterException', 'actionDescription': 'configureViewService', 'exceptionErrorMessage': 'OMAG-ADMIN-400-039 The configureViewService service of OMAG server cocoView1 has been configured with a null root URL for its remote OMAG Server', 'exceptionErrorMessageId': 'OMAG-ADMIN-400-039', 'exceptionErrorMessageParameters': ['configureViewService', 'cocoView1'], 'exceptionSystemAction': 'The system is unable to accept this value in the configuration document because the server would not be able to operate correctly.', 'exceptionUserAction': 'The root URL is supplied by the caller to the OMAG server. This URL value needs to be corrected before the server can operate correctly.'}
```
I think there was a change added to the admin server. This was known working in the 2.3 release. The dino view service does not utilize the root URL. It appears validation has changed - so either we'll need to set it in the notebook or change the validation.

It's possible this is a different change on master that this branch has picked up.

planetf1:
Step #5 on the 'building a data catalog' notebook appears to fail on 'assetOwnerPrintAssets'
```
​
assetOwnerPrintAssets(cocoMDS1Name, cocoMDS1PlatformName, cocoMDS1PlatformURL, petersUserId, ".*file.*")
​
1 asset found:
 
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-5-d31c7baee9fc> in <module>
----> 1 assetOwnerPrintAssets(cocoMDS1Name, cocoMDS1PlatformName, cocoMDS1PlatformURL, petersUserId, ".*file.*")

<ipython-input-1-390f80ab5b28> in assetOwnerPrintAssets(serverName, serverPlatformName, serverPlatformURL, userId, searchString)
    183         else:
    184             print (str(len(assets)) + " assets found:")
--> 185         printAssetListSummary(assets)
    186         printAssetListDetail(assets)
    187 

<ipython-input-1-390f80ab5b28> in printAssetListSummary(assets)
     38     print(" ")
     39     for x in range(len(assets)):
---> 40         printAssetSummary(assets[x])
     41 
     42 def printAssetListDetail(assets):

<ipython-input-1-390f80ab5b28> in printAssetSummary(asset)
      5 def printAssetSummary(asset):
      6     elementType = asset.get('type')
----> 7     assetType = elementType.get('elementTypeName')
      8     assetGUID = asset.get('guid')
      9     assetQualifiedName = asset.get('qualifiedName')

AttributeError: 'NoneType' object has no attribute 'get'
```

planetf1:
Step #3 on 'understanding an asset' also seems to fail with:
```
Created Patient Asset
No comment added
Unexpected response from server platform Data Lake Platform at https://datalake:9443
 
Response: 
{
    "timestamp": "2020-10-06T12:49:43.684+00:00",
    "status": 400,
    "error": "Bad Request",
    "message": "",
    "path": "/servers/cocoMDS1/open-metadata/access-services/asset-consumer/users/peterprofile/assets/3781dab8-b5c0-4a18-9339-b65b928fc677/comments"
}
 
Setting owner on Patient Asset to tanyatidie ...
Setting governance zones on Patient Asset ...
Done.
```
In the server log I see:
```
core_1          | Tue Oct 06 12:49:47 GMT 2020 cocoMDS2 Exception OMAG-REPOSITORY-HANDLER-0003 An unexpected error org.odpi.openmetadata.repositoryservices.ffdc.exception.PropertyErrorException was returned to reclassifyEntity by the metadata server during updateAssetZones request for open metadata access service Asset Owner OMAS on server cocoMDS2; message was OMRS-REPOSITORY-400-028 A property called zoneMembership has been proposed for a new metadata instance of category ClassificationDef and type AssetOrigin; it is not supported for this type in open metadata repository cocoMDS2
core_1          | Tue Oct 06 12:49:47 GMT 2020 cocoMDS2 Exception OMAG-REPOSITORY-HANDLER-0003 Supplementary information: log record id 19d3b0a5-dcf2-4c81-a214-1c05025c7ca5 org.odpi.openmetadata.repositoryservices.ffdc.exception.PropertyErrorException returned message of OMRS-REPOSITORY-400-028 A property called zoneMembership has been proposed for a new metadata instance of category ClassificationDef and type AssetOrigin; it is not supported for this type in open metadata repository cocoMDS2 and stacktrace of 
core_1          | OCFCheckedExceptionBase{reportedHTTPCode=400, reportingClassName='org.odpi.openmetadata.repositoryservices.localrepository.repositorycontentmanager.OMRSRepositoryContentValidator', reportingActionDescription='updateEntityClassification', reportedErrorMessage='OMRS-REPOSITORY-400-028 A property called zoneMembership has been proposed for a new metadata instance of category ClassificationDef and type AssetOrigin; it is not supported for this type in open metadata repository cocoMDS2', reportedErrorMessageId='OMRS-REPOSITORY-400-028', reportedErrorMessageParameters=[zoneMembership, ClassificationDef, AssetOrigin, cocoMDS2], reportedSystemAction='The system is unable to store the metadata instance in the metadata repository because the properties listed do not match the supplied type definition.', reportedUserAction='Verify that the property name is spelt correctly and the correct type has been used. Correct the call to the metadata repository and retry.', reportedCaughtException=null, reportedCaughtExceptionClassName='null', relatedProperties=null}
core_1          | 	at org.odpi.openmetadata.repositoryservices.localrepository.repositorycontentmanager.OMRSRepositoryContentValidator.validatePropertiesForType(OMRSRepositoryContentValidator.java:2061)
core_1          | 	at org.odpi.openmetadata.repositoryservices.localrepository.repositorycontentmanager.OMRSRepositoryContentValidator.validateClassificationProperties(OMRSRepositoryContentValidator.java:1588)
core_1          | 	at org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.OMRSMetadataCollectionBase.classifyEntityParameterValidation(OMRSMetadataCollectionBase.java:4029)
core_1          | 	at org.odpi.openmetadata.repositoryservices.enterprise.repositoryconnector.EnterpriseOMRSMetadataCollection.updateEntityClassification(EnterpriseOMRSMetadataCollection.java:3277)
core_1          | 	at org.odpi.openmetadata.commonservices.repositoryhandler.RepositoryHandler.reclassifyEntity(RepositoryHandler.java:1833)
core_1          | 	at org.odpi.openmetadata.commonservices.generichandlers.OpenMetadataAPIGenericHandler.setClassificationInRepository(OpenMetadataAPIGenericHandler.java:292)
core_1          | 	at org.odpi.openmetadata.commonservices.generichandlers.OpenMetadataAPIGenericHandler.setClassificationInRepository(OpenMetadataAPIGenericHandler.java:194)
core_1          | 	at org.odpi.openmetadata.commonservices.generichandlers.AssetHandler.updateAssetZones(AssetHandler.java:1042)
core_1          | 	at org.odpi.openmetadata.accessservices.assetowner.server.AssetOwnerRESTServices.updateAssetZones(AssetOwnerRESTServices.java:1468)

```

planetf1:
Step 3 on the open-discovery notebook fails with:
```
assetDeduplicatorEngineName = "AssetDeduplicator"
assetDeduplicatorEngineDisplayName = "Asset Deduplicator Discovery Engine"
assetDeduplicatorEngineDescription = "Discovery engine for scanning the asset catalog and identifying which assets are duplicate definitions of the same physical asset."
​
​
assetDeduplicatorEngineGUID = createDiscoveryEngine(cocoMDS1Name,
                                                    cocoMDS1PlatformName,
                                                    cocoMDS1PlatformURL,
                                                    petersUserId,
                                                    assetDeduplicatorEngineName,
                                                    assetDeduplicatorEngineDisplayName,
                                                    assetDeduplicatorEngineDescription)
​
print (" ")
print ("The guid for the " + assetDeduplicatorEngineName + " discovery engine is: " + assetDeduplicatorEngineGUID)
print (" ")
​
OMAG-REPOSITORY-HANDLER-404-010 Multiple OpenDiscoveryEngine entities where found with a name of AssetDeduplicator: the identifiers of the returned entities are []; the calling method is getDiscoveryEngineByName, the name parameter is name and the server is cocoMDS1
 * The system is unable to process a request because multiple entities have been discovered and it is unsure which entity to use.
 * Investigate why multiple entities exist.  Then retry the request once the issue is resolved.
OMAG-REPOSITORY-HANDLER-500-001 An unexpected error org.odpi.openmetadata.repositoryservices.ffdc.exception.PropertyErrorException was returned to createEntity by the metadata server during createDiscoveryEngine request for open metadata access service Discovery Engine OMAS on server cocoMDS1; message was OMRS-REPOSITORY-400-028 A property called displayName has been proposed for a new metadata instance of category EntityDef and type SoftwareServerCapability; it is not supported for this type in open metadata repository cocoMDS1
 * The system is unable to process the request because of an internal error.
 * Verify the sanity of the server.  This is probably a logic error.  If you can not work out what happened, ask the ODPi Egeria community for help.
```

Will re-check these tests on master.

mandy-chessell:
@planetf1 do not waste time in testing the notebooks - I am working through them - we only created the PR to allow the build problems to be shared

planetf1:
> @planetf1 do not waste time in testing the notebooks - I am working through them - we only created the PR to allow the build problems to be shared

Apologies for the misunderstanding. I'll test again once the PR is ready for final review. Have deleted the test results to avoid cluttering the build discussion

bogdan-sava:
I think using the [builder design pattern](https://en.wikipedia.org/wiki/Builder_pattern) will make code more efficient and easier to understand.

There are also methods from repository helper  for creation, where using [java builder design pattern](https://www.tutorialspoint.com/design_pattern/builder_pattern.htm) would help a lot.

For example, instead of create Classification like this, with many parameter and 3 of them  null:

```
Classification classification = repositoryHelper.getNewClassification(serviceName,
                                                                                  null,
                                                                                  null,
                                                                                  InstanceProvenanceType.LOCAL_COHORT,
                                                                                  userId,
                                                                                  OpenMetadataAPIMapper.ASSET_OWNERSHIP_CLASSIFICATION_NAME,
                                                                                  typeName,
                                                                                  ClassificationOrigin.ASSIGNED,
                                                                                  null,
                                                                                  getOwnerProperties(userId,
                                                                                                     owner,
                                                                                                     ownerType,
                                                                                                     methodName));
```
it can be used a builder in the sense of builder design pattern.

planetf1:
I noticed the failure
```
[ERROR] 'dependencies.dependency.version' for org.odpi.egeria:open-metadata-assemblies:pom:distribution is missing. @ line 72, column 21
```
The other modules were updated in a recent fix for #3752
Remove the specification of the classifier 
```
            <classifier>distribution</classifier>
```
ie leaving the dependency for the assembly as
```
        <dependency>
            <artifactId>open-metadata-assemblies</artifactId>
            <type>pom</type>
            <scope>test</scope>
        </dependency>
```

lpalashevski:
Update

Regression testing fo Data Engine OMAS is still in progress there are issues in regards to use of deprecated update methods causing UserNotAuthorizedException so far not straight forward to fix. We are still trying to do it and fix forward the code. Testing continues today. Keep you posted.

mandy-chessell:
Thanks @lpalashevski - which update method is causing the problem - if it is problematic, could I restore the method causing the problem to its previous behaviour - or would you prefer to continue to move foreward?

Any call can return UserNotAuthorizedException when the security plugin is enabled so it may be better to continue to work to support this exception?

popa-raluca:
> Thanks @lpalashevski - which update method is causing the problem - if it is problematic, could I restore the method causing the problem to its previous behaviour - or would you prefer to continue to move foreward?
> 
> Any call can return UserNotAuthorizedException when the security plugin is enabled so it may be better to continue to work to support this exception?

DE OMAS is using the old updateEntity method from the repository handler, that has no externalSourceName & externalSourceGUID specified. Also, DE is using SchemaTypeHandler - updateSchemaType & updateSchemaAttributes methods that don't specify the externalSourceName&externalSourceGUID. Similar methods need to be created for updating external entities.

I will create an issue to address this and commit the changes after this PR is merged. We're still in the middle of testing and we will come back with a final result later today

lpalashevski:
We also see  different exception in OMRSRepositoryHelper being used by IGC (IGCOMRSRepositoryEventMapper and IGCRepositoryHelper)

`java.lang.NoSuchMethodError: org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.repositoryconnector.OMRSRepositoryHelper.getNewClassification
`

This is breaking also Asset Catalog OMAS and fails related federated calls going to the remote repository.

Looks like we need to align the implementation ing IGC connectors.
@mandy-chessell @cmgrote

popa-raluca:
I created issue #3843 for handling the changes needed in DE

mandy-chessell:
This is great feedback.  I need to restore any OMRSRepositoryHelper methods that I have been inadvertently deleted since these changes should not affect the IGC connector.  I will also add the update methods back into handlers.

planetf1:
I notice there are some gradle conflicts after I consolidate the version files.
To fix, use intellij search (it should prompt once pasted to use regex) and replace using file pattern 'build.gradle'

Search string: `implementation \'([\w\.]+):([\w\-]+):([\w\-]+).*\'`
Replace string: `implementation '$1:$2'`

This strips the version specification and is likely the only reason for those conflicts beyond deleted/moved files

There are separate updates to the top level build.gradle (dependency versions) & settings.gradle (plugin versions) but I would expect auto-merging would work fine there.

mandy-chessell:
> I think using the [builder design pattern](https://en.wikipedia.org/wiki/Builder_pattern) will make code more efficient and easier to understand.
> 
> There are also methods from repository helper for creation, where using [java builder design pattern](https://www.tutorialspoint.com/design_pattern/builder_pattern.htm) would help a lot.
> 
> For example, instead of create Classification like this, with many parameter and 3 of them null:
> 
> ```
> Classification classification = repositoryHelper.getNewClassification(serviceName,
>                                                                                   null,
>                                                                                   null,
>                                                                                   InstanceProvenanceType.LOCAL_COHORT,
>                                                                                   userId,
>                                                                                   OpenMetadataAPIMapper.ASSET_OWNERSHIP_CLASSIFICATION_NAME,
>                                                                                   typeName,
>                                                                                   ClassificationOrigin.ASSIGNED,
>                                                                                   null,
>                                                                                   getOwnerProperties(userId,
>                                                                                                      owner,
>                                                                                                      ownerType,
>                                                                                                      methodName));
> ```
> 
> it can be used a builder in the sense of builder design pattern.

The repository helper is a released interface since it is used by repository connectors.  It can be extended but not radically changed.  I could turn this particular method into 2 methods to support the two types of classifications (assigned and propagated).  That would eliminate one of the nulls since that parameter would only be needed on one of the new method calls.  The other two nulls are only nulls temporarily, waiting for all OMASs to support provenance properly.  When they do, the first 2 values will be filled out.

open-metadata-implementation/common-services/generic-handlers/pom.xml
mandy-chessell: This is deliberate and not new.  We build the connection to the asset as it is onboarded in Asset Owner - this needs the connector type information which is extracted from the connector's ConnectorProvider.  The alternative is to hardcode the information about the connector in the OMAS.

mandy-chessell:
> I noticed the failure
> 
> ```
> [ERROR] 'dependencies.dependency.version' for org.odpi.egeria:open-metadata-assemblies:pom:distribution is missing. @ line 72, column 21
> ```
> 
> The other fvt modules were updated in a recent fix for #3752, so for the data manager fvt remove the specification of the classifier
> 
> ```
>             <classifier>distribution</classifier>
> ```
> 
> ie leaving the dependency for the assembly as
> 
> ```
>         <dependency>
>             <groupId>org.odpi.egeria</groupId>
>             <artifactId>open-metadata-assemblies</artifactId>
>             <type>pom</type>
>             <scope>test</scope>
>         </dependency>
> ```

Where is this error - I am guessing it is in a POM file but wondered which one?

mandy-chessell:
@cmgrote @lpalashevski is this the getNewClassification() method that needs to be reinstated in the RepositoryHelper?

```
    public Classification getNewClassification(String               sourceName,
                                               String               userName,
                                               String               typeName,
                                               String               entityTypeName,
                                               ClassificationOrigin classificationOrigin,
                                               String               classificationOriginGUID,
                                               InstanceProperties   properties) throws TypeErrorException
```

mandy-chessell:
The changes I have made to the SchemaTypeHandler is to change it so that `updateSchemaType` and `updateSchemaAttribute` calls the original repository handler method that does not do the provenance checking.

planetf1:
> > The other fvt modules were updated in a recent fix for #3752, so for the data manager fvt remove the specification of the classifier

> Where is this error - I am guessing it is in a POM file but wondered which one?

* Click on 'details' next to one of the builds like Java 11
* Click on link at bottom - view more details on azure pipelines
* Click on the maven task in the middle pane
* scroll to find error

The current error is about SchemaTypeAttribute

The maven error was relating to use of classifier, so for example one can do:
```
find . -name 'pom.xml' -print | xargs -n50 grep -y 'classifier'
```
from the top level -- other than the main pom

The specific error that I saw from an earlier build was in `open-metadata-test/open-metadata-fvt/access-services-fvt/data-manager-fvt`

lpalashevski:
> > @cmgrote @lpalashevski is this the getNewClassification() method that needs to be reinstated in the RepositoryHelper?
> > ```
> >     public Classification getNewClassification(String               sourceName,
> >                                                String               userName,
> >                                                String               typeName,
> >                                                String               entityTypeName,
> >                                                ClassificationOrigin classificationOrigin,
> >                                                String               classificationOriginGUID,
> >                                                InstanceProperties   properties) throws TypeErrorException
> > ```
> 
> Yes looks like thats the one judging by the stack trace I see in the logs:
> 
> ` java.lang.NoSuchMethodError: org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.repositoryconnector.OMRSRepositoryHelper.getNewClassification(Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Lorg/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/properties/instances/ClassificationOrigin;Ljava/lang/String;Lorg/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/properties/instances/InstanceProperties;)Lorg/odpi/openmetadata/repositoryservices/connectors/stores/metadatacollectionstore/properties/instances/Classification;`



@mandy-chessell Can you please revert the method in the OMRSRepositoryHelper interface itself? .. IGC repository helper is still breaking because of this.

Other than this after reverting the old schema update methods no additional related issues found. From Data Engine OMAS functional perspective we are good to go with this change and update deprecated methods afterwards #3843

lpalashevski:
Additional method `getSkeletonEntitySummary` needs to be reverted in OMRSRepositoryHelper to support IGC repository connector, this is the signature

```
EntitySummary getSkeletonEntitySummary(String                 sourceName,
                                           String                 metadataCollectionId,
                                           InstanceProvenanceType provenanceType,
                                           String                 userName,
                                           String                 typeName) throws TypeErrorException;
```

open-metadata-implementation/repository-services/repository-services-spring/src/main/java/org/odpi/openmetadata/repositoryservices/rest/server/spring/LocalRepositoryServicesResource.java
```diff
@@ -2323,6 +2355,58 @@ public VoidResponse saveEntityReferenceCopy(@PathVariable String       serverNam
...
+    public ClassificationListResponse getHomeClassifications(@PathVariable String serverName,
+                                                             @PathVariable String userId,
+                                                             @PathVariable String entityGUID)
+    {
+        return restAPI.getHomeClassifications(serverName, userId, entityGUID);
+    }
+
+
+    /**
+     * Retrieve any locally homed classifications assigned to the requested entity.  This method is implemented by repository connectors that are able
```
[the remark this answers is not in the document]
mandy-chessell: Not sure I understand this comment.

open-metadata-implementation/admin-services/docs/concepts/integration-daemon.md
```diff
@@ -7,14 +7,21 @@ An **Integration Daemon** is an [OMAG Server](omag-server.md)
 that provides metadata exchange services between third party
 technology and the open metadata ecosystem.
 
-Each integration daemon is paired with an 
-[access service](../../../access-services) and supports a particular
-class of technology.
+The integration daemon interacts with the open metadata
+ecosystem through [Open Metadata Access Service (OMAS)](../../../access-services)
```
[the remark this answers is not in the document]
mandy-chessell: plural

open-metadata-publication/website/solutions/metadata-server-exchange/README.md
```diff
@@ -0,0 +1,200 @@
+<!-- SPDX-License-Identifier: CC-BY-4.0 -->
+<!-- Copyright Contributors to the ODPi Egeria project 2020. -->
+
+# Metadata Server Exchange
+
+The metadata server exchange solution describes how third party metadata servers can exchange metadata through
+an **Open Metadata Repository Cohort** or **Cohort** for short.
```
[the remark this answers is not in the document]
mandy-chessell: Open Metadata Repository Cohort is correct - https://egeria.odpi.org/open-metadata-implementation/repository-services/docs/open-metadata-repository-cohort.html#:~:text=An%20open%20metadata%20repository%20cohort%20is%20a%20collection,with%2C%20and%20receive%20metadata%20from%2C%20any%20other%20member.

planetf1 (approved):
Approve - as per team discussions so as to get broader exposure before next release

Issue #3714, opened 2020-09-30
Admin Services should create view service config with correct subclass


grahamwallis:

planetf1:
this is fixed and merged

Issue #3662, opened 2020-09-23
View service admin proposal.


davidradl:
Currently there are 2 types of view services, integration and governance services. Integration takes a list of platform orientated configuration and the governance services need a remote server omag name and platform root url , so they can communicate with an omas. I suggest:

1) We enhance the ViewServiceConfig to have ViewServiceType with enum values of Integration and governance Solution.
2) Each view service could then specify this type in ViewServiceDescription
3) This would allow us to have an admin call to start all integration view services, which would take the list config and a separate start all Governance solution call to start all governance solution calls. Each could specify mandatory parameters and have the payload as the appropriate child config. The server file could use the new type to identify which subset of view services to start.    
4) Optionally we could change the start of each service so there is a start an integration service and a start a governance solution service rather than the generic start view service call. 

What do you think?   @grahamwallis @mandy-chessell

mandy-chessell:
I do not like the idea of combining starting a service with configuring it.  In fact it does not really make sense as actually you are starting the view server not the view service.  I think it would be better if starting the view server continues to use the standard `instance` method that all of the other OMAG Servers use rather than introducing a special approach for view servers.

In terms of configuring - I think if there are two types of service then there should be two admin services to make it clear what the request body should contain.

grahamwallis:
I'm not sure that being able to start IntegrationViewServices separately from SolutionViewServices is necessary. As @mandy-chessell says above, we normally use the server 'instance' to start the server including all enabled services. 

I have used the terminology IntegrationViewServices and SolutionViewServices in the changes I made to the admin API configuration properties classes. I have not used the term 'Governance' there so we need to make sure we are being consistent. 

I noticed that the configuring-the-view-services.md file (checked in as part of #3629/#3630) doesn't really explain the differences between the two groups of view services or specify what configuration properties are mandatory and optional for each group. We should add more description to make this clearer I think.

davidradl:
@mandy-chessell @grahamwallis my mistake - I did not mean start I meant configure

PR #3715, opened 2020-09-30
Adds registration constructor to Integraton and Solution View Service…

grahamwallis:
… Config classes and modifies createViewServiceConfig to create subclasses

### DIFF
--- a/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/IntegrationViewServiceConfig.java
+++ b/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/IntegrationViewServiceConfig.java
@@ -6,6 +6,7 @@ import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import org.odpi.openmetadata.adminservices.configuration.registration.ViewServiceRegistration;
 
 import java.util.List;
 import java.util.Objects;
@@ -48,6 +49,17 @@ public class IntegrationViewServiceConfig extends ViewServiceConfig {
             resourceEndpoints = template.getResourceEndpoints();
         }
     }
+
+    /**
+     * Call the superclass to set up the default values for a view service using a view service description.
+     *
+     * @param viewRegistration fixed properties about the view service
+     */
+    public IntegrationViewServiceConfig(ViewServiceRegistration viewRegistration)
+    {
+        super(viewRegistration);
+    }
+
 
 
     /**
--- a/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/SolutionViewServiceConfig.java
+++ b/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/SolutionViewServiceConfig.java
@@ -2,6 +2,8 @@
 /* Copyright Contributors to the ODPi Egeria project. */
 package org.odpi.openmetadata.adminservices.configuration.properties;
 
+import org.odpi.openmetadata.adminservices.configuration.registration.ViewServiceRegistration;
+
 import java.util.Objects;
 
 /*
@@ -28,6 +30,17 @@ public class SolutionViewServiceConfig extends ViewServiceConfig {
         super(template);
         /* There is nothing else to copy yet... */
     }
+
+    /**
+     * Call the superclass to set up the default values for a view service using a view service description.
+     *
+     * @param viewRegistration fixed properties about the view service
+     */
+    public SolutionViewServiceConfig(ViewServiceRegistration viewRegistration)
+    {
+        super(viewRegistration);
+    }
+
 
 
     /**
--- a/open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
+++ b/open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
@@ -3,8 +3,10 @@
 package org.odpi.openmetadata.adminservices;
 
 import org.odpi.openmetadata.adminservices.configuration.OMAGViewServiceRegistration;
+import org.odpi.openmetadata.adminservices.configuration.properties.IntegrationViewServiceConfig;
 import org.odpi.openmetadata.adminservices.configuration.properties.OMAGServerClientConfig;
 import org.odpi.openmetadata.adminservices.configuration.properties.OMAGServerConfig;
+import org.odpi.openmetadata.adminservices.configuration.properties.SolutionViewServiceConfig;
 import org.odpi.openmetadata.adminservices.configuration.properties.ViewServiceConfig;
 import org.odpi.openmetadata.adminservices.configuration.registration.CommonServicesDescription;
 import org.odpi.openmetadata.adminservices.configuration.registration.ServiceOperationalStatus;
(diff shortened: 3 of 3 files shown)


grahamwallis:
Fixes #3714

open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/IntegrationViewServiceConfig.java
```diff
@@ -49,6 +50,17 @@ public IntegrationViewServiceConfig(IntegrationViewServiceConfig template)
         }
     }
 
+    /**
+     * Call the superclass to set up the default values for a view service using a view service description.
```
davidradl, line 54: registration not description

open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/SolutionViewServiceConfig.java
```diff
@@ -29,6 +31,17 @@ public SolutionViewServiceConfig(SolutionViewServiceConfig template)
         /* There is nothing else to copy yet... */
     }
 
+    /**
+     * Call the superclass to set up the default values for a view service using a view service description.
```
davidradl, line 35: registration

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
```diff
@@ -313,11 +315,44 @@ public VoidResponse configureAllViewServices(String userId,
     private ViewServiceConfig createViewServiceConfig(ViewServiceRegistration     registration,
                                                       ViewServiceConfig           requestedViewServiceConfig)
     {
-        ViewServiceConfig viewServiceConfig = new ViewServiceConfig(registration);
+
+        ViewServiceConfig viewServiceConfig = null;
+
+        if (requestedViewServiceConfig instanceof IntegrationViewServiceConfig) {
```
davidradl, line 321: I suggested in https://github.com/odpi/egeria/issues/3662 that we could have a type of the view service - that we could interrogate rather than using reflection.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
```diff
@@ -313,11 +315,44 @@ public VoidResponse configureAllViewServices(String userId,
...
+            createdViewServiceConfig.setResourceEndpoints(requestedIntegrationViewServiceConfig.getResourceEndpoints());
+            viewServiceConfig = (ViewServiceConfig) createdViewServiceConfig;
+        }
+        else if (requestedViewServiceConfig instanceof SolutionViewServiceConfig) {
+            /*
+             * The requested configuration is for a Solution View Service
+             */
+            SolutionViewServiceConfig requestedSolutionViewServiceConfig = (SolutionViewServiceConfig)requestedViewServiceConfig;
+            SolutionViewServiceConfig createdViewServiceConfig = new SolutionViewServiceConfig(registration);
+            createdViewServiceConfig.setOMAGServerPlatformRootURL(requestedSolutionViewServiceConfig.getOMAGServerPlatformRootURL());
```
davidradl, line 336: this line and next line could be moved out of the ifs -as these fields exist in ViewServiceConfig. They can be removed from the integration service code as well.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
```diff
@@ -313,11 +315,44 @@ public VoidResponse configureAllViewServices(String userId,
...
+            createdViewServiceConfig.setOMAGServerName(requestedSolutionViewServiceConfig.getOMAGServerName());
+            viewServiceConfig = (ViewServiceConfig) createdViewServiceConfig;
+        }
+        else {
+            /*
+             * Assume that the requested configuration is a vanilla view service configuration
```
davidradl, line 342: I suggest we fail the request if we do not recognise the view service as one of the 2 types we support

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
```diff
@@ -313,11 +315,44 @@ public VoidResponse configureAllViewServices(String userId,
...
+            viewServiceConfig = createdViewServiceConfig;
+        }
+
+        /*
+         * Always copy the view service options if any are present
+         */
         Map<String, Object> viewOptions = requestedViewServiceConfig.getViewServiceOptions();
         viewServiceConfig.setViewServiceOptions(viewOptions);
-        viewServiceConfig.setOMAGServerPlatformRootURL(requestedViewServiceConfig.getOMAGServerPlatformRootURL());
-        viewServiceConfig.setOMAGServerName(requestedViewServiceConfig.getOMAGServerName());
```
davidradl, line 320: why do we need to delete these lines - they should work where they are; unless I am missing something

davidradl (changes requested):

open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/IntegrationViewServiceConfig.java
grahamwallis: I used the same comment (almost) as in the ViewServiceConfig. Happy to change it if important but I thought it made sense anyway

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
grahamwallis: We support all 3 types

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
grahamwallis: Do you mean the omagServerName ad omagServerPlatformRootURL? They are only relevant on Solution and Vanilla configs; they are bot used at all in an Integration view service

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
grahamwallis: Perhaps - but it's another thing that the user has to set - I think I prefer it with just the class in the JSON

open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties/SolutionViewServiceConfig.java
grahamwallis: As previous comment

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
grahamwallis: As above - they are not used on an Integration view service

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
davidradl: I would not expect the user to set it - we use the name of the view service to get the registration details -where I would expect this flag to reside, most of the registration document does not need to sent in .

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
davidradl: I thought we only support the integration type and the governance solution one, I am suggesting if it is not one of them we should fail the request as it is meaningless.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminForViewServices.java
davidradl: ok makes sense.

Issue #3726, opened 2020-10-01
"saveInstanceReferenceCopies" never called in real OMRSMetadataCollection


jnhagelberg:
The problem here is that `LocalOMRSMetadataCollection` does not override `saveInstanceReferenceCopies`, so the default implementation, which splits the BATCH_INSTANCES_EVENT into individual EntityDetails and Relationships is always used.

The problem with this is that it prevents implementations from leveraging the presence of multiple messages to optimize the message processing.  They have no way of knowing that the message was part of a batch.  They still have to process the messages one at a time.  This is every slow, since the for us processing of each message results in an update to a cloudant document along with a bunch of locking.  For our use case, in most cases all of the messages in the batch end up updating the same cloudant document, so being able to process all of the messages from the batch together will give us a big performance boost.

Doing this is very straight forward.  I will be submitting a PR with the change shortly.

wbittles:
Hi Jeff, 
it would really help with a review if you could provide a bit of background context Not only will it speed up the review it will also help other users know when to switch to batch processing. 
For instance 
1) Describe the use case that generates the workload, this will help other users know what type of scenario would justify the switch to batch processing.
2)  Why is the change a performance fix, (what path lengths/ locks/copies etc)  are removed by this fix. This will give some clue  on how switching to batch processing will impact the relevant use cases and environments.

planetf1:
We're already on the final stages of shipping 2.3 which was cut nearly a week ago - pretty much in line with prior releases.
Therefore whilst I think this change has merit, I would agree with @wbittles and target master (and 2.4) only.

jnhagelberg:
@wbittles In the `cp4d` use case, we are using egeria to sync metadata between different systems.  The most typical scenario is when someone imports the metadata for bunch of relational tables and they all need to be synced over.  As I'm sure you know, it takes many entities/relationships to sync table.  There are 3 entities and 2 relationships for the table itself and about 2 entities and relationships per column.   To sync a table with 20 columns, there are about 85 entities/relationships we need to process.  We are already using the BATCH event to send all 85 entities/relationships.  However, right now egeria only provides them to us one at a time.  This results in 85 different updates to the document where we are storing the table.

However, if egeria provided us the entire batch at once, would be able to consolidate the logic, completely assemble the table in memory, and save it using just one call to cloudant.

PR #3727, opened 2020-10-01
odpi/egeria#3726 - Option to allow saveInstanceReferenceCopies() to be called in real metadata collection [master]

jnhagelberg:
See https://github.com/odpi/egeria/issues/3726

We would also like to get this into egeria 2.3, if that is still possible.  The PR for that is #3728.

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/README.md
+++ b/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/README.md
@@ -7,6 +7,85 @@
 
 The Open Lineage Janus connector allows the Open Lineages Services to connect with a JanusGraph database.
 
+The Connector can be configured to connect to either an embedded JanusGraph database or a standalone JanusGraph server. 
+For more details on possible configurations please use the documentation [offered by JanusGraph](https://docs.janusgraph.org/basics/deployment/).  
+   
+Configuring the properties of the JanusGraph client when connecting to the database should be done following the [official documentation](https://docs.janusgraph.org/basics/configuration/).
+   
+--
+- Embedded JanusGraph
+
+In order to configure the connector with an embedded JanusGraph lineageGraphConnection configuration should be configured with the following `connectorProviderClassName`:  
+
+```
+ "connectorProviderClassName" : "org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.graph.LineageGraphConnectorProvider"       
+```
+
+Example: 
+ ```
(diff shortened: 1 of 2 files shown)


jnhagelberg:
@wbittles I have pushed changes to address your review comments

planetf1:
Note there are some really useful comments and discussion in the 2.3 PR, but as 2.3 is shipping we need to focus on the fix for master/2.4 - which is this PR.

see https://github.com/odpi/egeria/pull/3728

jnhagelberg:
@wbittles @mandy-chessell Do have any further comments about this?

jnhagelberg:
@planetf1 Any comments?

mandy-chessell (changes requested):
The changes that are proposed are not what I would have expected - in fact more complex than I think is needed.  The `isAutomicallySplitBatchMessages()` method is not necessary.  All that is needed  is to implement the method in LocalOMRSMetadataCollection to pass the call through to the real metadata collection.  If the real metadata collection has not implemented the method, the default implementation in `OMRSMetadataCollection` is used which splits up the messages.  If the real metadata collection wants to override the default implementation then they can.

The administration changes are also not needed.

Finally, please format any code changes in a layout that is consistent with the rest of the module - as per our coding guidelines.

jnhagelberg:
@mandy-chessell The problem with that is that if we always forward the call to the real metadata collection we are breaking backward compatibility.  For example, our implementation does override `saveInstanceReferenceCopies()`, but the method is never being called.  Granted, we did add that method with the expectation that it would be called, but at this point because egeria is not calling it our implementation is largely untested.  However, we do have tests would catch any problems.

Having said that, it is definitely simpler to not have the toggle and associated configuration.  If you are not concerned about the backward compatibility, I can change the implementation to remove the toggle.  If we do that, though, it definitely needs to be documented as a behavior change in the release notes.

jnhagelberg:
@mandy-chessell Based in your comments I have removed the toggle.  If you decide it is still needed I can remove that commit from the PR.

mandy-chessell:
With the simple fix, the backward compatibility for the repository connector that has not overridden the method because it pre-dates the new method is fine.

The simple fix is only a problem for a repository connector that has overridden the method with an invalid (buggy) implementation.  I am not too worried about the second case as this is a repository connector that is aware of the method and has attempted to implement it.  Presumably they have tested it too and are just annoyed/puzzled why it is not being called.

jnhagelberg:
@mandy-chessell Sounds good.  In that case, can you please review the changes again and let me know if you have any further comments?

mandy-chessell (approved):

jnhagelberg:
@planetf1 Mandy approved the code review.  Can you please merge the changes if there are no further comments?

jnhagelberg:
@mandy-chessell @wbittles ^^

PR #3739, opened 2020-10-02
RELEASE 2.3 - readmes and version update - final for release

planetf1:
* Updates release version to 2.3
* Removes release note readmes (standard for release)
* Switches vdc chart to pick up 2.2 connectors (last available release)


planetf1:
Note - the gradle build error is expected (to do with how we manage versions and reuse the maven pom)

PR #3835, opened 2020-10-08
# Added hbase dependencies for janusgraph

immmus:
Hi @grahamwallis ! I added possibility of configure junusgraph for hbase storage.

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
+++ b/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/pom.xml
@@ -97,6 +97,26 @@
 
         <dependency>
             <groupId>org.janusgraph</groupId>
+            <artifactId>janusgraph-hbase</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>ch.qos.logback</groupId>
+                    <artifactId>logback-classic</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>ch.qos.logback</groupId>
+                    <artifactId>logback-core</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hbase</groupId>
+            <artifactId>hbase-shaded-client</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.janusgraph</groupId>
             <artifactId>janusgraph-driver</artifactId>
         </dependency>
 
@@ -175,6 +195,10 @@
                                 <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-cql:*
                                 </ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-cassandra:*
+                                </ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>org.janusgraph:janusgraph-hbase:*
+                                </ignoredUnusedDeclaredDependency>
+                                <ignoredUnusedDeclaredDependency>org.apache.hbase:hbase-shaded-client:*
                                 </ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>joda-time:joda-time:*</ignoredUnusedDeclaredDependency>
                                 <ignoredUnusedDeclaredDependency>org.antlr:antlr-runtime:jar:*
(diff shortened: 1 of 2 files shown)


planetf1 (changes requested):
I'm worried about the dependency chain this will cause.

Ideally our connectors are dynamically loaded -- but currently they are static dependencies, and much of our code depends on the graph repo connector.

HBase brings with it a lot of hadoop dependencies, and past experience with our atlas and ranger code is that this has been back level & exposing all our code to security issues.

The output of 'mvn dependency:tree' before and after would be interesting.

I would prefer that we see if we can allow hbase to be dynamically configured at runtime only. Additionally in a 'real' environment we'll be using janusgraph in a client/server mode with the actual storage backend being somewhere else. in that case all we need/should have locally is a thin client.

So whilst I do think it's very useful to add hbase, we need to do it in a way that doesn't impact dependencies for the rest of our code.

cc: @grahamwallis

grahamwallis:
@immus Hi Ilya - Thanks for looking at this. It would be useful to support whatever persistence backends people want to use. I assume that you are planning to use hbase for your production repositories?

I was slightly confused by the PR because it only seems to contain the pom changes. Have you added to the graph configuration code yet?

I think @planetf1 makes a very good point about being able to add backends dynamically, without introducing static dependencies. When I first wrote the graph repository connector it only had Berkeley and Lucene for persistence and indexing. Since then @lcpopa added the ability to configure Cassandra and Elastic. But we are still using the original and static approach that I used at the start. Maybe now is the time that we should try to move this to a dynamically connector based model.

As @planetf1 says, in production some organisations are likely to use a remote Janusgraph Server, and this is something that I think @lcpopa was looking at. This would offload Janusgraph and allow us to include just a client in Egeria, but I think we will still need to support embedded configurations, at least for dev with Berkeley/Lucene. I am less certain about others but happy to discuss. We all have lots to learn in this area.

@immus I'd appreciate your thoughts. 

Also @mandy-chessell from your conversations with other consumers of Egeria, do we have a feeling for what deployments we should aim to support?

immmus:
> @immus Hi Ilya - Thanks for looking at this. It would be useful to support whatever persistence backends people want to use. I assume that you are planning to use hbase for your production repositories?
> 
> I was slightly confused by the PR because it only seems to contain the pom changes. Have you added to the graph configuration code yet?
> 
> I think @planetf1 makes a very good point about being able to add backends dynamically, without introducing static dependencies. When I first wrote the graph repository connector it only had Berkeley and Lucene for persistence and indexing. Since then @lcpopa added the ability to configure Cassandra and Elastic. But we are still using the original and static approach that I used at the start. Maybe now is the time that we should try to move this to a dynamically connector based model.
> 
> As @planetf1 says, in production some organisations are likely to use a remote Janusgraph Server, and this is something that I think @lcpopa was looking at. This would offload Janusgraph and allow us to include just a client in Egeria, but I think we will still need to support embedded configurations, at least for dev with Berkeley/Lucene. I am less certain about others but happy to discuss. We all have lots to learn in this area.
> 
> @immus I'd appreciate your thoughts.
> 
> Also @mandy-chessell from your conversations with other consumers of Egeria, do we have a feeling for what deployments we should aim to support?

Hi yes we are planning to use hbase. I made this patch and already set up the graph configuration for hbase storage.
Screenshot 2020-10-09 at 11 51 01
This config is enough to connect hbase, as you can see below, tables and indexes are created, everything works fine.
Screenshot 2020-10-09 at 11 51 47
Screenshot 2020-10-09 at 11 52 36
Screenshot 2020-10-09 at 11 53 03

grahamwallis:
@immus - Thanks - that looks really good. 

@planetf1 @mandy-chessell - What's the preferred way to do this whilst avoiding static dependencies?

planetf1:
Excellent info to see hbase working with Janus & Egeria @immmus  :-)

So on dependencies - my view is that this is really a deployment option. so we could
 * At a minimum Ensure anything that explicitly refers to hbase is in specific maven artifacts, and probably a seperate part of the assembly. 
* Not including maven dependencies for hbase, use 'provided' scope and get those in place through a) admin /install docs b) Adding in via k8s/compose/ansible in any 'demo' or tutorial configurations

For Janus in particular I expect most production cases would be using in client/server mode in any case, in which case this configuration only affects the backend, so wouldn't add any client-side (ie our server/platform) dependencies?)

I think we absolutely
 - do not want hbase dependencies shipped/included in base egeria if at all possible  -- we got burnt by hadoop in the past  (on security/CVEs as well as dependency mismatches).
 - If we do include them they must be very well isolated to individual optional modules - which need to be seperated both as maven artifacts and in assemblies
 - ideally need to have a clear client/server separation (for janus in this case) for any artifacts we ship
 - allow for other tighter relationships through documentation alone

Also worth noting that our connectors are not yet pluggable - so graph repo connector gets pulled in everywhere, so including hbase is extremely undesirable

Unfortunately I don't have specifics to suggest just yet, which is why I DO NOT APPROVE the merge of this PR, though I do absolutely congratulate the effort to prove egeria working with hbase and want to see how we can capture this for more general use.

Any ideas?

PR #3846, opened 2020-10-09
Add guids as properties

yevgenmar:
Analytics module is build based on metadata stored in repository. The metadata on repository may change with time. The changes to the metadata can be traced by GUIDs which need to be part of the initial module.


open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-api/src/main/java/org/odpi/openmetadata/accessservices/analyticsmodeling/model/module/Table.java
```diff
@@ -8,6 +8,9 @@
  *	Abstraction of the table in MOSER module.
```
mandy-chessell, line 8: What is MOSER?

yevgenmar: Modeling Server web application in Cognos Analytics

open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-api/pom.xml
```diff
@@ -49,11 +49,11 @@
 			<groupId>com.fasterxml.jackson.core</groupId>
 			<artifactId>jackson-databind</artifactId>
 		</dependency>
-        <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
-            <scope>test</scope>
-        </dependency>
+		<dependency>
```
planetf1, line 52: you'll probably need to mirror this change in the build.gradle. N

planetf1 (changes requested):

open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-api/pom.xml
yevgenmar: It is just formatting, testng was not removed, it is still in gradle dependency: 
testImplementation 'org.testng:testng:7.1.0'

But I've noticed gradle and maven are using different versions of testng.
      <dependency>
        <groupId>org.testng</groupId>
        <artifactId>testng</artifactId>
        <version>7.3.0</version>
        <scope>test</scope>
      </dependency>
Should I define 7.3.0 for gradle?

planetf1: Hi, After a merge to master on Friday, the gradle versions are now all consolidated and can be found in the top level build.gradle (search for 'constraints'). That is now version 7.3

The versions specified in individual files have been removed -- it was never the right approach, but the first step of the migration (maven->gradle tool created them that way )

yevgenmar: Then it should be good.

mandy-chessell (approved):
I am going to approve this since Nigel's comment's seem to be addressed

mandy-chessell:
recheck

Issue #3836, opened 2020-10-08
Admin Services API for View Services


grahamwallis:
I would have assigned this issue to myself, Mandy and David but for some reason I seem to be able to either assign it to myself (only) or to Mandy and David (only). I am happy to 'own' it, but it needs a conversation first, so assigning to Mandy and David so it gets their attention.

I have been extending the Admin Guide for the administration of view services and noticed the following:

1. There is no API for retrieving the configuration of an individual view service. This seems like an obvious operation to include and should have the following REST interface:
```
GET {serverURLRoot}/open-metadata/admin-services/users/{adminUserId}/servers/{serverName}/view-services/{serviceURLMarker}
```
The response should be a ViewServiceConfigResponse containing a ViewServiceConfig object.

2. There is no API for clearing an individual view service. Again, this seems like an obvious operation to include and should have the following REST interface:
```
DELETE {serverURLRoot}/open-metadata/admin-services/users/{adminUserId}/servers/{serverName}/view-services/{serviceURLMarker}
```

3. There is an API for enabling all configured view services. However, it currently is coded to accept a RequestBody consisting of a single ViewServiceConfig. This is analogous to how the corresponding operation is designed in the Access Services Admin API, which accepts a map of accessServiceOptions (which may or may not be sufficient, but probably doesn't matter as I think the access services will be enabled with default settings). However, for 'enable all' operation for view services, it does not seem to me that it's possible to adopt default values. Each Solution View Service will require an `omagServerName` and `omagServerPlatformURL` which *could* be defaulted, but it seems very likely that different OMASs would be on different platforms/servers so this seems an odd approach. Furthermore, each Integration View Service will require a set of resource endpoints for which there is no practical default value, other than to make it empty, which would render view services that are not usable. In my opinion I think this operation should probably be removed from the View Services Admin API.

I'd appreciate comments and discussion on the above points. In the description in the Admin Guide I have, for the time being, assumed that the first two operations will be added, and that the third will be removed. If we decide that this is the most desirable outcome I am happy to make the changes to the API. If not I will change the descriptions in the documentation, which I will not check in yet.

davidradl:
Point 1 looks useful, though not in keeping with the style of the other admin calls, which do not return what has been created or updated in their responses.    
Point 2 seems reasonable.
[...]

Issue #3708, opened 2020-09-29
View server does not shutdown when requested


planetf1:
I started 'cocoView1' - a view server with subject area & Dino configured. The configuration was not as I wanted, so I tried to shut it down - but got an error ie:

```
 
Stopping server cocoView1 ...
DELETE https://datalake:9443/open-metadata/admin-services/users/garygeeke/servers/cocoView1/instance
Response: 
{
    "class": "VoidResponse",
    "relatedHTTPCode": 500,
    "exceptionClassName": "org.odpi.openmetadata.adminservices.ffdc.exception.OMAGConfigurationErrorException",
    "exceptionCausedBy": "org.odpi.openmetadata.commonservices.ffdc.exceptions.PropertyServerException",
    "actionDescription": "deactivateTemporarily",
    "exceptionErrorMessage": "OMAG-ADMIN-500-001 Method deactivateTemporarily for OMAG server cocoView1 returned an unexpected exception of org.odpi.openmetadata.commonservices.ffdc.exceptions.PropertyServerException with message OMAG-MULTI-TENANT-400-002 The OMAG server cocoView1 has been requested to shutdown but the following services are still running: []",
    "exceptionErrorMessageId": "OMAG-ADMIN-500-001",
    "exceptionErrorMessageParameters": [
        "cocoView1",
        "deactivateTemporarily",
        "org.odpi.openmetadata.commonservices.ffdc.exceptions.PropertyServerException",
        "OMAG-MULTI-TENANT-400-002 The OMAG server cocoView1 has been requested to shutdown but the following services are still running: []"
    ],
    "exceptionSystemAction": "The system is unable to configure the OMAG server.  No change was made to the server's configuration document.",
    "exceptionUserAction": "This is likely to be either an operational or logic error. Look for other errors.  Validate the request.  If you are stuck, raise an issue."
}
```
It reports services are running -- quite possibly this relates to the two view services that are active (SA/Dino) - however the exception is not reporting any running services - leaving the user confused.

This is probably a bug in error handling and it is likely correct not to stop the server

The current config doc for this server is:

(cannot be retrieved)

It seems as if after this failure the server still shuts down anyway, with an attempt to delete again then returning

```
Stopping server cocoView1 ...
DELETE https://datalake:9443/open-metadata/admin-services/users/garygeeke/servers/cocoView1/instance
Response: 
{
    "class": "VoidResponse",
    "relatedHTTPCode": 500,
[...]

PR #3848, opened 2020-10-09
Updates to view service administration API

grahamwallis:
This PR implements GET of an the config for any individual view service and DELETE of an individual view-service.

It also corrects a number of errors found along the way.

With regard to the last point in issue #3836, which is about enabling of all view-services with a common set of configuration properties or defaults, this PR does NOT implement that. It removes that method from the API since it is not clear how it will work. If we reach a design for it, it can be implemented in a subsequent PR.


planetf1:
will it help with #3708 ?

grahamwallis:
@plaetf1 - "will it help with #3708?" - I don't think so - I think that is being caused by something else, as noted in comments on #3708.

grahamwallis:
Fixes #3848

Issue #3849, opened 2020-10-09
After recent presentation server UI refactor, package-lock is left behind


planetf1:
After building master today (HEAD is at 44c4be42cbd6fff1795ed13529786a1055b0614c ), a package-lock.json is left
as a pending git change

```
➜  charts git:(master) ✗ git status                                                                                                                    <<<
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   ../../../open-metadata-implementation/user-interfaces/presentation-server/nodejs/package-lock.json

no changes added to commit (use "git add" and/or "git commit -a")
```

Either this file should indeed be checked in, or it should be excluded, or any custom handling needs to be
encoded in either the webpack build, or surrounding maven project

PR #3855, opened 2020-10-09
#3849 restore old gitignore line for nodejs client

davidradl:


planetf1 (approved):
I noticed a regression with package lock. Not used vscode but agree you with this now and can consolidate later (ie vscode at top level)

PR #3891, opened 2020-10-15
Add files via upload

lizsping:
Proposed story board for Callie Quartile


planetf1:
Thanks for contributing the content..

There's quite a few scenario documents for coco pharmeuticals at https://github.com/odpi/data-governance/tree/master/docs/coco-pharmaceuticals 

Is this a start of a series of different scenario docs? Should they extend the above? Or do you think they are closer to egeria itself? If so I think we need to choose a new location other than the developer guidelines?

Is this a proposal for a new feature? Or is it trying to capture how you see egeria working today? Most likely how you propose the UI interaction should work?

On the doc itself it's probably worth running the spell check as I noticed some typos, stray bullets, capitalization - but happy to mark up if it helps when I understand the context ?

davidradl:
@lizsping I agree with Nigels comments, the file should probably live where Nigel points to. Also

- The file should be draw.io or markdown and pdf , so others can update it. 
- the title of the file is Gary Geeke but the content refers to Callie Quartile.
- I think we need a copyright on docs.

planetf1:
As there's no further updates here I am going to suggest we close this PR
@lizsping if you are able to update please do reopen the PR and continue the discussion

Issue #3884, opened 2020-10-14
Enable static content server for polymer UI in containers


planetf1:
The static content from the polymer based UI under https://github.com/odpi/egeria/tree/master/open-metadata-implementation/user-interfaces/ui-chassis is now in the egeria UI project https://github.com/odpi/egeria-ui and needs to be served via a web server

This issue is being opened to add deployment of this within the egeria lab image

c: @sarbull 

- [x] Add new egeria docker image for nginx 
- [x] Add egeria ui content for static serving
- [x] Add new image to odpi-egeria-lab helm chart
- [x] Add configuration to link ui server chassis with static server
- [x] Add new image to docker compose lab environement
- [x] Add configuration in docker compose
- [x] Update docker hub information for new image
- [x] update any additional docs necessary
- [ ] Enable https
- [ ] check image security

To consider later but excluded for now
 - switching to a redhat UBI based image for nginx
 - making use of egeria ui artifact from npm

Included elsewhere
 - Further docs in egeria server chassis & Admin guide on configuring the ui server chassis manually (cc: @sarbull )

planetf1:
No further changes will be made in the egeria repo. Opened https://github.com/odpi/egeria-ui/issues/18 to migrate the build of this container over to the repo that hosts the static content - see https://github.com/odpi/egeria-ui/issues/18

PR #3893, opened 2020-10-15
Add static content server to fix ui server chassis in k8s environment

planetf1:
* Adds new docker image (egeria-static UI) for static content serving
 - Currently build as part of egeria
 - does a git pull from egeria-ui
 
This is a tactical approach - we really need to create a package and/or docker image within the egeria-ui project, but this will
be dependent on adding additional pipelines, so for immediate benefit here we have something that at least works

* Adds static content server to k8s chart

Adds a new deployment for the above within our lab environment to act as an example

* Configures egeria ui 
 Adds in reference to static content server so requests routed successfully

docker-compose - still to do/evaluate
docs
 .. may be in seperate PRs so we can validate the end to end process including build first

See #3884


open-metadata-resources/open-metadata-deployment/docker/egeria-uistatic/pom.xml
```diff
@@ -0,0 +1,146 @@
...
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+
+</project>
```
sarbull, line 146: @planetf1 why is maven needed for this?

planetf1: Only because that is how we currently drive all our docker builds, and we already have the build pipelines setup. In some of the images we also have to retrieve files from other maven artifacts in the build, which the maven resources plugin manages.

There is no intrinsic need to use maven to build docker, and indeed it adds extra complexity.

I wanted to get something in and working quickly ... so to to stabilize master. Then work on improvements. One of those is that I think the build of the image should be done from sources within the egeria-ui project -- and we can have a simple build pipeline potentially that just uses the Dockerfile directly. That way we can easily trigger off changes and keep the image updated.

sarbull: i see, i have no problem with that, agree. i was imagining that the Dockerfile would be in the `odpi/egeria-ui` repository, but for the quickest way of stabilizing the master looks awesome so far, great work!

planetf1:
Now working after a fix to the zuul proxy url.

However on openshift the security policy must allow 'setgid() and setuid()' to work -- for future investigation

planetf1:
This now works in docker-compose too.

Propose we merge this and then reflect/review

open-metadata-resources/open-metadata-deployment/compose/tutorials/.env
```diff
@@ -4,7 +4,7 @@
 # Version of egeria to use
 egeria_version=2.4-SNAPSHOT
 # Override if you wish to use images other than from docker hub odpi/egeria etc
-egeria_repo=odpi
+egeria_repo=planetf1
```
sarbull, line 7: is this ok?

open-metadata-resources/open-metadata-deployment/docker/egeria-uistatic/dist/nginx.conf
```diff
@@ -0,0 +1,15 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Contributors to the Egeria project
+
+server {
+    listen 80;
```
sarbull, line 5: should have the HTTPS port also

planetf1: Agreed should look at https. I'll get this in & then let's catch changes in another iteration through? Means anyone using the charts etc in master can be working again whilst we refine. I think that nginx config needs more refinement too

Issue #2099, opened 2019-11-26
TypeDefAttribute name clashes with core properties


grahamwallis:
There are a number of types that have type-defined attributes whose names clash with core properties in the InstanceAuditHeader.

They are:
- [x] SoftwareServerPlatform (type, version)
- [x] SoftwareServer (type, version)
- [x] SoftwareServerCapability (type, version)
- [x] CloudPlatform (type)
- [x] CloudTenant (type)
- [x] CloudService (type)
- [x] Comment (type)
- [x] GraphStore (type)
- [x] LogFile (type)
- [x] Database (type, version)
- [x] DatabaseServer (type, version)
- [x] MetadataRepository (type)
- [x] MetadataServer (type)
- [x] RepositoryProxy (type)
- [x] BusinessCapability (type)
- [x] ~~SchemaAttributeType (type)~~ -- this is a relationship endpoint, so not a conflict (leaving as-is)
- [x] DataStore (createTime)
- [x] DataSourcePhysicalStatusAnnotation (createTime)
- [x] EnterpriseAccessLayer (metadataCollectionId)
- [x] MetadataCollection (metadataCollectionId)
- [x] ExternalReference (version)
- [x] PropertyFacet (version)
- [x] CohortMember (version)
- [x] ImplementationSnippet (version)
- [x] PolicyAdministrationPoint (type) (1.7)
- [x] PolicyDecisionPoint (type) (1.7)
- [x] PolicyEnforcementPoint (type) (1.7)
- [x] PolicyInformationPoint (type) (1.7)
- [x] PolicyRetrievalPoint (type) (1.7)

For the time being core property search is disabled in graph repository to work around the above. When the attributes are renamed,, core property search should be re-enabled.

(CG edit: after much head-scratching on debugging a problem, updated the list above to a complete list -- note in particular that `type` clashes with the name of a property in InstanceAuditHeader, and appears in many other open metadata types... Would it be easier to update the InstanceAuditHeader property to be `instanceType` instead of `type` (or would this break backwards-compatibility?))

grahamwallis:
This issue is waiting for the type def patch capability to be implemented

grahamwallis:
Keep open

cmgrote:
Agreed on August 4th workshop to rename the properties of the open types themselves (deprecate existing, add a new property) -- do not rename any of the header properties.

cmgrote:
Properties are all deprecated and added in the open types archive itself; still to do:

[...]

PR #3919, opened 2020-10-20
Replaces control property clashes in open types (#2099)

cmgrote:
Should resolve the property name clashes as documented in #2099, including:

- OpenMetadataTypesArchive updates marking each original property as deprecated, with `replacedBy` pointers to each new property name
- Updating all the diagrams to reflect the new property names (including creating draw.io tabs for any that had not yet been migrated from Omni)
- Saving new PNGs for each changed diagram, reflecting the new property names
- Updating the OpenMetadataAPIMapper class with the new property names
- Adding the deprecated property names to OpenMetadataAPIMapper where necessary **
- Updating anywhere the OpenMetadataAPIMapper's constants were used elsewhere (OCF, OMAS, etc) with deprecated property name handling, where necessary **

** I took "necessary" to mean any place where the old property names were used to do some action _other than_ adding a new InstanceProperty/Value. (Since those are write actions, I don't think we need to handle writing values to both the new property name and deprecated property names?)  But anywhere a removal occurs, I remove both the new property name's and the deprecated property name's values, and return the non-null one (preferring the new property name's value, but if null will return the old property name's value, if any).


mandy-chessell (approved):
This looks good - there may be regressions in modules that do not use the generic handlers

PR #3939, opened 2020-10-23
Move java 11 + gradle build verifications to github actions

planetf1:

Adds the gradle build as a github action rather than a Azure dev pipelines build


planetf1:
See #3940

planetf1:
Also moves the java 11 build verification to github actions.

Both these two builds are only used to ensure build & tests run. We do not save any artifacts from the test runs ie in artifactory, nor publish any maven artifacts/docker images. No additional credentials etc are needed, making them an easy option to move. See issue for more details

Issue #3938, opened 2020-10-23
NPEs in LocalOMRSMetadataCollection


grahamwallis:
There are a number of NPEs seen when running the CTS.

grahamwallis:
Fixed by #3946

PR #3946, opened 2020-10-23
Adds protection to various methods in LocalOMRSMetadataCollection

grahamwallis:

The changes here are to avoid NPEs that were occurring during CTS runs and increase consistency of null checking.


grahamwallis:
Fixes #3938

Issue #3758, opened 2020-10-06
type-explorer is broken in coco pharma notebook environment


planetf1:
A change in 2.4 has broken the configuration of type explorer

  ... configuring the type-explorer Governance Solution View Service for this server...
   ...... (POST https://dev:9443/open-metadata/admin-services/users/garygeeke/servers/cocoView1/view-services/type-explorer )
   ...... Response:  {'class': 'VoidResponse', 'relatedHTTPCode': 400, 'exceptionClassName': 'org.odpi.openmetadata.adminservices.ffdc.exception.OMAGConfigurationErrorException', 'actionDescription': 'configureViewService', 'exceptionErrorMessage': 'OMAG-ADMIN-400-036 Unable to configure server cocoView1 since view service type-explorer is not registered in this OMAG Server Platform', 'exceptionErrorMessageId': 'OMAG-ADMIN-400-036', 'exceptionErrorMessageParameters': ['cocoView1', 'type-explorer'], 'exceptionSystemAction': "The system is unable to add this view service to the server's configuration document.", 'exceptionUserAction': 'Check that the name of the view service is correctly specified in the configuration request.  If you are not sure, issue the call to list the registered view services and verify the values you are using.  If the name is right, but the view service should be registered,then the developer of the view service needs to add this registration to the code of the view service. An view service is registered in the OMAG Server Platform by adding a description of the view service to the view service registration. Once the view service being requested is registered, retry the configuration request.'}

Suspect this is relating to the name change.

cc: @grahamwallis

planetf1:
Caused by #3736 I think

Let me know @grahamwallis if you want to do this or I will -- and if there are any other relevant changes? Would be good to keep this working in master. (and consider automation of the testing)

grahamwallis:
Hi @planetf1 - Yes please if you wouldn't mind investigating and correcting that would be great. I'm currently doing similar updates to Rex and it will be good to get all the view service changes in and then make sure the admin services and notebooks are all consistent. 

Apart from the obvious failure in the configuration attempt above, it's worth mentioning that Tex is not a [Governance] Solution View Service - it's an IntegrationViewService. Not sure if this is just a copy/paste in the notebook, or whether it is a little deeper.

planetf1:
Good to clarify the type -- the current implementation was rather pressed for time to get something out, and the configuration initially was the same as a Governance Solution View service, which is why it got added to that list. I can correct.

planetf1:
[...]

PR #3952, opened 2020-10-23
#3758 update config for rex

planetf1:

Updates the lab notebook (config) to work with endpoints for all of rex, tex and dino
cc: @grahamwallis

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/graphrepository/repositoryconnector/GraphOMRSGraphFactory.java
+++ b/open-metadata-implementation/adapters/open-connectors/repository-services-connectors/open-metadata-collection-store-connectors/graph-repository-connector/src/main/java/org/odpi/openmetadata/adapters/repositoryservices/graphrepository/repositoryconnector/GraphOMRSGraphFactory.java
@@ -902,9 +902,8 @@ public class GraphOMRSGraphFactory {
                                                                                                                   thisMetadataCollectionId));
 
             throw new RepositoryErrorException(
-                    GraphOMRSErrorCode.GRAPH_DB_HAS_DIFFERENT_METADATACOLLECTION_ID.getMessageDefinition(methodName,
-                                                                                                         GraphOMRSGraphFactory.class.getName(),
-                                                                                                         thisRepositoryName),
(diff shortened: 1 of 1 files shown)


open-metadata-resources/open-metadata-labs/egeria-server-config.ipynb
```diff
@@ -892,9 +892,89 @@
...
     "print (\"Configuring the Integration View Services\")\n",
     "\n",
-    "# For now rex (repository explorer) follows the same pattern as glossary author, but this will change to be more similar to dino & tex\n",
+    "# repository explorer integration view service\n",
     "viewService=\"rex\"\n",
-    "configureGovernanceSolutionViewService(adminPlatformURL, adminUserId, viewServerName, viewService, remotePlatformURL,remoteServerName)\n",
+    "rexConfigBody = {\n",
+    "            \"class\":\"IntegrationViewServiceConfig\",\n",
+    "            \"viewServiceAdminClass\":\"org.odpi.openmetadata.viewservices.rex.admin.RexViewAdmin\",\n",
+    "            \"viewServiceFullName\":\"Type Explorer\",\n",
```
grahamwallis, line 900: This should be Repository Explorer

open-metadata-resources/open-metadata-labs/egeria-server-config.ipynb
```diff
@@ -892,9 +892,89 @@
...
+    "                    \"class\"              : \"ResourceEndpointConfig\",\n",
+    "                    \"resourceCategory\"   : \"Platform\",\n",
+    "                    \"description\"        : \"Development Platform\",\n",
+    "                    \"platformName\"       : \"Development\",\n",
+    "                    \"platformRootURL\"    : devPlatformURL\n",
+    "                },\n",
+    "                {\n",
+    "                    \"class\"              : \"ResourceEndpointConfig\",\n",
+    "                    \"resourceCategory\"   : \"Server\",\n",
+    "                    \"serverInstanceName\" : \"cocoMDS1\",\n",
```
grahamwallis, line 929: Whilst this configuration is valid - I wonder whether it might be more illustrative if we distinguish the serviceInstanceName from the serviceName. If there is only ever one instance of the service then it genuinely makes no difference and is arguably clearer the way it is here (with the two fields having the same value). If there *might* be more than one instance then reflecting that in the serviceInstanceName is clearer.

grahamwallis:
@planetf1 Many thanks for making these changes. Of my review comments the first one is important, the second one is optional. Happy to chat through them.

planetf1:
> @planetf1 Many thanks for making these changes. Of my review comments the first one is important, the second one is optional. Happy to chat through them.

Added #3961

PR #3963, opened 2020-10-26
Catch and report failures to initialise the graph database

grahamwallis:


planetf1:
I missed this before branching for 2.4. I can cherrypick the fix once you are ready and add to 2.4

PR #3966, opened 2020-10-27
Initial draft of Asset Manager OMAS and Catalog Integrator OMIS API

mandy-chessell:
This pull request is in draft status to allow early review of the new APIs.
There are also changes to the Integration Deamon's configuration to control the permitted directions of synchronization.


mandy-chessell:
Looks like the last merge of master completely messed up this PR

open-metadata-resources/open-metadata-archives/open-metadata-types/src/main/java/org/odpi/openmetadata/opentypes/OpenMetadataTypesArchive1_2.java
```diff
@@ -1415,7 +1415,7 @@ private RelationshipDef getExternalIdLinkRelationship()
          * Set up end 1.
          */
         final String                     end1EntityType               = "Referenceable";
-        final String                     end1AttributeName            = "resource";
+        final String                     end1AttributeName            = "resources";
```
cmgrote, line 1418: Are we allowed to change the name of an attribute in a previous metadata type archive (1.2 in this file)?  Wouldn't this break backwards compatibility?

mandy-chessell: No - because it is documentation - not used in the code ....

PR #3980, opened 2020-10-27
REST API parameters validation;

yevgenmar:
add function description in swagger;
change dataSource to database in api


open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-server/src/main/java/org/odpi/openmetadata/accessservices/analyticsmodeling/server/AnalyticsModelingRestServices.java
```diff
@@ -58,41 +65,48 @@ public AnalyticsModelingOMASAPIResponse getDatabases(String serverName, String u
 		RESTCallToken token = restCallLogger.logRESTCall(serverName, userId, methodName);
 
 		try {
+			validateUrlParameters(serverName, userId, null, null, startFrom, pageSize, methodName);
+			
 			DatabasesResponse response = new DatabasesResponse();
 			List<ResponseContainerDatabase> databases = getHandler()
 					.getDatabaseContextHandler(serverName, userId, methodName).getDatabases(startFrom, pageSize);
```
davidradl, line 72: I notice that in getDatabases it has 

"
getDatabases(Integer startFrom, Integer pageSize) 
			throws AnalyticsModelingCheckedException {
		
		setContext("getDatabases");

		InstanceProperties instanceProperties = omEntityDao.buildMatchingInstanceProperties(Collections.emptyMap(), true);
		List<EntityDetail> entities = omEntityDao.findEntities(instanceProperties, Constants.DATABASE, 0, 0);
"

This does not look right:
- the supplied startFrom and pagesize is being ignored
- it is hard coding startFrom and pageSize to 0. 0 is a goof default for startFrom but the default pageSize should come from the server config maxPageSize

davidradl (commented):
I think change is fine, but as per the comment the getDatabases does not seem to honour the requested paging - I suggest reviewing the other calls are honouring paging.  

You may want a standard method to default startFrom and pageSize if they are not on the  request

open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-server/src/main/java/org/odpi/openmetadata/accessservices/analyticsmodeling/server/AnalyticsModelingRestServices.java
yevgenmar: This is being refactored using RelationalDataHandler where start index and page size pushed into find function.
In the current implementation, the function retrieved all databases without paging, sorted them by names, and returned page of the sorted set. See last two lines.

davidradl (approved):
OK it sounds like this is in hand

PR #4002, opened 2020-10-28
New Request to merge Server Author work with signed commits

dkneisly:
No code changes, just a new branch with new commits.

### DIFF
--- a/open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-api/src/main/java/org/odpi/openmetadata/accessservices/analyticsmodeling/ffdc/AnalyticsModelingErrorCode.java
+++ b/open-metadata-implementation/access-services/analytics-modeling/analytics-modeling-api/src/main/java/org/odpi/openmetadata/accessservices/analyticsmodeling/ffdc/AnalyticsModelingErrorCode.java
@@ -109,6 +109,11 @@ public enum AnalyticsModelingErrorCode implements ExceptionMessageSet {
 			"The request parameter {0} has invalid value.",
 			"The system is unable to process the request.",
             "Verify parameter value."),
+	FAILED_FETCH_DATABASES(
+			"OMAS-ANALYTICS-MODELING-016", 
+			"Databases could not be fetched.",
+			"The system is unable to process the request.", 
+			"Review the exception to resolve the issue."),
 
 	UNKNOWN_ERROR(
 			"OMAS-ANALYTICS-MODELING-500", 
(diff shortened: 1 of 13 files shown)


davidradl (changes requested):
I think this code needs to go into the jsonjs source folder- as this is the live up to date source. 
I do not think that we should add this enhancement  into the back level cra folders, leaving  nodejs folder with a back level version of server author.

dkneisly:
I thought we were headed toward using the CRA version, but you sound hesitant to rip the band-aid.. we can chat about this on Monday.

davidradl:
> I thought we were headed toward using the CRA version, but you sound hesitant to rip the band-aid.. we can chat about this on Monday.

The rest of the Ui is being updated in the other folder, until we move over to the cra folders - I think we need the latest version of everything in the existing folder - so this is the latest golden truth. When we come to make the cra folders live, we could move the nodejs content to the cra folders.does that make sense

planetf1:
@davidradl remember that if you change the folders the maven build needs to be updated (fairly obvious)
But also open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/egeria-presentation-server.xml
as this defines the source for the built UI files that are used to deploy the server via a container.

I think it does make sense ultimately to move the nodejs code to a new repo, but for now the minimum is to update this definition.

An intermediate, perhaps better step, would be to package all the nodejs build output into the maven component correctly, and then extract from here when building the assembly. I skipped this initially as I wasn't convinced maven was a good fit.

You can test any change by inspecting the assembly from a full build & running the docker container. 
Without this it will break the tutorial environments , and unfortunately, currently, we don't have enough automated validation

davidradl:
> @davidradl remember that if you change the folders the maven build needs to be updated (fairly obvious)
> But also open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/egeria-presentation-server.xml
> as this defines the source for the built UI files that are used to deploy the server via a container.
> 
> I think it does make sense ultimately to move the nodejs code to a new repo, but for now the minimum is to update this definition.
> 
> An intermediate, perhaps better step, would be to package all the nodejs build output into the maven component correctly, and then extract from here when building the assembly. I skipped this initially as I wasn't convinced maven was a good fit.
> 
> You can test any change by inspecting the assembly from a full build & running the docker container.
> Without this it will break the tutorial environments , and unfortunately, currently, we don't have enough automated validation at the PR level. Worse a breakage will stop all our merge builds, docker builds, security scans, and release pipeline from working

@planetf1 thanks Nigel, the server side code has been refactored in the cra folder and new function added to the cra client  folder. So the nodejs latest client dino, text rex and glossary author code needs to be copied over to  the cra folders to join the latest server author client code and the server code and as you say the build needs to updated and the Docker file. I hope to coordinate these changes.

planetf1:
Ok. I'll keep an eye on the prs and let me know if you have questions about the assemblies or images

Issue #3725, opened 2020-10-01
Dino hits exception querying server config


planetf1:
* Deploy the coco pharmaceuticals notebook environment & run the configure/start notebooks
* Login to the presentation server UI with garygeeke/admin
* click on cocoMDS3
* expand 'server configuration'

Each time this is done I get a UI error:

Unexpected response from callPOST : {}

Meanwhile the node.js app reports debugging info:
```
before /servers/coco/dino/users/garygeeke/server/cocoMDS3/stored-and-active-configuration
segment1 servers
after /servers/coco/dino/users/garygeeke/server/cocoMDS3/stored-and-active-configuration
deserializeUser called with id 6
findByUserId
passport.deserializeUser user is [object Object],err isnull
/servers/* post called /servers/coco/dino/users/garygeeke/server/cocoMDS3/stored-and-active-configuration
Got body: { serverName: 'cocoMDS3', platformName: 'Core' }
downstream url https://lab-datalake:9443/servers/cocoView1/open-metadata/view-services/dino/users/garygeeke/server/cocoMDS3/stored-and-active-configuration
response.data
{
  class: 'DinoServerDoubleConfigResponse',
  relatedHTTPCode: 200,
  requestSummary: {
    platformName: 'Core',
    operation: 'serverGetInstanceConfiguration',
    serverName: 'cocoMDS3'
  },
  storedConfig: {
    class: 'OMAGServerConfig',
    versionId: 'V2.0',
    localServerId: 'de3a4b60-8aa9-44ae-8f57-f5451872ebff',
    localServerName: 'cocoMDS3',
    localServerType: 'Metadata Server',
    organizationName: 'Coco Pharmaceuticals',
    localServerURL: 'https://lab-core:9443',
    localServerUserId: 'cocoMDS3npa',
    localServerPassword: 'cocoMDS3passw0rd',
    maxPageSize: 100,
    serverSecurityConnection: { class: 'Connection', connectorType: [Object] },
    eventBusConfig: {
      class: 'EventBusConfig',
      topicURLRoot: 'egeria.omag',
      configurationProperties: [Object],
      additionalProperties: [Object]
    },
    accessServicesConfig: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ],
    repositoryServicesConfig: {
      class: 'RepositoryServicesConfig',
      auditLogConnections: [Array],
      localRepositoryConfig: [Object],
[...]

PR #4026, opened 2020-10-30
Improvements to Dino (issue 3725)

grahamwallis:
Dino now tolerates servers that have partially complete cohort information or configurations.

It also is better at handling configured service options with first class treatment for the known uses (e.g. supported or default zones, karma points) and a fallback for options that are merely presented as 'objects', with overspill detection and delegation to portal.

Server status history is moved to the portal and accessed via button nested under top level heading.

Server audit trail is nested into a section with top level heading.

I believe these changes will prevent the issue that was originally observed in #3725. If additional guards are required they can be added in a similar manner to the changes in this PR.

### DIFF
--- a/open-metadata-implementation/user-interfaces/presentation-server/nodejs/package-lock.json
+++ b/open-metadata-implementation/user-interfaces/presentation-server/nodejs/package-lock.json
@@ -1292,6 +1292,21 @@
         "lodash-es": "^4.17.15"
       }
     },
+    "@sindresorhus/is": {
+      "version": "0.14.0",
+      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
+      "integrity": "sha512-9NET910DNaIPngYnLLPeg+Ogzqsi9uM4mSboU5y6p8S5DzMTVEsJZrawi+BoDNUVBa2DhJqQYUFvMDfgU062LQ==",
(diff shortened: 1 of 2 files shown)


grahamwallis:
@plaetf1 - "..why would dino be seeing partially incomplete configuration data... Does it imply an issue with the view service/back-end ?"

No not at all - it was problems only in Dino. The problem that was causing the failures you reported was that Dino was not able to handle the accessServiceOptions in the configuration of cocoMDS3. They are a little nasty as it is just a Map<String,Object> so it is not immediately obvious how to format whatever we might find in there. That is now fixed and Dino has first-class treatment for the options uses that we know of - things like supportedZones, defaultZones, karma point plateau, and has a fallback option for things that arrive out of the blue as arbitrary objects, even if they are very large.
I also found another problem - when I tested with cocMDS3 spun up outside of the k8s environment I got an additional failure caused by cocoMDS3 being the only server in its cohort, so there was a gap where Dino was expecting to find remote registrations. Also now fixed.

planetf1:
Suggest we merge, and fix any subsequent issues in a new issue (have had some problems building locally). Will test once merge build updates docker images.

planetf1 (approved):

planetf1:
The fix looks good. I've not been able to reproduce the error. (tested on docker image)

Issue #3971, opened 2020-10-27
Create release 2.4


planetf1:
Create 2.4 release (agreed on status call 27 Oct)

- [x] Create branch for 2.4
- [x] Reassign any issues not being worked on to the next release
- [x] Update master to **2.5-SNAPSHOT** 
- [x] Ensure any remaining fixes are merged 
- [x] Update branch version from 2.4-SNAPSHOT to 2.5
- [x] Remove release notes from branch 
- [x] update release notes & rename 
- [x] Verify notebooks
- [x] Check polymer UI 
- [x] Check node (ps) UI 
- [x] Run CTS 
- [x] Run FVTs (this is only FVTs automated in the build process)
- [x] Check samples #4056
- [x] Check swagger doc 
- [x] Verify docker image/charts/compose 
- [x] Launch release pipeline 
- [x] Approve release
- [x] Verify artifacts all on maven central 
- [x] Create GitHub release
- [x] ensure any fixes ported to master

planetf1:
Added fixes for open lineage & graph repo to issue (see PR)

planetf1:
CTS (inmemory) completes successfully with no errors.
Graph testing - tbd

planetf1:
Likely minor issue noted with one of the governance zone samples - see https://github.com/odpi/egeria/issues/4056

planetf1:
Graph cts test complete - working

planetf1:
After a few iterations on publishing from the release pipeline, it is currently failing due to
```
Method not allowed during maintenance, Could not sync artifact open-lineage-services-spring-2.4-javadoc.jar.asc. Server response:\n Method not allowed during maintenance, 
````
Failing on many artifacts.

Unfortunately this seems to occur every release - at all times of day.

planetf1:
A few modules are failing - see #3606 (analytics-modeling-client), #3750 (presentation-server)

planetf1:
Notes added on #3194 to improve reliability -- these issues are now impacting every release cycle, and probably adding a few days of work per month.

planetf1:
Continuing to retry the final publishing step which is failing with intermittent errors. Most artifacts now published & all essential artifacts look to pass validation

planetf1:
Release is now complete and published

PR #4058, opened 2020-11-02
#3971 update release notes for 2.4

planetf1:

Updates release notes for 2.4

Please review/comment & suggest any updates or push to this PR branch.

I would like to get an update out tomorrow since the build is now publishing. However we can make further incremental updates after that as the release notes are in master.

Reviewers have added 1 or more commits in this release.


sarbull:
We should also mention the UI extraction.

planetf1:
> We should also mention the UI extraction.

Agreed - thanks for spotting

planetf1:
@sarbull - I've added a small section on the static UI. Any good?

grahamwallis:
I wonder whether we should say "The Repository explorer User Interface ('REX') must now be configured with a list of available platforms and servers..." rather than saying that it "can" be. We could add that this is to improve usability and security.

release-notes/release-notes-2-4.md
```diff
@@ -1,29 +1,33 @@
...
 
-# Release 2.4 (Planned October 2020)
+# Release 2.4 (October 2020)
 
-Release 2.4 adds support for:
-   * Project management of IT change
-   * IoT and large scale cohorts
+* The [Integration Daemon](https://github.com/odpi/egeria/blob/master/open-metadata-implementation/admin-services/docs/concepts/integration-daemon.md) now makes is simpler to exchange metadata with third party technology 
+  such as catalogs and databases.
+* The Repository explorer User Interface ('REX') can now be configured with a list of available
```
grahamwallis, line 8: Need to be explicit that this is the presentation-server implementation of Rex.

release-notes/release-notes-2-4.md
```diff
@@ -1,29 +1,33 @@
...
+  such as catalogs and databases.
+* The Repository explorer User Interface ('REX') can now be configured with a list of available 
+  platforms and servers to improve security and usability. 
+* The latest version of Repository Explorer, Type Explorer and Dino, is found 
+  in the 'Presentation Server' UI. See 
+  [Presentation Server component documentation](https://github.com/odpi/egeria/tree/master/open-metadata-implementation/user-interfaces/presentation-server) 
+  and 
+  [Configuring the Presentation Server](https://github.com/odpi/egeria/blob/master/open-metadata-implementation/admin-services/docs/user/configuring-the-presentation-server.md)
+  .
+* The [Business UI ](https://github.com/odpi/egeria/tree/master/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring)
```
davidradl, line 16: I do not think Business UI is the right term for this . The Glossary Author is a "Business UI" in the Presentation server. I suggest Web Component UI

davidradl (changes requested):

release-notes/release-notes-2-4.md
planetf1: agreed - switched it around a bit (and can->must)

release-notes/release-notes-2-4.md
planetf1: I'm going to counter propose staying with the 'UI Chassis'. I agree Business UI is confusing for end user, but Web Component UI could be either - what is it.

We need to clarify the role of the UIs, but that's a longer task. For now I think staying with UI chassis makes sense, as that at least is already used, linked, and referred to in the broader docs & code.

planetf1:
I plan to merge this once built. If we have followup updates please raise a new issue.
The release artifacts are going out so I want to ensure anyone hitting the link sees reasonable information even if imperfect.

PR #4060, opened 2020-11-03
[UI] remove subject-area dependency 

bogdan-sava:

Issue #3994, opened 2020-10-28
Latest Spring-boot >= 2.3.4 update causes SubjectArea initialization failure (reflection)


planetf1:
Using spring-boot 2.3.4 & spring 5.3.0 egeria will fail to start with many errors relating to 'java.lang.IllegalStateException: zip file closed' when trying to do reflection on the subject area package

```
                Starting server cocoMDS3 ...
                   ...cocoMDS3 failed to start
                       error message is: OMAG-ADMIN-500-004 The Subject Area OMAS service detected an unexpected java.lang.IllegalStateException exception with message zip file closed during initialization
```
```
Wed Oct 28 10:37:00 GMT 2020 cocoMDSx Startup OMAS-SUBJECT_AREA-0001 The Subject Area Open Metadata Access Service (OMAS) is initializing a new server instance
Wed Oct 28 10:37:00 GMT 2020 cocoMDSx Exception OMAS-SUBJECT_AREA-0005 The Subject Area Open Metadata Access Service (OMAS) is unable to initialize a new instance in server java.lang.IllegalStateException; the zip file closed exception occurred with error message: {2}
Wed Oct 28 10:37:00 GMT 2020 cocoMDSx Exception OMAS-SUBJECT_AREA-0005 Supplementary information: log record id ff7e1b00-0849-4b74-a432-bb075d4c2e6a java.lang.IllegalStateException returned message of zip file closed and stacktrace of
java.lang.IllegalStateException: zip file closed
	at java.base/java.util.zip.ZipFile.ensureOpen(ZipFile.java:823)
	at java.base/java.util.zip.ZipFile.jarStream(ZipFile.java:618)
	at java.base/java.util.zip.ZipFile$1.stream(ZipFile.java:1088)
	at java.base/java.util.jar.JarFile.stream(JarFile.java:535)
	at org.reflections.vfs.ZipDir.lambda$getFiles$2(ZipDir.java:21)
	at org.reflections.Reflections.scan(Reflections.java:260)
	at org.reflections.Reflections.scan(Reflections.java:216)
	at org.reflections.Reflections.<init>(Reflections.java:137)
	at org.reflections.Reflections.<init>(Reflections.java:182)
	at org.odpi.openmetadata.accessservices.subjectarea.server.mappers.MappersFactory.<init>(MappersFactory.java:27)
	at org.odpi.openmetadata.accessservices.subjectarea.server.mappers.MappersFactory.<init>(MappersFactory.java:54)
[...]

PR #4063, opened 2020-11-04
#3994 update springboot/springdoc after underlying issue fixed

planetf1:

This also addresses the issue seen in #3994 and can now be merged since #4060 is now merged which addresses an underlying issue (see issue, but probably an edge case in error handling in spring?)


mandy-chessell (approved):

planetf1:
FVT failed, but this was due to not pulling in the fix to UI server chassis that addressed the problem with subject area client.

planetf1:
See updates in #3994 . Basically the prereq fix did indeed fix the problem reported in the issue of the Polymer UI failing -- by removing subject area.

once subject area is activated in the main server chassis we see the same issue.

Closing PR. Followup in issue.

PR #4086, opened 2020-11-11
add additional node context for lineage

pmarius22:


lcpopa:
Marked all the queries I noticed that have no commit instruction. We need to have those since we use janus in an embedded configuration and transaction management is manually done

lcpopa (changes requested):

lcpopa (approved):
Will add the manual transaction management in a separate PR

Issue #4093, opened 2020-10-13
Background job still runs after data-engine proxy server instance is deleted


lpalashevski:
While testing data-engine proxy server noticed that background thread (poller job) is not canceled and resource is not released.
The server instance was deleted via admin-services API

Example:
DELETE https://localhost:8081/open-metadata/admin-services/users/admin/servers/datastage-proxy/instance

After closing the connector we should also cleanup/stop all background jobs/threads it created.

cmgrote:
The issue is that there is no call to the `disconnect()` method of `DataEngineProxyOperationalServices` from within the `deactivateRunningServiceInstances` method of `OMAGServerOperationalServices` -- hence transferred the issue to this repo, as all of this code is here in the core repo (not specific to any data engine proxy connector).

PR #4094, opened 2020-11-12
#4093 Ties in disconnect to shutdown data engine proxy services

cmgrote:


cmgrote:
@lpalashevski ^^ in case you want to re-test to confirm before merging?

mandy-chessell (approved):

PR #4098, opened 2020-11-13
Gradle build improvements including FVT execution

planetf1:
Further updates to the gradle build as per #3370

The main change in this PR is to support execution of our FVTs 

By ensuring the gradle build has the existing FVTs we can be more confident it is generating a good build/package,
and we can then use this as a base for more, longer tests - such as for docker, UI, with confidence.

By ensuring we have some tests that actually launch the server, we have more confidence the gradle built has created a working, configurable chassis & assembly

The gradle build is also much quicker and more flexible so an ideal place to add further testing.

Changes
 - Fixes some build dependency issues with the assembly
 - Updates gradle wrapper to 6.8.3
 - modifies groovy scripts to allow for use of properties (works well with maven), and system properties (needed for the gradle plugin)
  - implements appropriate Integration test in gradle build (supports parallelism)
  - See #4903 #4904 for 2 specific FVTs which need further work - not ideal, but at least we get some verification in place now, and a rough pattern for future use
  - see #4906 for general FVT improvements (longer term) - there are many deficiencies with this current setup, but this PR is a valuable incremental improvement

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Line.java
@@ -54,18 +54,16 @@ abstract public class Line implements Serializable, OmasObject {
     private Date effectiveFromTime = null;
     private Date effectiveToTime = null;
     private Map<String, String> additionalProperties;
-    protected String typeDefGuid;
     protected LineType lineType;
     // this is the line name
     protected String name;
     protected LineEnd end1;
     protected LineEnd end2;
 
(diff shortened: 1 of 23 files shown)


planetf1:
still in progress.

planetf1:
Gradle build currently failing with:
```
Error: Unable to access jarfile ../../../../../open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-2.8-SNAPSHOT-distribution.tar.gz/server/server-chassis-spring-2.8-SNAPSHOT.jar
```
Since the FVTs use the packaged jar for testing, and it's not present, the tests fail.
This is a good test and was as intended.
Investigating root cause (typically works second-time, so timing related)

planetf1:
The server launch was failing as the FVT server start task was depending on the zipped distribution archive, rather than the unpacked tar.

So sometimes the server existed (as the untar occurs immediately once the zipped archive is created), but often it did not...

Should be an accurate task dependency now

PR #4105, opened 2020-11-16
Add base entity and SupplementaryPorperties

mandy-chessell:


grahamwallis (commented):
Looks good to me. I didn't understand the different treatment of ExternalID but I'm sure there is a good reason for it.

PR #4113, opened 2020-11-18
UI-fix security check for controllers 

pmarius22:
split settings controller in  two for a better separation. Changed security config to request authentication for resources

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/EgeriaUIPlatform.java
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/EgeriaUIPlatform.java
@@ -9,6 +9,7 @@ import org.odpi.openmetadata.http.HttpHelper;
 import org.odpi.openmetadata.userinterface.uichassis.springboot.auth.AuthService;
 import org.odpi.openmetadata.userinterface.uichassis.springboot.auth.SessionAuthService;
 import org.odpi.openmetadata.userinterface.uichassis.springboot.auth.TokenAuthService;
+import org.odpi.openmetadata.userinterface.uichassis.springboot.service.ComponentService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.InitializingBean;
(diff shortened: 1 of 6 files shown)


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/SecurityConfig.java
```diff
@@ -25,14 +25,19 @@ protected void configure(HttpSecurity http) throws Exception {
...
-            .antMatchers("/**").permitAll()
+            .antMatchers("/").permitAll()
             .antMatchers("/error/**").permitAll()
+            .antMatchers("/node_modules/**").permitAll()
+            .antMatchers("/src/**").permitAll()
             .antMatchers("/h2/**").permitAll()
-            .antMatchers("/api/**").authenticated()
             .antMatchers("/api/css/**").permitAll()
             .antMatchers("/api/js/**").permitAll()
-            .antMatchers("/api/auth/login").permitAll()
```
bogdan-sava, line 34: This shouldn't be removed
It is the authentication request

bogdan-sava (changes requested):

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/SecurityConfig.java
sarbull: this file should not be changed

/** -> makes sure it gets everything that can be in the UI(in a different server, you can't just add them all here)
/api -> makes sure that it gets everything that can be in the API

the `/api/css/**`and `/api/js/**`are there because it's a config thing that comes from the API

please revert this file

marius-patrascu: reverting this file will cause all the requests to pass through without any authentifcation.

sarbull: `.antMatchers("/**").permitAll()` -> allows all

`.antMatchers("/api/**").authenticated()` -> filters all so that the `/api/**` requests are authenticated

bogdan-sava (approved):

Issue #3869, opened 2020-10-12
Ensure SSL verification done by default


planetf1:
When we initially added https to egeria, we added code to our client code to disable https certificate validation

It is essential for secure environments that these checks are enabled, potentially by default

This is currently not configurable though an unmerged PR #3238 proposed an additional variable to configure this

See #1403

planetf1:
Looking at this now..

planetf1:
In several places in our code, verification of the certificates used on a connection (including the hostname associated with the certificate) is done

Hard-coded ssl disabling is currently done in
 * SpringRESTClientConnector.java - this is consistently used as a client connector for most of our code, so a hardcoded override here is a big impact

Configurable ssl disabling is done in
 * HttpHelper.java (using strict.ssl) - this is a utility fn used when an application needs to disable the checks (globally)
 * LineageGraphRemoteConnectorProvider.java (using gremlin.remote.driver.cluster.sslSkipCertValidation)
 * OMAGServerPlatform.java - this is how we pick up the strict.ssl configuration and use this to call the http helper above
 * EgeriaUIPlatform.java - as above, but for the spring based UI

The changes needed are:
 * The SpringRESTClientConnector should not always disable ssl
 * The http helper already offers suitable code (Basically the same) to disable ssl globally
 * This disabling should be done in the application rather than library (it was originally done this way as part of our migration to security - it's now time to address this in any remaining client code)

As well as in testcode
 * HttpHelperTest.java

planetf1:
The likely affected 'applications' will include the following - note these are in gradle form, a useful cut/paste from our assembly definitions :-)
```
            into('utilities') {
                from { project(':open-metadata-resources:open-metadata-archives:design-model-archives:cloud-information-model').shadowJar }
                from { project(':open-metadata-resources:open-metadata-archives:design-model-archives:glossary-canonical-model').shadowJar }
                from { project(':open-metadata-resources:open-metadata-archives:open-metadata-types').shadowJar }
                from { project(':open-metadata-resources:open-metadata-archives:open-connector-archives').shadowJar }
                fileMode = 0755
            }
            into('samples') {
                from { project(':open-metadata-resources:open-metadata-samples:access-services-samples:asset-management-samples:asset-create-avro-sample').shadowJar }
                from { project(':open-metadata-resources:open-metadata-samples:access-services-samples:asset-management-samples:asset-create-csv-sample').shadowJar }
[...]

PR #4136, opened 2020-11-23
Enable SSL verification by default (2.6)

planetf1:
See #3869 

I RECOMMEND WE INCLUDE THIS AFTER BRANCHING FOR THE 2.5 RELEASE
 - since we need to be sure we've caused additional problems with additional connectors - igc, postgres, user interfaces/node etc & have all the docs covered

Functionality
 - no longer force-disables ssl verification in spring rest connector
 - updates FVT test cases to explicitly disable verification (as test builds use a self-certified certificate and unknown dynamic hosts)
 - updates client samples to use a java system property 'strict.ssl' to disable checks

Tested a selection of clients and notebooks

Caveats
 - Raised #4138 for presentation server change


planetf1:
@cmgrote cc: to determine impact on igc/atlas connectors

planetf1:
Now additionally tested  - docker image, Spring UI, Presentation Server - no regressions noted.

lpalashevski (approved):

PR #4140, opened 2020-11-24
Initial load trigger load for single asset

pmarius22:


open-metadata-implementation/access-services/asset-lineage/asset-lineage-spring/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/spring/AssetLineageResource.java
```diff
@@ -37,4 +37,21 @@ public GUIDListResponse publishEntities(@PathVariable String serverName,
...
+    /**
+     * Find the entity by guid and publish the context for it
+     *
+     * @param serverName name of server instance to call
+     * @param userId     the name of the calling user
+     * @param guid       the guid of the entity to build context
+     * @param entityType the name of the relationship type
+     * @return a list of unique identifiers (guids) of the available entity with the given type provided as a response
+     */
+    @GetMapping(path = "/publish-entity/{entityType}/guid/{guid}")
```
lpalashevski, line 50: I am not sure if we have agreed on Egeria-wide REST API endpoint design practices but having /guid/ in the path seems bit odd to me. I would prefer to have it without this part and call parameter id instead. The fact that is guid is just a choice and format of an identifier. 

So my proposal for this endpoint is: /publish-entity/{entityType}/{id}

@popa-raluca any thoughts from the previous practice with rest endpoints?

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/outtopic/AssetLineagePublisher.java
```diff
@@ -113,7 +113,7 @@ public void publishGlossaryContext(String glossaryTermGUID) throws OCFCheckedExc
         Map<String, Set<GraphContext>> context = glossaryHandler.buildGlossaryTermContext(serverUserName, entityDetail);
 
         if (MapUtils.isEmpty(context)) {
-            log.debug("Context not found for the entity {} ", entityDetail.getGUID());
+            log.error("Context not found for the entity {} ", entityDetail.getGUID());
```
lpalashevski, line 116: Thinking about this better now, does not seem like ERROR. There can be valid entity that does not have relationships and what asset lineage should do is log this as INFO because of the fact that it is filtering it out. Also suggesting to skip sending the lineage event (otherwise we are sending empty lineage events). This also applies for similar method "publishProcessContext"

lpalashevski (changes requested):

lpalashevski (approved):

PR #4141, opened 2020-11-24
Data Engine OMAS FVT

alexandra-bucur:


planetf1:
Note that the gradle build doesn't yet run the other fvts correctly. There's an attempt in #4098 but I have a dependency problem (timing) between the creation of the assembly (from where we run the chassis) and the actual test.

So I would probably ensure the code at least builds (maven+gradle) but wouldn't worry about launching the fvts in gradle.
(tips also welcome in the above mentioned pr...)

popa-raluca (approved):

open-metadata-test/open-metadata-fvt/access-services-fvt/data-engine-fvt/src/test/java/org/odpi/openmetadata/accessservices/dataengine/RepositoryService.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+import static org.apache.http.ssl.SSLContexts.custom;
+
+public class RepositoryService {
```
davidradl, line 38: I am wondering - is this a mock object? I was not expecting mock objects in an FVT.
Or is this a helper class that might be more useful more widely- in which case - should it live in test code. 
Some comments around this and other code to explain the testing intent would be really helpful for me to understand what you are looking todo here.

open-metadata-test/open-metadata-fvt/access-services-fvt/data-engine-fvt/src/test/java/org/odpi/openmetadata/accessservices/dataengine/fvt/DataEngineFVT.java
```diff
@@ -0,0 +1,94 @@
...
+        assertEquals(softwareServerCapability.getEngineVersion(), entity.getProperties().getPropertyValue(VERSION).valueAsString());
+        assertEquals(softwareServerCapability.getPatchLevel(), entity.getProperties().getPropertyValue(PATCH_LEVEL).valueAsString());
+        assertEquals(softwareServerCapability.getQualifiedName(), entity.getProperties().getPropertyValue(QUALIFIED_NAME).valueAsString());
+        assertEquals(softwareServerCapability.getSource(), entity.getProperties().getPropertyValue(SOURCE).valueAsString());
+
+    }
+
+    @ParameterizedTest
+    @MethodSource("org.odpi.openmetadata.accessservices.dataengine.PlatformConnectionProvider#getConnectionDetails")
+    public void verifyLineageMappingsForAJobProcess(String userId, DataEngineClient dataEngineClient, RepositoryService repositoryService)
```
davidradl, line 66: I am missing something here, how does the lineage graph get populated for this FVT?

PR #4150, opened 2020-11-26
Adds tighter checking of request bodies and cleans up a number of comments and todo items

grahamwallis:


open-metadata-implementation/view-services/rex-view/rex-view-api/src/main/java/org/odpi/openmetadata/viewservices/rex/api/rest/RexRelationshipRequestBody.java
```diff
@@ -68,7 +69,6 @@ public String toString()
                 ", platformName=" + platformName +
                 ", relationshipGUID=" + relationshipGUID +
                 ", enterpriseOption=" + enterpriseOption +
-                ", gen=" + gen +
```
davidradl, line 71: I assume we use toString for debugging , in that case wouldn't we want all the content like gen and depth etc. As an observation, normally there would be an equals and hashCode associated with a java bean.

grahamwallis: I have removed gen (completely).

PR #4159, opened 2020-11-27
AL OMAS - enhance exception handling for initial load

popa-raluca:
Enhance exception handling to catch more generic Exception & improve logging for publishing the context for single entity

### DIFF
--- a/open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/AssetLineageRestServices.java
+++ b/open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/AssetLineageRestServices.java
@@ -149,12 +149,15 @@ public class AssetLineageRestServices {
      */
     private String publishEntityContext(AssetLineagePublisher publisher, EntityDetail entityDetail) {
         try {
-            log.info("Asset Lineage OMAS published the context for entity with guid {}", entityDetail.getGUID());
+            log.info("Asset Lineage OMAS started building the context for entity with guid {}", entityDetail.getGUID());
+
             return publishContext(entityDetail, publisher);
-        } catch (JsonProcessingException | OCFCheckedExceptionBase e) {
-            log.error("Failed to publish context for entity of type " + entityDetail.getType() + " and guid "
-                    + entityDetail.getGUID());
+        } catch (Exception e) {
+            log.error("Exception while trying to publish context for entity of type " + entityDetail.getType() + " and guid "
+                    + entityDetail.getGUID(), e);
         }
+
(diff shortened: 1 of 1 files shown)


lpalashevski (approved):

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/AssetLineageRestServices.java
```diff
@@ -149,12 +149,15 @@ public GUIDListResponse publishEntity(String serverName, String userId, String e
...
         try {
-            log.info("Asset Lineage OMAS published the context for entity with guid {}", entityDetail.getGUID());
+            log.info("Asset Lineage OMAS started building the context for entity with guid {}", entityDetail.getGUID());
+
             return publishContext(entityDetail, publisher);
-        } catch (JsonProcessingException | OCFCheckedExceptionBase e) {
-            log.error("Failed to publish context for entity of type " + entityDetail.getType() + " and guid "
-                    + entityDetail.getGUID());
+        } catch (Exception e) {
+            log.error("Exception while trying to publish context for entity of type " + entityDetail.getType() + " and guid "
```
mandy-chessell, line 156: This exception handling is not adequate for Egeria.  Any use of `log.error()` suggests that this exception is meaningful to problem diagnosis.  This means it must be captured either in the audit log as an exception (ie using `auditLog.logException()`) or returned as a formal exception on a REST response. Since this seems to be part of event processing rather than a REST response then the audit log seems the right choice.

mandy-chessell (changes requested):

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/AssetLineageRestServices.java
lpalashevski: Hi @mandy-chessell  I agree in principle. We are going to move all the error loggers with auditLog.logException() and also move some log.info to proper AuditLog info messages starting next week.

The reason why we want this intermediary logging right now It because we were experiencing technical issues using slf4j auditlog connector which resulted in missing original exception that break the initial load process. 

Having in mind that this code is still under development can we consider this as technical dept for now and merge this change?

p.s. Another reason is because we are planning to do extensive "initial load" run using actual meta-data load over the weekend.

lpalashevski: Issue #4160 created to record and track the technical dept

marius-patrascu (approved):

lpalashevski:
Closing this PR. Should be fixed by #4246

Issue #4256, opened 2020-12-03
[server-chassis-spring] add full-platform profile


bogdan-sava:
All the OMAS and view-server capabilities dependencies will be moved under a profile called full-platform.
The profile will be active by default

In case you want to exclude all of them and use loader.path to bring in only the desired functionality profile full-platform can be deactivated by a property called adminChassisOnly.

Usage: 
`mvn clean install -DadminChassisOnly`

bogdan-sava:
Swagger examples:

Building with option **adminChassisOnly**:

<img width="1482" alt="Screen Shot 2020-12-03 at 8 41 42 AM" src="https://user-images.githubusercontent.com/40362260/100973674-a73ff680-3543-11eb-99e0-a51269a3e2ed.png">

Building with **adminChassisOnly**, but run with loader.path and having **asset-catalog** dependencies loaded:

<img width="1494" alt="Screen Shot 2020-12-03 at 8 40 49 AM" src="https://user-images.githubusercontent.com/40362260/100973811-f25a0980-3543-11eb-9d13-d3621cd4f801.png">

bogdan-sava:
merged

PR #4257, opened 2020-12-03
[server-chassis-spring] add full-platform profile

bogdan-sava:
see https://github.com/odpi/egeria/issues/4256 for behavior of the change


open-metadata-implementation/server-chassis/server-chassis-spring/README.md
```diff
@@ -35,6 +35,26 @@ server in the OMAGServerPlatform many times over multiple platform restarts.
...
+mvn clean install -DadminChassisOnly
+```
+ the server-chassis-spring will contain only the following services:
+ * Administration Services - Operational
+ * Administration Services - Platform Configuration
+ * Administration Services - Server Configuration
+ * Administration Services - Server Configurations
+ * Platform Services
+ 
+ In this case, for any extra functionality, such as desired access services or view server,
```
planetf1, line 54: at an appropriate point (doesn't have to be this PR) we should include information on this in the full admin guide - which should also explain how to use in docker etc

open-metadata-implementation/server-chassis/server-chassis-spring/README.md
```diff
@@ -35,6 +35,26 @@ server in the OMAGServerPlatform many times over multiple platform restarts.
 A detailed description of the internals of the OMAGServerPlatform during server start up
 is available [here](../../../open-metadata-publication/website/omag-server).
 
+## Maven build profiles
```
planetf1, line 38: in future we'll need to add to gradle too

open-metadata-implementation/server-chassis/server-chassis-spring/README.md
```diff
@@ -35,6 +35,26 @@ server in the OMAGServerPlatform many times over multiple platform restarts.
...
+ the server-chassis-spring will contain only the following services:
+ * Administration Services - Operational
+ * Administration Services - Platform Configuration
+ * Administration Services - Server Configuration
+ * Administration Services - Server Configurations
+ * Platform Services
+ 
+ In this case, for any extra functionality, such as desired access services or view server, 
+ use [ **loader.path** spring-boot functionality ](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-executable-jar-format.html#executable-jar-property-launcher-features).
+
```
planetf1, line 56: I wonder what we should change in the assembly if the user makes use of this profile? We should probably ensure the relevant jars are added there (for maven and gradle)

planetf1 (approved):
A really useful step to full dynamic loading.

In future I think we could do with
 - more doc elaboration
 - gradle support
 - addressing dependencies in the assemblies
 - making this approach the default

PR #4265, opened 2020-12-03
added role for admin in user chassis properties

pmarius22:
updated the roles for UI. Default behavior when a role is missing is restrict everything and in order for a role to have access to everything the "all" permission should be applied.

### DIFF
--- a/open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/AssetContextHandler.java
+++ b/open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/AssetContextHandler.java
@@ -50,7 +50,6 @@ public class AssetContextHandler {
     private final HandlerHelper handlerHelper;
     private final List<String> supportedZones;
 
-    private AssetContext graph;
 
     /**
      * @param invalidParameterHandler    handler for invalid parameters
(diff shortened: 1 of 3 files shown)


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/resources/application.properties
```diff
@@ -59,7 +59,8 @@ authentication.mode=token
 # component visibility for different roles
 # these properties are used to create a model for UI to show or hide components
 # role.visibleComponents.ROLE_NAME=coma,separate,values
-role.visibleComponents.USER=asset-catalog,about
+role.visibleComponents.USER=asset-catalog,about,asset-lineage,vertical-lineage
+role.visibleComponents.ADMIN=all
```
sarbull, line 63: i believe "all" is unnecessary, we should stick with @bogdan-sava's approach when user doesn't have anything configured it will default to provide access to everything

sarbull (changes requested):

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/resources/application.properties
marius-patrascu: I feel like this approach is more "mistake proof". In case someone forgets to configure the permissions for a role it should not get access to everything, the permissions for accessing everything should be given knowingly. 
I have discussed this this approach with @bogdan-sava and we agreed that this is a better handling of the "admin" role.

sarbull: @marius-patrascu the user should initially have a ROLE attached, such as "ADMIN" or "USER", so this means that "all" is still irrelevant if the role includes it or not

bogdan-sava (approved):

PR #4267, opened 2020-12-03
Unique messages for OMRS

mandy-chessell:
There are a couple of OMRS error messages that do not have unique strings in them.  They are causing the unit tests to fail.  Nigel is working on why these test are not running automatically in the builds.


planetf1 (approved):

planetf1:
Yes..yes more at  https://github.com/odpi/egeria/runs/1493853873

PR #4282, opened 2020-12-08
Change response status for unsuccessful authentication

alexandra-bucur:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/pom.xml
```diff
@@ -320,6 +320,12 @@
             <artifactId>glossary-view-api</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework.ldap</groupId>
+            <artifactId>spring-ldap-core</artifactId>
+            <version>2.3.3.RELEASE</version>
```
bogdan-sava, line 326: version should be property in main pom, and dependency added in the managed dependencies in the main pom

bogdan-sava (changes requested):

bogdan-sava (approved):

PR #4264, opened 2020-12-03
#4252 ensure surefire always triggers in test phase for all modules

planetf1:
Signed-off-by: Nigel Jones <nigel.l.jones+git@gmail.com>

Fixes #4252 

 - Adds surefire as a default build plugin for all modules
 - surefire is intelligent enough to skip pom modules & others without tests

PR #4294, opened 2020-12-10
Fix UI unit tests

pmarius22:


planetf1 (changes requested):
Can I suggest we merge #4264 first -- or we you able to test the UT fix with that change?  (in that PR we re-enable the maven launch of surefire - which wasn't running)

marius-patrascu:
Hi @planetf1, I am fine with merging the other PR first.

planetf1 (approved):

Issue #3392, opened 2020-07-28
server-chassis logging pattern


bogdan-sava:
Logging pattern recently changed to 
`%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n`
and cannot be set up from application.properties anymore.

It still can be set by placing logback.xml into the classpath. Ex:
```
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <include resource="org/springframework/boot/logging/logback/base.xml"/>
</configuration>

```

As a requirement, I think, logging pattern should be configurable and not depend on the implementation.

planetf1:
Our libraries should only use the slf4j-api and hence not include any logging implementation
our runnable jars - ie chassis, UI, utilities, samples - will need a binding, and we typically use logback for this
(test cases also either use logback-test.xml or bind with slf4f-simple)
logback will then use the config it finds on the classpath - which could be from one we specifically add, a dependency, or found externally.

Is this occuring in the chassis, UI, a specific utility?

Looking at our current tree,  we only have logback.xml for some of our samples, CTS client & archive utilities

I checked git history and it seems we've never had an overriding logback.xml for the server chassis or UI etc

The format I see running the chassis with no additional config is:
```
17:59:43.192 [main] INFO  o.o.o.s.s.OMAGServerPlatform - Starting OMAGServerPlatform on planetf1mac with PID 40630 (/Users/jonesn/IdeaProjects/egeria-maven/open-metadata-implementation/server-chassis/server-chassis-spring/target/server-chassis-spring-2.2-SNAPSHOT.jar started by jonesn in /Users/jonesn/IdeaProjects/egeria-maven)
17:59:43.200 [main] INFO  o.o.o.s.s.OMAGServerPlatform - No active profile set, falling back to default profiles: default
17:59:48.648 [main] INFO  o.s.b.w.e.tomcat.TomcatWebServer - Tomcat initialized with port(s): 9443 (https)
17:59:48.694 [main] INFO  o.a.coyote.http11.Http11NioProtocol - Initializing ProtocolHandler ["https-jsse-nio-9443"]
17:59:48.694 [main] INFO  o.a.catalina.core.StandardService - Starting service [Tomcat]
17:59:48.694 [main] INFO  o.a.catalina.core.StandardEngine - Starting Servlet engine: [Apache Tomcat/9.0.36]
```

[...]

PR #4324, opened 2020-12-15
fix Superfluous developer logging at OMAG Server Platform start

bogdan-sava:


bogdan-sava:
Please see #4324

bogdan-sava:
this is also fix for https://github.com/odpi/egeria/issues/3392

planetf1 (changes requested):
See discussion in #4324

I'd be inclined to skip this change as
 a) The root cause in many ways is the split role of the module affected - being both a library and executable utility. - though this change does address an immediate impact so this is only an observation/preference not a blocking 'no'
 b) The underlaying issue is not fixed by this change as I see it (though not run code) as the default spring server chassis config file has a default log level of INFO

bogdan-sava:
That is not true.
The default is override by application.properties.

bogdan-sava:
The default spring-boot is override by default aplication.properties as we discussed long time ago, but leaves option to setup levels on deploy.

planetf1:
I completely agree with the intent of the change, I just couldn't see how the code change prevented the info messages given what I saw in the default spring file and the logback xml for the 'library'. That being said I have just ran the chassis and as you asserted, the logging is back as we had expected, so good job :-) Sure we can do the split on the library but this does indeed seem to address the immediate logging issue. thanks.

PR #4326, opened 2020-12-16
#1102 Change default directories for stored data to be in a containab…

### LINKED ISSUE #1102: Single location for read/write data files in egeria
One obervation that applies across egeria -- we really need to ensure that, without too much config, all data (writing) is constrained to a single/small number of folders. this is essential when working with containers which are basically read-only, but with specific read-write volumes mounted at a specific location. Each volume has overhead - sometimes sys admin created, so having files in many places would be tricky.

I will raise a seperate issue to track this, since it's much broader than just lineage - which is only following the existing pattern! .. and there isn't really precedent set yet.

_Originally posted by @planetf1 in https://github.com/odpi/egeria/pull/1100_

planetf1:
…le tree

Moves all dynamic data under a 'data' subtree in egeria so that the remainder of egeria can be read only for improved security, and to make it easier to backup, move any file based data dynamically created by egeria. Specifically this is also required for a container environment so that we can create a docker volume, or k8s pvc at an appropriate location to allow for persistence in those environments

See #1102 for more information 

Fixes #1102


planetf1:
Added release notes updates.

Note that one impact of this change, is that if a user simply upgrades egeria without acting upon the note in the release notes, they will not have any configs found

planetf1:
Have now been able to test audit logs.  ok

planetf1:
I've had several verbal approvals. If no negative comments here I propose to merge around midday today.
Please raise an issue with any further corrections/concerns after this point

Issue #4252, opened 2020-12-02
Maven builds not running all tests


planetf1:
PR #4248 failed in the Gradle build in https://github.com/odpi/egeria/pull/4248/checks?check_run_id=1485869158
```
> Task :open-metadata-implementation:adapters:open-connectors:integration-connectors:files-integration-connectors:test FAILED

Gradle suite > Gradle test > org.odpi.openmetadata.adapters.connectors.integration.basicfiles.ffdc.AuditCodeTest > testAllAuditCodeValues FAILED
    java.lang.AssertionError at AuditCodeTest.java:24


FAILURE: Build failed with an exception.
```

Yet the maven build completes successfully.

On closer examination we see in https://github.com/odpi/egeria/pull/4248/checks?check_run_id=1485869159:
```
[INFO] --- maven-surefire-plugin:3.0.0-M5:test (default-test) @ files-integration-connectors ---
[INFO] 
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

So it would seem the gradle build ran the tests, maven did not.

Need to investigate to see if this is just this component, or if we are missing other tests too.

cc: @mandy-chessell

planetf1:
Top level pom needed an explicit entry in the build phase for running surefire.

Output (with a failing test as an example) is now:
```
13:24:16,316 [INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.205 s - in org.odpi.openmetadata.repositoryservices.rest.properties.TestVoidResponse
13:24:16,374 [INFO] Tests run: 3, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.108 s - in org.odpi.openmetadata.repositoryservices.auditlog.TestOMRSAuditingComponent
13:24:16,428 [ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.138 s <<< FAILURE! - in org.odpi.openmetadata.repositoryservices.auditlog.TestOMRSAuditCode
13:24:16,430 [ERROR] testAllAuditCodeValues(org.odpi.openmetadata.repositoryservices.auditlog.TestOMRSAuditCode)  Time elapsed: 0.036 s  <<< FAILURE!
[...]

PR #4330, opened 2020-12-17
#4252 Maven not running all tests

planetf1:

This is an additional change for 4252 to try and run all tests

I have removed the explicit references to junit/jupiter/testng libraries since surefire has logic to try and automatically determine the right versions to use. Removing the explicit declaration seems to allow surefire to correctly find more tests.

For example the files integrator tests now function again - this was broken in the later stages of the original PR as I cleaned up a more local surefire definition in the pom & migrated to the top level pom, which resulted in a slightly different default config


planetf1:
With this change only one test target is showing 0 tests run. and that is
```
12:43:57,758 [INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.058 s - in org.odpi.openmetadata.accessservices.analyticsmodeling.test.utils.TestUtilities
```
which actually doesn't have any testscases as it's test utils code (probably should be renamed)

Prior to this we had 15 instances of tests not running in the build. Most likely due to not letting surefire figure out whether they were junit 4/5/testNG by itself. So more tests had been running than before the original PR where there were 53 instances, ie in fact the original PR made an improvement, but left some tests still not working (testNG in this case)

cc: @mandy-chessell

popa-raluca (approved):

Issue #4302, opened 2020-12-11
Persistance storage - docker image


planetf1:
Currently the docker image built for base egeria does not provide any documentation, or direct support, for using persistent storage. For example the configuration files used to configure servers, audit logs , and of course a graph metadata store.

Thus far our docker image has been used for a tutorial environment, where this lack of persistency is actually a good thing .... 
Additionally, egeria itself provides lots of options through connectors & configuration which might allow any storage to be externalized.

However for many simple (including learning, sample) environments it's necessary to support this in a simple way. At a docker level this might mean via a volume. In k8s we would create Persistent Volume Claims.  Further it should be clear in the documentation how this is most easily done.

planetf1:
The broad strategy here is to
 - prove how to configure egeria (and fix if required) so that all data files are in well defined directories that could be mapped to an external storage medium
 - ensure this is documented in our admin guide
 - Extend our docker image accordingly, plus docs
 - Extend our compose, charts accordingly

Detailed todos:
 - [ ] Extend coco pharma lab notebook to create all configuration in a well defined subdirectory (not root)
 - [ ] Extend notebook to support use of graph store instead of inmemory
 - [ ] Ensure graph store is also in a well defined subdirectory
 - [ ] Ensure admin guide is correct & helpful for end users wishing to make this kind of configuration
 - [ ] Extend docker image to easily support/define volumes to be used
 - [ ] Ensure docker usage docs updated
 - [ ] Extend lab helm chart to optionally allow persistant storage to be used

Further deliverables will follow - such as persistance fot the egeria operator, or other deployment options, this is a first step to provide the approach.

planetf1:
Coco pharma notebook now stores files under 'config' by default, making use of the API to set the configuration store
This provides a good example, and validation

ie:
```
jonesn:egeria/ (issue4302) $ ls -la config                                                                                             [15:25:39]
total 624
drwxr-xr-x  11 jonesn  staff    352 14 Dec 15:25 .
drwxr-xr-x  57 jonesn  staff   1824 14 Dec 15:25 ..
-rw-r--r--   1 jonesn  staff  45115 14 Dec 15:25 omag.server.cocoMDS1.config
-rw-r--r--   1 jonesn  staff  51982 14 Dec 15:25 omag.server.cocoMDS2.config
-rw-r--r--   1 jonesn  staff  41319 14 Dec 15:25 omag.server.cocoMDS3.config
[...]

PR #4333, opened 2020-12-17
Support Persistence in container environment

planetf1:
This PR adds support for persisting egeria data within a Docker or kubernetes environment
*work in progress*

Dependencies:
 * #4326 is needed. This PR will merge and build and indeed run without it, but it will not persist data (as nothing is put in the right place!)
 
Done so far
 - tested docker image : No change needed other than #4326
 - added README & synchronized with copy on docker hub
 - added note in Dockerfile as 2nd reminder of how to use persistence

Todo
 - docker compose lab sample
 - k8s lab sample

### DIFF
--- a/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/jupyter.yaml
+++ b/open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/jupyter.yaml
@@ -86,6 +86,8 @@ spec:
               value: "{{ .Release.Name }}-kafka:9092"
             - name: uiExternalURL
               value: "https://localhost:18443"
+            - name: repositoryType
+              value: {{ .Values.egeria.repositoryType}}
           ports:
             - containerPort: 8888
           readinessProbe:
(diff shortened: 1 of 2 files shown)


planetf1:
See #4302

planetf1:
After adding a 'securityContext.fsGroup:0' setting (this says which unix group is used for storage.. 0 (root - but only as a group) is what the redhat ubi8 jdk image uses as default), storage is correctly accessible by each pod.

I deployed, configured, then killed a pod. It restarted, and after restart I could see /deployments/data still contains the correct configuration etc files we would expect.

it's worth noting that pod rescheduling is a common experience. But currently with egeria the on-disk state is not enough to get the platform back up and running - the servers need to be started - or configuration set to autostart.

This is one area where an 'operator' will be able to do a slightly more thorough job - it could monitor the pods readiness then issue the right calls to start the configured servers.

However after restart now seeing that the notebook start (in a restarted pod ie platform) is returning notebook errors. investigating:
```
Checking OMAG Server Platform availability...
    Core Platform is active
        Checking OMAG Server cocoMDS2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
~/common/environment-check.ipynb in <module>
      6 print("\nChecking OMAG Server Platform availability...")
      7 
----> 8 activatePlatform(corePlatformName, corePlatformURL, [cocoMDS2Name, cocoMDS3Name, cocoMDS5Name, cocoMDS6Name])
      9 activatePlatform(dataLakePlatformName, dataLakePlatformURL, [cocoMDS1Name, cocoMDS4Name, cocoView1Name ])
     10 activatePlatform(devPlatformName, devPlatformURL, [cocoMDSxName])

```

Issue #4342, opened 2020-12-18
Security Analysis - Null Dereference


lpalashevski:
Umbrella issue for Code quality improvements re OWASP Redundant Null Checks and  [Null Dereference](https://owasp.org/www-community/vulnerabilities/Null_Dereference)

PR #4343, opened 2020-12-18
Fix NPEs and redundant null checks

popa-raluca:


popa-raluca:
Fixes for https://github.com/odpi/egeria/issues/4342

mandy-chessell (approved):
Code behaviour ok - will fix layout when in master

Issue #4462, opened 2021-01-07
Files integration daemon failing with errors 


davidradl:
@mandy-chessell raising an issue to track this . 

The log shows errors like this 
Tue Jan 05 17:22:26 GMT 2021 cocoInt2 Exception BASIC-FILES-INTEGRATION-CONNECTORS-0012 An unexpected org.odpi.openmetadata.adapters.connectors.integration.basicfiles.ffdc.exception.FileException exception was returned to the null integration connector when it tried to update the DataFolder null in the metadata repositories for directory null.  The error message was BASIC-FILES-INTEGRATION-CONNECTORS-400-004 An unexpected org.odpi.openmetadata.frameworks.connectors.ffdc.PropertyServerException exception was returned to the null integration connector by the Files Integrator OMIS getFolderElement method when trying to retrieve the FileFolder asset for directory /Users/davidradley/testplatforms/platform1/test1 (absolute path /Users/davidradley/testplatforms/platform1/test1).  The error message was OMAG-COMMON-503-001 A client-side exception was received from API call getFolderByPathName to OMAG Server cocoMDS1 at https://localhost:9443.  The error message was CLIENT-SIDE-REST-API-CONNECTOR-503-002 A client-side exception org.springframework.web.client.HttpClientErrorException$MethodNotAllowed was received by method getFolderByPathName from API call https://localhost:9443/servers/{0}/open-metadata/access-services/data-manager/users/{1}/filesystems/folders/by-name to server cocoMDS1 on platform https://localhost:9443.  The error message was 405 : [{"timestamp":"2021-01-05T17:22:26.830+00:00","status":405,"error":"Method Not Allowed","message":"","path":"/servers/cocoMDS1/open-metadata/access-services/data-manager/users/OMAGServer/filesystems/folders/by-name"}]

[...]

PR #4463, opened 2021-01-07
#4462 Amend the File rest call implementation

davidradl:


mandy-chessell:
This fix is already in master

mandy-chessell:
Not sure what to do with this fix as it duplicates the fix I put in yesterday.  I am puzzled why there is no update conflicts being detected by git.   However, we do need the first commit (afeacd7).  I am going to risk pushing this in and deal with any mess in the FilesResource.java.  The lines in the file are consistent, so hopefully it will merge ok and not object to the fact that the change is a no-op.

davidradl:
I suspect if the changes were absolutely identical, git will not see there is a change, so will not flag a conflict and will just reduce the diff that this commit has.

PR #4470, opened 2021-01-08
[ui-chassis-spring] token validation

bogdan-sava:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/pom.xml
```diff
@@ -51,6 +51,22 @@
...
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-data-redis</artifactId>
+            <version>${spring-boot.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>io.lettuce</groupId>
+            <artifactId>lettuce-core</artifactId>
+            <version>5.3.3.RELEASE</version>
```
planetf1, line 62: the explicit version should be removed

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/AuthFilter.java
```diff
@@ -37,7 +37,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             LOG.error("TOKEN EXPIRED", e.getMessage());
             SecurityContextHolder.getContext().setAuthentication(null);
         }catch (JwtException e){
-            LOG.error("Token error", e);
+            LOG.debug("Token error", e.getMessage());
```
planetf1, line 40: format error

PR #4475, opened 2021-01-08
Dino view service extensions to handle engine services

grahamwallis:


open-metadata-implementation/view-services/dino-view/dino-view-server/src/main/java/org/odpi/openmetadata/viewservices/dino/server/DinoViewRESTServices.java
```diff
@@ -1594,6 +1594,95 @@ public DinoServiceListResponse serverGetIntegrationServices(String serverName, S
...
+        final String methodName = "serverGetEngineServices";
+
+        RESTCallToken token = restCallLogger.logRESTCall(serverName, userId, methodName);
+
+        DinoServiceListResponse response = new DinoServiceListResponse();
+
+        if (requestBody != null)
+        {
+
+            // The serverName parameter to the RequestSummary is the target server not the server running the VS
```
davidradl, line 1617: what is the VS?

PR #4480, opened 2021-01-11
#4001 add docs on using an external mount for a custom connector

### LINKED ISSUE #4001: Add external mount point for connectors to egeria docker image
I would like to have the ability to add my connector from local machine be loaded in classpath for egeria server with containers such as with a docker-compose file.  Seems this can be done with an addition of a directory like /deployment/server/externallib to the classpath for the image so users could mount volume for local connector .jar files and load those up

Then in our local docker-compose we just do:

volumes:
     - ~/.m2/path-to-my-jar-file-directory-example/:/deployments/server/external_libs

This has some security implications too.

planetf1:

Fixes #4001 

Adds documentation on how to add a custom jar (ie a connector) into the classpath so that it can be dynamically loaded
using docker. A specific example is provided.

Note there are various ways of doing this -- the objective was to ensure at least one way was documented with a suitable example. As new helm charts are added we'll revisit other approaches - which may be different for k8s.


planetf1:
@CDaRip2U Any comments appreciated - or contributions as to how you do it. This is one way that works, but there's a plethora of options ... hopefully these notes give someone an appreciation of what routes they could take

planetf1:
will merge to get initial instructions out, but further comments welcome and can be raised as new issues.

PR #4481, opened 2021-01-11
Update Dockerfile

domingospanta:
Copy and paste error corrected on the environment variable LOADER_PATH.


mandy-chessell:
Instructions on fixing the DCO in your commit is here:

 https://github.com/odpi/egeria/blob/master/developer-resources/why-the-dco.md

planetf1:
Co-incidentally this is fixed in PR #4480 which also extends the documentation for
adding additional connectors to the image.

I think that superceeds your proposal - but thanks for spotting. Would also very much appreciate any comments you have
on 4480 .

planetf1:
I am going to close this as I believe the referenced PR has addressed the requirement
Please open an issue if you believe there is more work to do/issues with it, and a PR is welcome too

PR #4498, opened 2021-01-19
#4497 amend classification process in enterprise connector

davidradl:


mandy-chessell (changes requested):
This change backs out the change to allow classifications to be stored in a different repository to the entity - we need to fix the problem with the new approach rather than back out the change since it will cause more failures when entities are coming from repositories such as IGC.

davidradl:
@mandy-chessell ok closing this pr out

Issue #4454, opened 2021-01-06
Simple helm chart for deployment


planetf1:
The current egeria 'lab' helm chart is targeted at a very specific usage - running the notebooks and showing a coco
pharma scenario

Once beyond that education users want to be able to play with egeria - for example creating a simple data catalog.

Providing an easy way to get egeria up and running (in k8s) would make this next step easier. The simplest possible
configuration could be
 - no cohort
 - single egeria server (automatically configured) - local graph repo with persistence
 - a few select connectors (like csv, perhaps postgres when available)
 - no plugin security etc.
 
 However such a user would also want to be able to access at least some UIs. For this we would also need
  - cohort
  - view server
  - deploy presentation-server

Previously, for the VDC helm chart, we configured servers up-front via curl scripts. In the notebooks we use python (the actual rest calls are somewhat obfuscated as they are now effectively within a python api definition). We also have an admin client (and a java code sample). In our FVTs we use groovy. The key will be to do this automatically, with only very simple (and defaulted) configuration such as a server name.

The egeria operator work should also offer this capability, though of course goes further in allowing the configuration to be modified, extended, scaled. That's not ready yet, and a quick tactical example could help with some current user difficulties in deploying egeria

planetf1:
I've now started putting a simple chart together:

- [ ] Initial simple chart deployment of egeria/kafka/presentation server
- [ ] Automatic setup/configuration of server+view server

planetf1:
Checked in further updates.

There is now an attempt to configure an egeria server as part of the chart deployment. However this is work in progress and not yet functional. The intent is to have the script within a k8s configmap. Currently that script is a shell script with no error checking -- to prove the process & mechanisms -- but will likely be updated to groovy or python as this is more sustainable moving forward.

Will return to this next week.

planetf1:
* Egeria metadata server (graph) deployed
* Server will autostart after initial config (*see below)
* View Server setup

Currently the server deploys, configures, starts but needs more testing

*note
For autostart we initially deploy the server without the spring application server property set to autostart a server. Once configured and started for the first time, this configuration is updated (via a configmap) so that subsequently it will autostart

I hope to document/blog and offer up this chart as a simple way to get a basic config working.

However the real focus should switch to the operator work -- the autostart issue was another example of where having something monitoring the servers - which has more context - is what we needed. This is exactly what operators are for

This chart also uses shell scripts - and little error checking - this is painful with curl/script. The operator will address this - most likely in go

[...]

PR #4500, opened 2021-01-20
Helm chart for simple deployment of egeria

planetf1:
See #4454

Deploys single egeria platform, hosting a mds server with a local graph repo & a view server, kafka/zookeeper, and presentation server UI.

This is both simpler than the 'lab' chart, but also a little more customizable. Has been requested by a few teams as a step on from using the labs.

However for real deployments work will progress/revert to that on the operator which offers a much more sustainable and flexible model.

(Draft: still testing, docs tbd)

### DIFF
--- a/open-metadata-implementation/access-services/digital-service/digital-service-api/pom.xml
+++ b/open-metadata-implementation/access-services/digital-service/digital-service-api/pom.xml
@@ -17,6 +17,12 @@
     <artifactId>digital-service-api</artifactId>
     <name>Digital Service OMAS API</name>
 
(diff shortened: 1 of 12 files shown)


planetf1:
Tested
 - Server startup seems resiliant
 - pod can be deleted, and on recovery storage volume is still available, with same data
 - view server & mds server auto start
 
 This seems an appropriate first-pass for this new chart which has fulfilled the initial objectives.
 Feedback can be handled in subsequent issues & intent will be continued through to the operator work

PR #4510, opened 2021-01-22
Refactor Data Engine to use the new generic handlers

alexandra-bucur:


alexandra-bucur:
Commit history is altered by some commit amends made to update the author/sign off that was missing. A new pull request with the same functionality is 4650.

Issue #4519, opened 2021-01-26
Exception in tex/rex pointing at repository proxy


planetf1:
The following exception was seen in tex (master, 15 Jan) when a repository (called cocoMDS4, but is actually a repository proxy for atlas) is pointed at to get a list of types:
```
Mon Jan 25 17:03:38 GMT 2021 cocoView1 Exception OMAG-COMMON-0001 The Open Metadata Service has generated an unexpected java.lang.NullPointerException exception during method getTypeExplorer.  The message was: null
Mon Jan 25 17:03:38 GMT 2021 cocoView1 Exception OMAG-COMMON-0001 Supplementary information: log record id f9892ca0-348e-4e91-b4d2-8b545f9c2b82 java.lang.NullPointerException returned message of null and stacktrace of
java.lang.NullPointerException
at org.odpi.openmetadata.viewservices.tex.api.properties.TypeExplorer.resolveEntities(TypeExplorer.java:146)
at org.odpi.openmetadata.viewservices.tex.api.properties.TypeExplorer.resolve(TypeExplorer.java:118)
at org.odpi.openmetadata.viewservices.tex.handlers.TexViewHandler.getTypeExplorer(TexViewHandler.java:291)
at org.odpi.openmetadata.viewservices.tex.server.TexViewRESTServices.getTypeExplorer(TexViewRESTServices.java:167)
at org.odpi.openmetadata.viewservices.tex.server.TexViewRESTResource.getTypeExplorer(TexViewRESTResource.java:66)
```
The GUI reports this error as:
```
Could not get types for server (status : 400). OMAG-COMMON-400-016 An unexpected java.lang.NullPointerException exception was caught by getTypeExplorer; error message was null
System detail: The system is unable to process the request.
Recommended user action: Review the error message and other diagnostics created at the same time.
```

Unfortunately I don't have detailed configuration from the user for exactly how the server was configured. It appears to be a repository proxy, and does not support all types.  I do not know if CTS has been run against it. I don't have a list of supported types.

The actual code causing the exception is:
```
       // For each entityExplorer add it to its superType's subTypes
         for (String entityTypeName : entities.keySet()) {
             EntityExplorer entityExplorer = entities.get(entityTypeName);
             TypeDefLink superType = entityExplorer.getEntityDef().getSuperType();
             // Not every entity type has a superType
             if (superType != null) { << Ok so it isn't then
                 String superTypeName = superType.getName(); << so is it correct, unset/null?
                 // Find supertype in explorer
[...]

PR #4524, opened 2021-01-26
Adds extensive exception processing to the resolve methods

grahamwallis:

In response to issue #4519 

This cannot fix up a broken or incomplete type gallery, but it does add extensive exception processing and detailed error logging so the problem should be traceable to the particular problem associated with an entity, relationship or classification type def and will pinpoint it to entity def or superType malformation or absence, relationship def or relationship end malformation or absence or classification def, superType or validEntityDef malformation or absence.

Hopefully this will add enough diagnostics that the problem with the type gallery can be identified.


planetf1 (approved):
Thanks. Will be very useful. Will also add to 2.6 as it may aid in debugging other issues there

PR #4533, opened 2021-01-27
#4529 Default to no persistence in lab chart

### LINKED ISSUE #4529: Release 2.6 - revert persistence to disabled for k8s lab chart
In 2.6 I added the capability to have persistence in the lab chart. All egeria data is stored in a volume, and if the pod is restarted this volume gets used. So configuration, cohort registry, local graph all get preserved across runs.

Part of the reason for this is that we have some users of the notebooks that are using the environment as a stepping stone to doing more integration development work, or to experiment with the platform

However the notebooks are written assuming a clean setup each time. For example without this the config notebook will actually start the servers, the CTS will not run clean, and there may be confusing results from other notebooks.

As part of the change, I also switched the default in the k8s chart to use the local graph repository. This is noticeably slower, with some delays of tens of seconds, depending on the k8s infra being used

 In retrospect this could be a problem especially for a first-user experience

I propose to change the default back to no-persistence / inmem - this is only changing 2 default values in values.yaml. The logic to turn on/off is already there 

The additional notebook (egeria-base) will stick with persistence / graph as the intent there is to create a simple single-server auto-starting environment to experiment with

I think this improves usability, means we don't change existing behaviour which could lead to issues being raised/

planetf1:

Fixes #4529 

Defaults to no persistence/inmem repo for lab chart


planetf1:
Fix already incorporated into 2.6, but hit an issue in testing - #4547 - fix added here.

Issue #4631, opened 2021-02-01
Revise purge test cases in CTS


cmgrote:
Are the purge operations (`purgeEntity`, `purgeRelationship`) intended only to permanently remove the current (deleted) version of an instance, or to entirely purge that instance (ie. including all of its historical versions) from a repository?

At the moment, the CTS assumes the latter for assertion27 of TestSupportedEntityLifecycle -- that even after being purged, an entity's previous version can still be retrieved.

The JavaDoc description of the method `purgeEntity` is somewhat ambiguous in this regard: "Permanently removes a deleted entity from the metadata collection.  This request can not be undone."  However, under the assumption that purges would be done for compliance reasons, it seems problematic that the version of an entity just prior to delete (where it has all of the same information as the deleted version, just not the deleted status) can still be retrieved after a purge?

cmgrote:
Confirmed on dev call that the intended behaviour of purge is to remove the entire instance -- current version and all of its history.

cmgrote:
Will update the CTS test cases in TestSupportedEntityLifecycle and TestSupportedRelationshipLifecycle to match this purge behaviour.

cmgrote:
@mandy-chessell just to confirm, this means that the logic of the `InMemoryOMRSMetadataStore` likely needs to change?

- `removeEntityFromStore` is currently called from `purgeEntity` of the InMemoryOMRSMetadataCollection.
- `removeEntityFromStore` removes the entity from the current store; however, it adds it to the `entityHistoryStore` (List).
 
Instead of adding it to the `entityHistoryStore`, the method should presumably iterate through the entire `entityHistoryStore` List and remove any versions of this purged entity (?)

(And same again for relationship instances?)

cmgrote:
Confirmed on today's call that indeed the behaviour of the in-memory repository should change -- I'll make the change and include in the linked PR prior to merging.

cmgrote:
I had originally removed `assertCondition(true, ...` clauses from the `EntityNotKnownException` catches related to historical searches where the entity in question had been purged; however, on further inspection I see that this `assertCondition(true, ...` pattern shows up in many places in the CTS -- so it was not unusual for it to exist here.

Furthermore, without the assertion we end up with `UNKNOWN_STATUS` results for a repository that does correctly implement these purges, whereas a repository that does not support historical search gives a `CONFORMANT_NO_SUPPORT` result.

[...]

PR #4648, opened 2021-02-02
#4631 validate purge operation entirely removes instance

cmgrote:


grahamwallis (commented):
Much better than it was :-) 
I did notice a couple of things. We now do a 'negative' test of historic query - i.e. the entity / relationship is known to have been deleted and purged and we expect (and test) that the historic query does not return it. presumably we should also add a 'positive' historic query test where we do expect to get an earlier version of the entity back.

Also - the assertion (which are always assert(true)) immediately after an operation shouldn't be required in order to validate that the repository supports the operation. It's a while since I looked at it in detail, but I thought that a capability is discovered without that extra assertion by avoiding the 'function not supported' case. Sorry that's a bit vague - if I doesn't make sense let me know and I'll take a longer look.

cmgrote:
> I did notice a couple of things. We now do a 'negative' test of historic query - i.e. the entity / relationship is known to have been deleted and purged and we expect (and test) that the historic query does not return it. presumably we should also add a 'positive' historic query test where we do expect to get an earlier version of the entity back.

Agreed, but suggest it's a separate issue (raised #4666).

> Also - the assertion (which are always assert(true)) immediately after an operation shouldn't be required in order to validate that the repository supports the operation. It's a while since I looked at it in detail, but I thought that a capability is discovered without that extra assertion by avoiding the 'function not supported' case. Sorry that's a bit vague - if I doesn't make sense let me know and I'll take a longer look.

Afraid at this hour it doesn't make sense to me...  Is the simplest test just to remove the assertion (leave the catch with just a comment, no code) and re-test to confirm that it behaves the same way?

cmgrote:
I removed the `assert(true)` statements and the checks still work as expected, so I think that resolves the second point.

PR #4677, opened 2021-02-10
Add table filter for module creation

yevgenmar:
User can define in request body list of tables to include in or exclude from module.
This body is considered as table filter.


mandy-chessell (approved):
Still a lot of missing javadoc.

Issue #4549, opened 2021-01-29
Remove netflix zuul library


lpalashevski:
We need to check the impact of removing netflix zuul library form the ui-spring-chassis spring boot UI application.
As discussed on the community call, there are several reasons why we need to move this way:

- Detaching this extra dependency will help improving the security lifecycle and updates we regularly provide (See #4441 for more information)
- Removing zuul will increase flexibility and allow users to decide for themselves on most appropriate hosting configuration with API gateway component i.e. still go with zull setup or choose different technology such as Nginx for the same purpose.

cc: @sarbull @bogdan-sava @planetf1

sarbull:
removing zuul doesn't increase any flexibility, we can deploy the UI and API separately with the zuul right now, we just have to tell the API where the UI is deployed(by URL). i don't think it can get simpler than this

if we want to remove zuul, we should remove everything that has to do with the UI that it is stored in the `ui-chassis-spring` and linked to the `egeria-ui`

the main idea of zuul was to allow the API to deliver the UI contents through it's context increasing the security level of what we send to the client, the session and all requests that are at Spring level

i don't think of removing zuul as being a step forward, moving this kind of config at server level(nginx) will actually minimize the flexibility of how users deploy..

i would rather go for the two complete separate deployments:
1. API (with CORS activated) (as a facade of the backend / `ui-chassis-spring`)
2. UI deployed statically basically anywhere with a config file / json where you add the URL to the API

we didn't even check to see why we can't update Spring version and we already think of zuul as leftover

lpalashevski:
The driving pricliple here was separation of concerns.

I am not in favour of emmbedding  "reverse proxy" capability inside the backend app (server-cahssis-ui) just because it comes handy from development perspective. When you think from operational and maintainance it is restrictive because of the two reasons mentioned above. Additionally...


About the flexibility: We have to also consider different options and deployment architectures whrere clients can choose for different reasons: security, availability etc aspect that might be already established as practices and standards in an organization. I do not see need for us to prefer any deployment option.

About the maintainance: We are trying to minimize the effort and automate as much as possible to have regular updates on dependencies and spend less time identifying cross dependencies when some library contains vulnerability. In this case we need to update spring but we cannot because of compatibility issues with zuul (we depend on external release cycle).

For pointers why we cannot update see following discussion https://github.com/odpi/egeria/issues/4263#issuecomment-738044829

sarbull:
[...]

PR #4710, opened 2021-02-18
(Draft) [ui-chassis-spring] remove zuul dependencies

bogdan-sava:


planetf1 (commented):
Agree with removing zuul.
If this is merged, what is the impact on the current docs for running the UI
Also what is the effect on the current container environment (labs & simple). Has it been tested?
I think we need to confirm both before merging, especially as our next release is imminent

lpalashevski:
#4549

planetf1 (changes requested):
This PR removes the requirement for zuul - however without zuul the current approach of launching the UI will no longer be the same. If this changes function we either need to ensure each PR is an incremental update without breakage, or we combine them together in a single PR?

Or is the intent that we do #4723 first ?

If they need to be done together, Taking the commit's and cherrypicking may be the easiest approach

planetf1:
We also need to add documentation, as removal of zuul affects how a user would deploy the UI if they are setting up themselves.

sarbull:
This PR is incomplete, check this one for the complete removal of Zuul Module: https://github.com/odpi/egeria/pull/4922

Issue #4711, opened 2021-02-18
Extend CTS with test elapsed times


cmgrote:
In doing repository testing, it can be interesting to have some rough idea of how much time is spent in each particular test case.  We could add "elapsed times" to each test case in order to track this -- giving repository developers some general idea of which test cases are the most cumbersome for their repository to address.

PR #4712, opened 2021-02-18
#4711 adds elapsed time per CTS test case

cmgrote:
(execute phase only for multi-phase test cases)


planetf1:
I've not checked the implementation, but it would be useful if the cts notebook elaborated on the times in a formatted manner

Issue #4702, opened 2021-02-16
Extend OMRS find methods with a timeout exception?


cmgrote:
At the moment our various OMRS-level query methods (`findEntitiesByProperty`, `findEntitiesByPropertyValue`, etc) have exceptions that indicate various functional error conditions, which is all good.

However, in light of our discussions around some particularly expensive queries (`findEntitiesByPropertyValue` with no type restrictions and an "ends-with" or "contains" regex, I'm looking at you!) I am thinking it would be good to extend the set of exceptions that these methods can throw with some sort of timeout-based exception?

In other words, allow the repository (connector) to decide that it has spent enough time trying to execute that query and pro-actively cancel it: if the time it has spent trying to respond to it has exceeded some configurable threshold.  (To avoid needing to change the interface for these methods themselves, that threshold could be something we default to some value (10s? 20s?) but that we allow someone to override in the connection configuration -- in the same way we do at the moment for max page size.)

We could also consider adding further methods (new signatures) that allow this value to be overridden on a per-call basis, in case someone wants to setup a global default timeout threshold, but still allow those accessing the API to knowingly ask for something that they recognise will take longer to retrieve (and therefore accept that the response will be slow).

This gives a sort of on-the-wire "health warning" against a query, whilst simultaneously giving an underlying repository (connector) a conformant means by which to reject a query that might otherwise cause a denial of service (if left to run)?

cmgrote:
Implement a new `RepositoryTimeoutException` as sub-type of `RepositoryErrorException` -- then no impact on the interfaces, but we can distinguish if desired.

Issue #4665, opened 2021-02-03
Overloaded properties in type definitions


cmgrote:
Currently there are some examples (or at least 1) where a given tree of type definitions has the same property defined at multiple levels. One example:

- `SchemaAttribute` defines a property `minimumLength`
- `RelationalColumn` also defines a property `minimumLength`, even though it subtypes `SchemaAttribute`

Aside from the conceptual ambiguity this potentially causes, the tangible impact comes against underlying repositories that are storing these properties: particularly when the property is being stored against a namespace taken from the type definition that defines the property (necessary per discussion in #4647, where the same property could be defined against different type definitions and actually have a different type itself (eg. string vs numeric)). In such cases, particularly in a repository that must handle history (where the typedefs themselves may change over the history of the instances that are stored), having such ambiguity in the model will cause complexity particularly in areas like search: where a request will arrive simply for searching against `minimumLength` and depending on the time period involved we may want to look at only one of these or both.  In addition to the potential for bugs inherent in adding such complexity, there is also likely to be a performance impact associated.

So the question is if there is any reason to retain both properties in such cases?  If not, my proposal would be to retain only the highest-level property (`SchemaAttribute`'s in the example above).  It would therefore still be inherited by and available on `RelationalColumn` but we would resolve any complexity / search performance / ambiguity caused by the overload.

cmgrote:
Stumbled upon further examples (still between `RelationalColumn` and `SchemaAttribute`):

- `length`
- `isNullable`

grahamwallis:
@cmgrote You may want to look at what I do n the graph repo. Just recently I have been trying to wherever possible collapse a search into a single graph traversal, rather than doing multiple and aggregating the results. 

[...]

PR #4713, opened 2021-02-18
Fixes / enhancements as discussed on 18/02 call

cmgrote:
- Removes the overloaded properties mentioned in #4665 
- Adds a new `RepositoryTimeoutException` as discussed in #4702 
- Adds conditions to the CTS search test cases for catching and logging the `RepositoryTimeoutException` as a discovered property of that search method (presuming this neither contributes nor detracts from whether that test scenario is considered a success or failure?)


cmgrote:
Merged quickly given we agreed on the call + so that we have some time for testing prior to next release...

PR #4722, opened 2021-02-22
 Improve certs generated for, and deployed to, Egeria OOTB

### LINKED ISSUE #4670: omag-server-platform-transport-level-security observations
Looking at file : omag-server-platform-transport-level-security. I notice:
1)
TS=${BUILD}/open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources/keystore.p12
should it be 
TS=${BUILD}/open-metadata-implementation/server-chassis/server-chassis-spring/src/main/resources/truststore.p12

2) It contains 
VER=2.4-SNAPSHOT
3) I suggest the example launch script be supplied as a separate file rather than embedded in documentation. 
4) I suggest the references to gensamplecerts.sh and the openssl config be links.

planetf1:
Fixes #4670

## Summary
 * Provides scripts (for egeria developer or as an example) to generate Certificate Authority & certificates for egeria.
 * Scripts assume bash, and availability of 'openssl' and 'keytool'
 * replaces current checked-in certificates with unique certs by usage
 * Updates ui-chassis & server-chassis with new key aliases 
 * facilitates testing / docs of unique trusted certs and certificates (vs 1 file everywhere)
 * Sustainable/extensive vs individual ad-hoc generated certs - inconsistent across developers
 
## In detail
 * Script is provided at open-metadata-resources/open-metadata-deployment/certificates/gensamplecerts.sh
 * Intended to be run by core developer (locally) to update certs currently included by egeria 
 * Also offers an example for other developers, and to support our broader usage in other repos & testing
 * Creates an Egeria root certificate authority, an intermediate certificate authority, certificates for a variety of egeria clients and servers
 * Checks in files as-created (so that new certs could be made off the same CAs)
 * Provides the ui-chassis & server-chassis with their own certs (copied to existing keystore.p12)
 * Each .p12 includes only what is required for it's intended purpose
 * Simple example of truststore with just the root certification authority 
 * Gives us a base upon which to refactor/modify any code we have to work with security properly, aid in deployment, integrate with other UIs with some (!) confidence we're handling the certs appropriately (ie react UI)
 
## Tested
  * checked integrity of certificate chains (relative to generated root)
  * replaced existing egeria certs with new ones
  * ran basic tests with Egeria
  * tested client
  * Imported root CA into firefox - server certs then look good for polymer ui (useful test) and server chassis
  * still need to do some ssl traffic monitoring to check validity/enforcement and 1way vs mutual
  
## Caveats
  * Clearly as the certs are checked in this is insecure, but offers an example
  * openssl can be somewhat tricky to use, so there are many ways to do similar -- often enterprises may have a dedicated team to issue the certs, and external certification authorities will be used, so it's unlikely most people will need to deal with the whole lifecycle which is demonstrated here (minus revocation). There are many many parameters and options which can make the difference between certs working and not & so all we can do is offer a starting point
  
## Todos (not this PR probably)
   * Documentation of how to configure own certs for each component that owns them (important & next job?)
   * Doc how to import root cert ie into firefox
   * Deploying new certs to react UI, polymer UI, connectors (may require changing exact list of certs generated)
   
## Notes for reviewers
   * Review the files actually checked in - which are needed, which are not
   * Rename files/certs appropriately - if suggestions come up
   * Consider what else we should do around this - ie those changes to other repos (ideally in diff PR) - is this a ci/cd activity? Do we externalise all these certs outside the base repo & add when we package?
    
## Future Ideas
   * Addition of cert generation to k8s samples
   * Generate certs on demand (with letsencrypt etc) - maybe easier for cloud service vs code distribution

### DIFF
--- a/open-metadata-resources/open-metadata-labs/common/environment-check.ipynb
+++ b/open-metadata-resources/open-metadata-labs/common/environment-check.ipynb
@@ -34,7 +34,7 @@
     "\n",
     "#\n",
     "# This is the location of the file system that is used in the lab.  It defaults to a \"data\" subfolder in\n",
-    "# the platform's home folder.  Here you will finc the runtime files of the platforms and the data files\n",
+    "# the platform's home folder.  Here you will find the runtime files of the platforms and the data files\n",
     "# used in various labs.  The default value can be overridden using the \"fileSystemRoot\" environment variable.\n",
     "#\n",
     "fileSystemRoot = os.environ.get('fileSystemRoot', 'data')\n",
@@ -1572,7 +1572,7 @@
     "        print (\"Term not updated\")\n",
     "        processErrorResponse(serverName, serverPlatformName, serverPlatformURL, response)\n",
     "    \n",
-    "# create a synonym relationship betwen 2 glossary terms    \n",
+    "# create a synonym relationship between 2 glossary terms    \n",
     "def createSynonym(serverName, serverPlatformName, serverPlatformURL, userId, guid1, guid2):\n",
     "    subjectAreaURL = serverPlatformURL + '/servers/' + serverName + '/open-metadata/access-services/subject-area/users/' + userId\n",
     "    createSynonymURL = subjectAreaURL + '/relationships/synonyms'\n",
@@ -2665,7 +2665,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.3"
+   "version": "3.7.6"
   }
  },
  "nbformat": 4,


planetf1:
Very much interested in comments on this change. 

I am trying here to get close to how certs are really used & use those in what we ship OOTB, ensure certs are unique (to avoid accidental usage) which then in itself should make things easier to test and have confidence we'll work well in the real secure world.

planetf1:
I plan to merge this Tue 9 Mar if no objections are recieved before then

planetf1:
Agreed on community call 9 Mar that we can merge this.

PR #4738, opened 2021-02-26
Update link in egeria-dojo-day-1-4-participating-in-the-community.md

lpalashevski:
As reported by user on Slack looks like the "Code of conduct" link pointing to the .md as file instead html page (suspect because of the different naming convention and how it gets translated to html..). 

Updating the link to the actual github location.

### DIFF
--- a/open-metadata-resources/open-metadata-tutorials/egeria-dojo/egeria-dojo-day-1-4-participating-in-the-community.md
+++ b/open-metadata-resources/open-metadata-tutorials/egeria-dojo/egeria-dojo-day-1-4-participating-in-the-community.md
@@ -24,7 +24,7 @@ This community is stronger if it consists of individuals from different organiza
 and backgrounds.  This means we need a respectful and inclusive
 culture to allow a diverse range of people to participate.
 
-This culture is reflected in the [Code of Conduct](../../../CODE_OF_CONDUCT.md)
+This culture is reflected in the [Code of Conduct](https://github.com/odpi/specs/wiki/ODPi-Code-of-Conduct)
 that every participant must adhere to.
 
 Within the community, contributors are focused on making the contributions


mandy-chessell (changes requested):
The fix is pointing to the wrong file - It need to point to the CODE_OF_CONDUCT.md file in the egeria git repository - not the ODPi specs.

mandy-chessell:
It looks like the link in the CODE_OF_CONDUCT.md file need to be turned into a clickable link

lpalashevski:
maybe we need to do the same for other website md files referencing code of conduct:

Community-Guide.md
Egeria-Operations.md
index.md

README.md (?)
CONTRIBUTING.md (?)

looks like CODE_OF_CONDUCT.md file is not following the same .md->.html pattern like the rest of the files. Not sure if this is some configuration with github pages/jekyll or we  need to rebuild the site but might be because of the different naming convention. Lets dicuss it first.


cc: @planetf1

planetf1:
It may be worth checking our other repos also (perhaps the most viable is to ensure they link back to the content in egeria where possible)

I am only aware of the _config.yml & haven't done any other work with github pages configuration.

From https://nicolas-van.github.io/easy-markdown-to-github-pages/ I see
 - It does confirm just replacing .md with .html should work (which is what the original document does)
  - The example above does explicitly list the additional .md files - I'm not sure what the default pattern is. 
  
I can only suggest some experimenting and research is needed.

planetf1 (changes requested):
see comment added previously

mandy-chessell:
Replacing html with md will create a broken link in github and intelliJ

lpalashevski:
Based on Nigels' input I would suggest to try explicitly including the .md files we have in the root of the repo in _config.yml and see if it generates missing pages. We can also try only with CODE_OF_CONDUCT.md but I am afraid that it might break the site (if in case it forces only included files).

```
include:
  - index.md
  - CODE_OF_CONDUCT.md
  - CodeQualitySecurity.md
  - Community-Guide.md
  - Content-Organization.md
  - CONTRIBUTING.md
  - CONTRIBUTORS.md
  - Egeria-Operations.md
  - GOVERNANCE.md
  - MAINTAINERS.md
  - README.md
 ```

planetf1:
I think there are more fixes needed to address the root issue so suggesting we have this only as draft to reduce chance of breakage

Issue #4740, opened 2021-02-26
Documentation Improvement for the environment settings Jupyter Notebook


NishaNagasimha:

mandy-chessell:
Please help us to understand the purpose of this design change to the evironment-check notebook by providing some explanation in this issue.

Today this notebook is run at the start of each lab to verify the runtime environment and provides some standard functions that are used in the labs.  It is not part of the interactive lab experience.  The intent today is that we will use it as an input to create some python client libraries for the Egeria services.  As such the function definitions are organized into cells that correspond with a back end service.  The collection of functions in the cell will feed the definition of the client library.
The comments associated with the function are to help Egeria maintainers working on the python functions rather than the lab users.

If I understand the PR #4741 , it looks like this existing cell organization is being replaced by an organization that alternates between markdown and code.  The python comments are being replaced with markdown comments.  What is the purpose of this change?

NishaNagasimha:
Hi Mandy,

Thank you for reviewing the issue - #4741 is a repeat of the above issue. When I made changes before, I did not realize it would put it into a new issue. Sorry about that.

- The reason I made changes and added markdown to this notebook is that the comments were like the theory of understanding the setup. 
- Usually, if it is theory or explanation we add markdown in notebooks, to create a separation between code and its corresponding explanations. 
- I understand from your comment that this is not going to remain a notebook? If that is the case, then may I please know if this will become a python script? if so, then I will be sure to remove the markdown and just add the spelling changes!

Thank you!

PR #4741, opened 2021-02-26
Modified documentation for the environment Jupyter Notebook

NishaNagasimha:
**Updates**
- Incorporated Markdown for the comments in this notebook
- Corrected the spelling of `between` and `find`.

### DIFF
--- a/open-metadata-resources/open-metadata-labs/common/environment-check.ipynb
+++ b/open-metadata-resources/open-metadata-labs/common/environment-check.ipynb
@@ -25,7 +25,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 18,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


mandy-chessell:
We should have a discussion about this change.  The environment-check notebook is not meant to be run as part of the labs - it is a placeholder that will be replaced by a set of installable functions.  When that happens all of these changes will need to be reverted to change the markdown back to comments.

The question is, before this happens, is it easier for the notebook writers to read the comments in markdown or as standard Python comments?

NishaNagasimha:
Hi Mandy,

Thank you for reviewing the issue - #4741 is a repeat of #4740. When I made changes before, I did not realize it would put it into a new issue. Sorry about that.

- The reason I made changes and added markdown to this notebook is that the comments were like the theory of understanding the setup.
- Usually, if it is theory or explanation we add markdown in notebooks, to create a separation between code and its corresponding explanations.
- I understand from your comment that this is not going to remain a notebook? If that is the case, then may I please know if this will become a python script? if so, then I will be sure to remove the markdown and just add the spelling changes!

Thank you!

Issue #4704, opened 2021-02-16
Refactor CTS results to reduce Java heap requirements


cmgrote:
At the moment our CTS results are in JSON format, but for a complete run of the current tests the results file (without any whitespace or indentation) weighs in at a hefty 150-200MB.  This is basically unreadable by almost any text editor as a JSON-formatted document.

I've also just looked at a Java profile of running the CTS, and the final retrieval (or any interim retrievals, to check for instance whether the CTS has completed) need to construct this object via Jackson as a `TestLablReportResponse` -- this takes 800+MB of memory on the heap to do each time.

We also have a few issues open on areas where additional tests are to be added to the CTS, which would only increase these sizes.

Should we consider some alternative approach to storing the results of the CTS?  The current approach already seems at (beyond?) the limits of what we'd reasonably expect a REST-based interface to handle...

cmgrote:
Suggestion is to break down the singular large query into an iterative set of smaller queries against each individual test case to get the details -- perhaps leave the `TestLabReportResponse` just to give the overall profile summaries.

(Need to confirm that an interface for these smaller queries already exists, and probably provide a simple script that will make the appropriate iterative calls to retrieve all of the details of the test cases.)

planetf1:
I think it would be good to refactor - this caused us an issue before - see https://github.com/odpi/egeria/issues/2497

PR #4748, opened 2021-02-28
#4704 splits CTS results into summaries and details

cmgrote:
... with multiple API calls.

Also adds elapsed times on a per-method basis, to allow inspection afterwards for which methods seem to be performing well or not for a given repository -- the timings are calculated only for the MetadataCollection method calls, not any of the surrounding logic contained in the CTS itself.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -41,7 +41,7 @@ apply plugin: 'nebula-aggregate-javadocs'
 allprojects {
 
     group = 'org.odpi.egeria'
-    version = '2.7-SNAPSHOT'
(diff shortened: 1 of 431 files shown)


cmgrote:
Confirmed that with these changes, the maximum CTS addition to the heap is now ~150MB (where previously for the same it grew to over 800MB).

cmgrote:
Since I've left the old endpoints, methods, payloads, etc for backwards compatibility I will go ahead and merge (giving additional testing time prior to next release).

Issue #4734, opened 2021-02-24
Release 2.7


planetf1:
Create 2.7 release 

- [x] slack post
- [x] agree in team call (late Feb 2021)
- [x] Create branch for 2.7
- [x] Reassign any issues not being worked on to the next release
- [x] Update master to **2.8-SNAPSHOT** 
- [x] Update ref to presentation server image in chart(s)
- [x] Ensure any remaining fixes are merged 
- [x] Update branch version from 2.7-SNAPSHOT to 2.7
- [x] Remove release notes from branch 
- [x] update release notes & rename 
- [x] Verify notebooks (see #4842)
- [x] Check polymer UI 
- [x] Check node (ps) UI (see #4843)
- [x] Run CTS 
- [x] Run FVTs 
- [x] Check samples 
- [x] Check swagger doc 
- [x] Verify docker image/charts/compose 
- [x] create release candidate (manual github action)
- [x] Verify RC packaging on MC
- [x] Fix up maven artifacts as needed
- [x] Create final github releases (launches pipeline)
- [x] Verify artifacts all on maven central 
- [x] Release on maven central
- [x] ensure any fixes ported to master

planetf1:
Note - this is the first release that will aim to use the new release pipeline - https://github.com/odpi/egeria/pull/4733
There may be a few teething issues & it's very likely any errant maven artifacts failing validation will need to be
 - fixed (if fixable ie containing code)
 - removed (if empty/no code)
 - OR publishing of artifact disabled - see #4464

planetf1:
master will be branched for 2.7 on Sun 27 Feb

planetf1:
Testing new release pipeline prior to continuing with testing ...

planetf1:
Graph CTS results:
```
Date of conformance test:          2021-03-02T12:09:48.143+00:00
Number of tests that were run:     4965
Number of tests that failed:       0
Number of tests that were skipped: 0


id | name | description | profilePriority | conformanceStatus
-- | -- | -- | -- | --
0 | Metadata sharing | The technology under test is able to share met... | MANDATORY_PROFILE | CONFORMANT_FULL_SUPPORT
1 | Reference copies | The technology under test is able to save, loc... | OPTIONAL_PROFILE | CONFORMANT_FULL_SUPPORT
2 | Metadata maintenance | The technology under test supports requests to... | OPTIONAL_PROFILE | CONFORMANT_FULL_SUPPORT
[...]

PR #4840, opened 2021-03-02
Correct presentation server image version for release

planetf1:

See #4734

Switches version of presentation server docker image from 'latest' to the release label ie 2.7 or 2.8-SNAPSHOT as with our other images. We originally left as latest as the egeria react ui was not building versioned docker images, but that is now addressed.


planetf1:
The versioning of the presentation server images is aligned with npm ie 2.8.0-rc.0 so differs from the base
Since these charts will move to a new repo and we may have different versions of different dependencies it seems prudent to leave master as latest for now, and just freeze the version at release time. So closing this PR

PR #4849, opened 2021-03-03
Cut-over 2.7 release types to a new extensible master

cmgrote:


cmgrote:
@mandy-chessell if you can have a quick look and confirm I did this correctly?

Hoping it was as simple as what I did here (just created a new class with the `2_7` name that's a copy of the existing `OpenMetadataTypesArchive`, and then changed the existing `OpenMetadataTypesArchive` to build its original types using the `2_7` class as its previousTypes) (?)

cmgrote:
Looks like this is already done now via #4860 -- closing

mandy-chessell:
Sorry @cmgrote I did not see that you had done this already - if it is any consolation, you commit was completely correct :)

cmgrote:
No worries -- took a matter of seconds to make the change and at least I can be confident of doing it properly next time should the need arise!

PR #4884, opened 2021-03-09
Initial super-types for relationships

cmgrote:
The build runs through without any failures, including the FVTs, but I cannot imagine that this has not broken _something_...


cmgrote:
Hmm, so one initial thing I notice:

- since we said properties would never be deleted, there isn't a way to do this via the `TypeDefPatch`
- since we also have said we don't want to deprecate properties at a subtype level that are inherited from a higher level (confusing), we should delete them
- I think this means that I'll need to retroactively go through the various `OpenMetadataTypeArchive_` classes and make the changes to properties (where they are now on a supertype) directly in those classes?  (I'll still setup the supertype through a TypeDefPatch specifically on 2.8, though?)

planetf1:
Will try and kick this off again after fixing up the pipelines/checks.

planetf1:
I think the best thing is to make a new PR or push an update -- I think I've corrected the checks now

planetf1:
`git commit -m "retrigger checks" --allow-empty` will trigger rechecks - doesn't look like close/open does remove 'bad' checks from the old definitions

cmgrote:
Merging this initial set of changes for now, given we're on a separate branch already...

PR #4888, opened 2021-03-09
Miscellaneous fixes

mandy-chessell:
This PR includes the code for the OriginSeeker Governance Action Service and a number of fixes and doc improvements.


open-metadata-implementation/adapters/open-connectors/governance-action-connectors/pom.xml
```diff
@@ -50,10 +50,21 @@
...
            <groupId>commons-io</groupId>
            <artifactId>commons-io</artifactId>
-       </dependency>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <exclusions>
+                <exclusion>
```
planetf1, line 62: Out of interest, is this exclusion because you don't require the annotations, or did you hit a clash (the latter case has caused a latent issue in the past)?

open-metadata-implementation/adapters/open-connectors/governance-action-connectors/build.gradle
```diff
@@ -14,7 +14,7 @@ dependencies {
     implementation 'org.slf4j:slf4j-api'
     testImplementation 'org.testng:testng'
     implementation 'com.fasterxml.jackson.core:jackson-annotations'
-
+    implementation 'com.fasterxml.jackson.core:jackson-databind'
```
planetf1, line 17: the pom excludes annotations, the maven version doesn't - so they are inconsistent? I presume they should be?

open-metadata-implementation/adapters/open-connectors/governance-action-connectors/pom.xml
mandy-chessell: I copied it from another pom file - assumed you had added it - can take it out if not needed.

Issue #4699, opened 2021-02-15
FVT Server configuration and data appearing in distribution


mandy-chessell:
It looks like the files created by the FVTs are being included in the assembly.
<img width="300" alt="Screenshot 2021-02-15 at 20 24 09" src="https://user-images.githubusercontent.com/12354195/107990404-53826c80-6fcc-11eb-9cb1-690e61ec41eb.png">

planetf1:
Good spot. They shouldn't be there.
Could add a simple exclusion, though I would like to change the working directory the FVTs use, to be unique per test (this will also address a problem with the gradle based FVTs which are currently a draft PR)

planetf1:
Currently fixing this by ensuring the working directory for each test is unique.. ie written to the target dir for the test in question. 

This means
 * Test could run in parallel
 * Each test has it's own clean repo/config
 * No contamination in the assembly/distribution directory (it was the latter that was contaminated, the .gz was fine)
 
 In the process observed that the tests depend on multiple sources of data matching
  a) Ports, URL , id etc from the pom.xml (and in future gradle)
  b) Similar within the .groovy file
  c) Java code which was using FVTConstants
  
 In fact c) was one cause of the gradle tests failing (still a PR). 
 
 I propose to modify the test driver code to use parms passed through from the test driver (surefire or gradle) rather than hardcoded values. This can be via properties. However in the case where those properties aren't passed we can use the current FVTConstants as defaults.
 
 This should allow for future parallelism, better isolation, easier modification and should work for maven and gradle.

PR #4896, opened 2021-03-10
#4699 Ensure FVT tests can be supplied port dynamically by test driver (maven, gradle)

planetf1:

Fixes: #4699 
 - Ensures that each FVT module uses a working directory local to the module to
   * keep data isolated (tests not depending on prior configuration)
   * Avoid introducing data files into the assembly (original reason for #4699
 - Utilizes a new system property which is set within maven (and in future gradle) to override the port number hardcoded
 - Defaults to existing values if property not set - so that tests will still work in whatever way a developer may have been using them before outside the control of the test runner
 - changes references to lib & jar as per change to working directory
 - adds required truststore link (as we can't depent on current directory in assembly now) [note - ssl validation is still disabled, but this is required to launch)
 - removes manual 'cleanup' plugin definition - as all data files are now under target, default maven rules will do the cleanup
 - slightly reworked the data engine fvt -- this uses a property file to pick up configurations, so the pom still picks up from there, but the parsing of url/host is slightly reworked (and combined) to make it easier to change ports
 - uses unique (albeit explicitly put in pom) ports to allow for potential parallelism - whilst this may not be desirable from a load perspective it addresses the logical problem seen when gradle was invoking tests.
 
 Not done
   - this fix does not actually address gradle FVTs, but is a prereq. See #4098 
   - there are still various hardcoded references for server names, user ids - no further refactoring done
   - server property name is referenced as literal string

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/IsA.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/relationships/IsA.java
@@ -50,7 +50,7 @@ public class IsA extends Line {
     private String source;
 
     public IsA() {
-        super("Isa", "50fab7c7-68bc-452f-b8eb-ec76829cac85", LINE_END_1, LINE_END_2);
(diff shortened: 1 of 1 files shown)


planetf1:
Gradle build failing as I omitted to add the commons library to the build.gradle ...

Issue #3967, opened 2020-10-27
deployedImplementationType does not seem appropriate name in BusinessCapabilityType


mandy-chessell:
Recent change renamed the "type" property in BusinessCapability to "deployedImplementationType".  Something like "businessCapabilityType" seems more appropiate.
![0440-Organizational-Controls](https://user-images.githubusercontent.com/12354195/97280488-3f0b4000-1834-11eb-9b67-42eadd54b3bd.png)

mandy-chessell:
FYI @grahamwallis

grahamwallis:
@mandy-chessell Well found! :-)

grahamwallis:
Presume that we should still clear this up.

PR #4914, opened 2021-03-16
Fix repository refresh messages and primary key classifications

mandy-chessell:
This PR contains two key fixes:
* Another attempt to prevent refresh messages being processed in an repository proxy
* Fix to the Data Manager OMAS method for setting up the primary key (parameters were muddled on request to set up the classification.

In addition there are a number of fixes to the autmated curation lab (still not finished)


mandy-chessell:
Also removed the obsolete function from Discovery Engine OMAS that configured the discovery server (which no longer exists).

mandy-chessell:
Also fixed issue #3967 by updating the type property in BusinessCapability

PR #4926, opened 2021-03-19
Implement nginx proxy for lab helm chart

planetf1:
- Adds new nginx container
- configures nginx using template
- directs front end requests at static UI, backend at ui server chassis
- utilizes secrets for tls cert/key and configmap for nginx template
- environment provided by k8s deployment
- certs currently copied in-situ into a secret yaml file (not a secure approach - initial step)

Currently testing


planetf1:
Debugging the specific cert chosen for now:
```
2021/03/19 08:51:44 [emerg] 1#1: cannot load certificate "/etc/nginx/ssl/tls.crt": PEM_read_bio_X509_AUX() failed (SSL: error:0909006C:PEM routines:get_name:no start line:Expecting: TRUSTED CERTIFICATE)
nginx: [emerg] cannot load certificate "/etc/nginx/ssl/tls.crt": PEM_read_bio_X509_AUX() failed (SSL: error:0909006C:PEM routines:get_name:no start line:Expecting: TRUSTED CERTIFICATE)

```
The primary purpose of this initial change is to get something working to allow for zuul removal.
A later PR may be used to ensure a user specified secret may be used instead

planetf1:
@lpalashevski How familar are you with nginx's certificate needs? Currently I am basing the encoded certificate on those we generate under open-metadata-resources/open-metadata-deployment/certificates as a proof point, Just trying to debug the specific nature of the problem with the certs (where every tool seems different)

planetf1:
To view the certs one can use 
```
kubectl get secret lab-nginx-ssl -o json | jq '.data | map_values(@base64d)'
```

planetf1:
The error was caused by 2 problems in the helm template for nginx
 a) I had only a tls.cert, with the contents of tls.key (mapping error)
 b) I did not specify the private key password, and have added a new volume for this
 
 Not secure, but proving the concepts.

planetf1:
Current status - the existing UI does not work. Nginx is now being configured correctly and returns the same result.
cc: @bogdan-sava @lpalashevski @sarbull

planetf1:
See also https://github.com/odpi/egeria-ui/pull/111
If we make a change that involves the static UI switching to a different port (as it does in that PR) then I will need to adjust the ports here accordingly

Note also that once merged I still need to fix up docker compose before we can remove zuul

Finally note that in any case the polymer UI is currenly not working as per https://github.com/odpi/egeria-ui/issues/110

planetf1:
This is ready to be merged.
As per #4549 still need to update compose & make some doc updates in readmes, notebook.

lpalashevski (approved):

PR #4924, opened 2021-03-18
Revert "Remove Zuul Module #4549 #3795"

sarbull:
Reverts odpi/egeria#4922

PR #4930, opened 2021-03-22
Revert "Revert "Remove Zuul Module #4549 #3795""

sarbull:
Reverts odpi/egeria#4924


planetf1:
We should wait until the docker compose change is also done, as noted in the zuul issue

planetf1 (changes requested):
As per original issue, next change is to ensure compose works, then we can flip the switch without breakage.

lpalashevski:
> As per original issue, next change is to ensure compose works, then we can flip the switch without breakage.

Noted. Waiting on update from your side.

planetf1:
Once https://github.com/odpi/egeria/pull/4936 is reviewed and merged I think we're good to go with this one
(and will need to then retest k8s & compose just in case)

Given we're close to branching we may wish to hold off on the zuul removal until after the release, though I am ok with going ahead if you prefer. I would leave spring dependencies as is though

PR #4936, opened 2021-03-24
Add support for nginx proxy to docker-compose

planetf1:
* Adds nginx proxy to docker-compose 
* Polymer UI should now be accessed via https://localhost:10443 in a docker-compose environment
* Docs to be updated in separate PR


open-metadata-resources/open-metadata-deployment/compose/tutorials/.env
```diff
@@ -6,6 +6,10 @@ egeria_version=2.8-SNAPSHOT
 #
 # Version of egeria-react-ui to use
 egeria_react_ui_version=2.7.0-rc.0
+
+# Version of egeria-ui to use
```
sarbull, line 10: 👍

sarbull (approved):

lpalashevski (approved):

lpalashevski:
Looks good and I already tested the similar setup locally usign the same egeria-ui docker image and local nginx configuration that is almost identical with this.

open-metadata-resources/open-metadata-deployment/compose/tutorials/.env
planetf1: I've set to latest for now.. this needs some refinement moving forwards. for the next release we can consider setting it to a fixed value after the branch - open to whatever you are comfortable with.

planetf1:
> Looks good and I already tested the similar setup locally usign the same egeria-ui docker image and local nginx configuration that is almost identical with this.

Super.  I'm think we should discuss deprecating the compose environment moving forward, and focus on k8s as it's more flexible. to mitigate the k8s knowledge needed I'll improve/add docs around using microk8s or kind, which works well in a constrained environment and can run in docker

PR #4720, opened 2021-02-22
Create and use in OLS new AL endpoint for providing the context for assets

alexandra-bucur:
The changes include:
- new endpoint in Asset Lineage OMAS that retrieved the asset context for relational tables and data files
- stop publishing the ASSET_CONTEXT_EVENT in Asset Lineage
- ask for the asset context in OLS using the new AL endpoint and store the information into the graph through the Janus connector
- new configuration for OLS including the connection to Asset lineage (server, base URL, user, password) and jobs information (name and interval).

Signed-off-by: Alexandra Bucur <alexandra-nona.bucur@ing.com>

PR #4942, opened 2021-03-26
Create new OLS background job to query AL for updates

alexandra-bucur:
Based on the branch from #4720. We need to merge it after merging the first one.


lpalashevski (changes requested):
With the current implementation we are triggering "Initial Load" every time the job is executed. 

Because this is not optimal, we need to add logic to request delta updates (only publish entities since last successful update request). 
This can be done by passing "updatedAfterDate" parameter in publishEntities call with following logic:

1. On job run we read "assetLineageLastUpdateTime" from ols; if value present we pass it to "updatedAfterDate" parameter when calling publishEntities.
2. If "assetLineageLastUpdateTime" has no value this means it runs first time, then we call publishEntities without date parameter (initial load).
3. Once job finishes successfully we update "assetLineageLastUpdateTime" in ols (value is the job start time); If the call fails we do not save/update the time in ols.

Note that "updatedAfterDate" was added to pubishEntities endpoint in asset-lineage server but client interface is not up to date. We need this in place first.

lpalashevski (approved):

PR #4953, opened 2021-03-30
Update classification management over cohort

mandy-chessell:
This PR fixes a number of bugs in the way that classifications are saved and distributed over the cohort.


mandy-chessell:
The other change I made was to change the repository services so that it no longer catches Throwable (it catches Exception instead).

PR #5046, opened 2021-04-08
Update asset

yevgenmar:
Initial implementation of asset update.
Synchronization PUT end point.

merged as https://github.com/odpi/egeria/pull/5073


mandy-chessell (changes requested):
The REST API is using a String to represent a JSON structure.  This should be represented as a Java class with Jackson annotations to enable the JSON stricture ot be understood and maintained.

PR #5060, opened 2021-04-15
Add Lombok to Asset Catalog

alexandra-bucur:


planetf1 (commented):
The gradle and pom changes look very good so far.
I note that intellij now includes lombok support by default as of 2020.3

alexandra-bucur:
@planetf1 thank you :) Indeed, I have the same Intellij version and it worked so I forgot that some versions ago a plugin was needed :)

alexandra-bucur:
I added in the latest commit a global lombok config in the egeria root folder. It contains a property that indicates lombok to generate getters for boolean members starting with "get" prefix, as discussed during the community meeting.

Getter and setters generated by lombok as a result to having @Getter or @Setter annotations on class, can be overwritten simply by manually writing a getter/setter method in the class.

@Data is a convenient shortcut annotation that bundles the features of @ToString, @EqualsAndHashCode, @Getter / @Setter and @RequiredArgsConstructor together, so in future we ca use this wherever applicable.

More info on lombok can be found here: https://projectlombok.org/features/all.

pom.xml
```diff
@@ -2984,6 +2986,12 @@
                 <version>${lettuce.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>org.projectlombok</groupId>
+                <artifactId>lombok</artifactId>
+                <version>${lombok.version}</version>
+            </dependency>
```
planetf1, line 2993: I think there's still an open question here about having scope=provided so that we exclude lombok from any assemblies since it is a build time only component.

planetf1 (changes requested):

pom.xml
alexandra-bucur: I added an exclusion for lombok at the pring-boot-maven-plugin level in order for lombok to stop appearing in the jars. It looks like it works for me. What do you think about it?

planetf1: I think that would work for the spring boot maven plugin -- which is fine for your current use - but as we extend the use of lombok more broadly (which is appealing) we really want an approach that works with as little explicit action by the new user as possible. I think the 'scope' does this -- and is also what seems to be recommended more generally?

I know we still have the question about the dependency checker -- but that's somewhat of a limitation in that tool - for now could add an exclusion for lombok anyway, perhaps with a comment to explain why

alexandra-bucur: Ok then, I put provided back :)

planetf1: good by me :-)

planetf1 (approved):

marius-patrascu (approved):

Issue #4917, opened 2021-03-17
Windows build failure (merge)


planetf1:
The Windows Egeria Merge build is failing

https://github.com/odpi/egeria/actions/workflows/windows-daily.yml

For example for one specific log:
See https://github.com/odpi/egeria/runs/2126504680?check_suite_focus=true

This looks like an issue with launching the chassis for FVT tests. I made changes in this area recently under #4896

Linux/macOS are fine

planetf1:
Trying to debug this I also hit build problems in the conformance suite, which seemed to be caused by symbolic links .
Will add notes to readme for 2.9 & leave to future release

planetf1:
One option is to consider whether we advise windows users to work within WSL2 (the linux environment supported in windows which can run a full ubuntu etc environment) to avoid these platform specific issues and improve performance 3x

The one gotcha will be around IDE integration.

Are there any windows based developers that would like to take this issue on to work out what works best? Most of the core team are using MacOS and/or Linux

planetf1:
Building on a clean machine, the first issue hit is in the conformance suite:

```
[INFO] 
[INFO] --- apache-rat-plugin:0.13:check (rat-check) @ open-metadata-conformance-suite ---
[INFO] Added 3 additional default licenses.
[INFO] Enabled default license matchers.
[INFO] Added 2 custom approved licenses.
[INFO] Will parse SCM ignores for exclusions...
[INFO] Finished adding exclusions from SCM ignore files.
[INFO] 66 implicit excludes (use -debug for more details).
[INFO] 46 explicit excludes (use -debug for more details).
[INFO] 176 resources included (use -debug for more details)
[INFO] Rat check: Summary over all files. Unapproved: 29, unknown: 29, generated: 0, approved: 147 licenses.
```

The rat report shows :
```
29 Unknown Licenses

*****************************************************

Files with unapproved licenses:

  C:/Users/nigel/IdeaProjects/egeria/open-metadata-conformance-suite/docs/performance-workbench/test-cases/repository-classification-search-performance-test-case.md
  C:/Users/nigel/IdeaProjects/egeria/open-metadata-conformance-suite/docs/performance-workbench/test-cases/repository-classification-update-performance-test-case.md
  C:/Users/nigel/IdeaProjects/egeria/open-metadata-conformance-suite/docs/performance-workbench/test-cases/repository-entity-classification-performance-test-case.md
[...]

Issue #4259, opened 2020-12-03
Git Clone on Windows advisory


exonianp:
When downloading the project on Windows one is confronted with the issue of the long file names. 

`? Choose default git protocol HTTPS
- gh config set -h github.com git_protocol https
✓ Configured git protocol
✓ Logged in as exonianp

C:\Projects\Egeria>gh repo clone odpi/egeria
Cloning into 'egeria'...
remote: Enumerating objects: 22, done.
remote: Counting objects: 100% (22/22), done.
remote: Compressing objects: 100% (22/22), done.
remote: Total 297927 (delta 13), reused 1 (delta 0), pack-reused 297905
Receiving objects: 100% (297927/297927), 168.42 MiB | 31.26 MiB/s, done.
Resolving deltas: 100% (154982/154982), done.
warning: unable to access 'open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/configuration-encrypted-file-store-connector/src/main/java/org/odpi/openmetadata/adapters/adminservices/configurationstore/encryptedfile/.gitattributes': Filename too long
error: unable to create file open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/configuration-encrypted-file-store-connector/src/main/java/org/odpi/openmetadata/adapters/adminservices/configurationstore/encryptedfile/DocStoreErrorCode.java: Filename too long
error: unable to create file open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/configuration-encrypted-file-store-connector/src/main/java/org/odpi/openmetadata/adapters/adminservices/configurationstore/encryptedfile/EncryptedFileBasedServerConfigStoreConnector.java: Filename too long
error: unable to create file open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/configuration-encrypted-file-store-connector/src/main/java/org/odpi/openmetadata/adapters/adminservices/configurationstore/encryptedfile/EncryptedFileBasedServerConfigStoreProvider.java: Filename too long
error: unable to create file open-metadata-implementation/adapters/open-connectors/configuration-store-connectors/configuration-encrypted-file-store-connector/src/main/java/org/odpi/openmetadata/adapters/adminservices/configurationstore/encryptedfile/package-info.java: Filename too long
[...]

PR #5084, opened 2021-04-26
#4917 recommend use of WSL2 for Windows builds & remove scheduled windows build

planetf1:

A number of issues are currently seen with trying to build Egeria on Windows

- git long pathnames need to be configured (we had documented this)
- in some cases a System Registry setting was needed to provide for long pathnames at the OS level
- git symbolic links do not work properly (caused issues with CTS)
- FVTs were not working - possibly due to path seperators or other path limitations
- WIndows builds were slow - 2 - 3x slower. This is common to all Java apps, and seems to be down to Win32 File I/O APis

Meantime Microsoft have excellent support for Linux within WIndows itself using Windows Subsystem for Linux. In it's latest version it runs a complete Linux kernel and is very capable of running standard linux tools. Performance is comparable with Linux natively, and the above limitations do not apply. Further, leading IDEs such as IntelliJ & VSCode allow for full native IDE use whilst compiling and running within the Linux environment

With no further comments received on slack threads, and no further updates/requests in related issues I am proposing we now stop documenting use of 'native' Windows, and instead provide pointers to use WSL2 instead.

Fixes: #4917 #4259

### DIFF
--- a/developer-resources/tools/Maven.md
+++ b/developer-resources/tools/Maven.md
@@ -41,12 +41,12 @@ You can check if Maven installed on your machine by running the command `mvn --v
 
 Maven can be installed:
 
-- On Windows, by downloading the [binary zip archive](https://maven.apache.org/download.cgi) and
-[following Maven's installation instructions for Windows](https://maven.apache.org/guides/getting-started/windows-prerequisites.html).
 - On MacOS, by first installing the [HomeBrew](https://brew.sh) package manager and then running
 `brew install maven` from the command-line.
 - On Linux operating systems, by using your distribution's package manager (`yum install maven`, `apt-get install maven`, etc).
-
+- On Windows, you should use [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) Version 2 or above, install an appropriate Linux distribution, and follow
+the instructions for Linux.
+  
 Ensure you are using version 3.5.0 or higher in order to build Egeria.
 
 ----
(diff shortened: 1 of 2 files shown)


exonianp:
I totally agree with you Nigel.

For me the primary incentive to use Microsoft OS to install Egeria was to locally access MS SQL DBs (with Stored Procedures that prohibited the migration to another DB). 
As MS SQL is now fully supported on Linux - never mind the Linux support via WSL you rightly mentioned above- there is no reason for a Windows install.

Rather use the resources in better supporting/documenting the Linux installations so that people with limited resources can have a trouble free experience while installing in Linux e.g. see the issue with the security certificates ([reference to gensamples.sh is missing in the documentation]( https://egeria.odpi.org/open-metadata-implementation/admin-services/docs/user/omag-server-platform-transport-level-security.html) ; though  [just discovered them in a thread here](https://github.com/odpi/egeria/tree/5a11a4f215e8fadda51db267ed7c990333b5facd/open-metadata-resources/open-metadata-deployment/certificates)  )

Ideally, it will be great if one can start with a default installation of Linux (say Ubuntu 20.04) and just go through the installation steps of all the prerequisites that will ensure that Egeria and all its components get installed and deployed with no hassle  (e.g. installation of Java, easy/ open/lib ssl) so that the adopters can focus on the product itself.

planetf1:
Agreed in Team meeting 27 April

PR #5087, opened 2021-04-27
Maven repo settings

wbittles:
These settings where pointing to the old artifactory repos and where causing issues for intelliJ,


planetf1 (approved):
This file was previously used by our Azure pipeline builds to provide links to JFrog artifactory.

This was done through 'server' definitions and a series of profiles (which were not activated by default).

Since we have migrated to github actions we do not currently need any non-default values so I think it reasonable to delete this file to cleanup. Should we need something in the future it can be added back

planetf1 (approved):

Issue #3969, opened 2020-10-27
Capture code coverage data


planetf1:
Currently code coverage is captured from our 'Sonar' test runs 

This is done by enabling 'jacoco' in the build. This attaches an agent to the jvm used by surefire for running Unit Tests. Coverage data is captured and saved in a file which is then uploaded to sonar

We need to look at 
 - [ ] How is this affected by the move of our PR builds to github actions - is it still working
 - [ ] could we capture code coverage in actions instead
 - [ ] how are results published and made accessible
 - [ ] Are we sure FVT coverage data is captured. From the client (failsafe)? How about the server (launched from groovy)
 - [ ] Ensuring developer guidelines updated with information on finding and collecting code coverage data

planetf1:
I have added an example of capturing code coverage data from gradle at : https://github.com/odpi/egeria/pull/5159

The challenge with a multi module project is aggregating the results. In maven this is particularly difficult - and would almost certainly be easier to do via a shell script post-build

Gradle is more flexible, and where we will likely focus future build improvements, so an initial implementation is added there

Still need to look at
 - [ ] setting up a merge gradle build 
 - [ ] integration with external tools like codecov (upload a file to ? )
 - [ ] where else could we publish the results to
 - [ ] whether any other in-build processing could create a better build format
 - [ ] capturing data from FVTs
 - [ ] what we can/should do in maven

planetf1:
Recently I tried to add capture of code coverage data in FVTs (maven only) (above). This involved changing the configuration of the agent prep & report to surround the integration test phase, as opposed to unit tests.

However although I can see the java agent being prepared
```
[INFO] --- jacoco-maven-plugin:0.8.7:prepare-agent (agent) @ asset-consumer-fvt ---
[INFO] jacoco.argLine set to -javaagent:/Users/jonesn/.m2/repository/org/jacoco/org.jacoco.agent/0.8.7/org.jacoco.agent-0.8.7-runtime.jar=destfile=/Users/jonesn/IdeaProjects/egeria
.maven/open-metadata-test/open-metadata-fvt/access-services-fvt/asset-consumer-fvt/target/jacoco.exec
```
And then I can see that the exec plugin that is used to launch the server chassis is specifying the agent ie:
```
[INFO] --- process-exec-maven-plugin:0.9:start (chassis-start) @ asset-consumer-fvt ---
[INFO] arg: java
[...]

PR #5159, opened 2021-05-05
Capture additional code coverage data

planetf1:
- captures code coverage data for unit tests (only) [fvt can follow] for gradle build
 - Since we only have a verification build in gradle, it is done there, ie for a PR - not a merge
 - output is attached as a build artifact. It includes a CSV, as well as a tree of html files
 - Report is only ordered by package name. There is no breakdown by artifact/directory (the plugin does not support this)
 
 See #3969 for more info


planetf1:
After reviewing results, some further work is needed. the task needs to be run explicitly (easy - run './gradlew codeCoverageReport - except that then the FVTs would run twice)
Need to review the network of task dependencies to determine the best to add to ensure the report is built at the correct time. The FVTs themselves probably need modifying so that they run when the assembly is rebuilt or tests modified, but not at other times
Moved to draft

planetf1:
- Added support to maven poms for our FVT tests to try and capture code coverage data there-
- removed sonar exclusions from FVT tests (I think caused by original heavy types testing code)

planetf1:
Merging, as this should improve sonar reports, but due to the pipeline setup we'll only verify once merged. Further changes to improve code coverage capture can follow in other PRs.

Issue #5207, opened 2021-05-20
Lombok - coding usability


planetf1:
Lombok has recently been added to Egeria.

Firstly, we have now lost the embedded javadoc that is being packaged and sent to maven central, which prevents debugging of egeria properly.

Secondly
When compile errors, warnings, notifications occur at the CLI or in an IDE, the errors reported relate to the delomboked version of the code ie:

18:23:04,463 [WARNING] /Users/jonesn/IdeaProjects/egeria.maven/open-metadata-implementation/governance-servers/data-platform-services/data-platform-services-connector/target/delombok/org/odpi/openmetadata/dataplatformservices/api/model/DataPlatformDeployedDatabaseSchema.java:[5,68] org.odpi.openmetadata.accessservices.dataplatform.properties.DeployedDatabaseSchema in org.odpi.openmetadata.accessservices.dataplatform.properties has been deprecated

At least 3 developers in the last few weeks have then gone to fix the delomboked code, and got confused as to why the fix didn't work....

An explicit delombok is needed to correctly produce javadoc, but I would suggest this confusion is adding to the cognitive load of developers working on the platform which could trap new contributors to the platform.

I think we should:
1. Remove lombok plugin activation from the top level pom and ONLY use it where it is CURRENTLY needed (I was pushing for lombok to be available to any component that need it so take some of the blame for not scoping it more narrowly, but at least we have discovered it's characteristics better)
2. Propose that no further use of lombok is made until we can address the IDE usability issues
If not consider removing lombok (the existing delombok’d source can be used for those modules currently using lombok annotations)
3.  Investigate if any build changes can mitigate this effect
4. Make a decision on continuing to broaden lombok usage, maintain the status quo, or remove

planetf1:
Agreed in community call 20 Mar that we will go with #1 as a short term measure until we learn more about these issues (cc: @alexandra-bucur @lpalashevski @mandy-chessell @davidradl )

planetf1:
I plan to make change 1 today

planetf1:
@alexandra-bucur I've done step 1 - assigning for you to look at any of the remaining aspects/see if we can reduce the impact before expanding usage further?

alexandra-bucur:
@planetf1 I added the https://github.com/odpi/egeria/pull/5216 in the meantime in order to avoid the duplicate class errors in Intellij.

I will continue the investigation further for this issue :)

planetf1:
[...]

PR #5216, opened 2021-05-24
Add back Lombok to Egeria pom

alexandra-bucur:
Fixes https://github.com/odpi/egeria/issues/5207 regarding:
- javadoc is generated and present after the Maven build
- duplicate classes error is fixed without the help of maven-clean-plugin. There is a need to have the following Intellij setting: Preferences -> Build, Execution, Deployment -> Build Tools -> Maven -> Importing -> "Generated sources folders" with value "Don't detect". For example, for asset-catalog-api in Project Structure -> modules -> "Source Folders" we need to have only "src/main/java" and in "Excluded Folders" only "target". If things are too messy in all modules a git clone might solve things.


planetf1 (changes requested):
I removed the invocation of clean in the compile phase as part of https://github.com/odpi/egeria/pull/5214 as it appeared to be removing the java code (ie after lobok processing - in the delombok directory) from the tree BEFORE javadoc was run.

I did a note in the PR to explain I had made this change.

Previously this led to all of our javadoc empty, clearly now this change would just affect these components.

However I removed it and was able to compile with BOTH Maven at the command line, AND IntelliJ IDEA, without any issue with duplicate classes.

My understanding is that this depends on having the lobok plugin active.

Is there a particular sequence & environment where you do get duplicate sources reported so that we can pin this down in more detail without impacting the javadoc build?

alexandra-bucur:
Now I see what you said about javadoc. I'll put it in draft until I can get back to it. The issues appear in my local environment even though I do have the plugin installed, but I'll check again soon all details.

developer-resources/tools/IntelliJ.md
```diff
@@ -22,6 +22,8 @@ There is a [tutorial for IntelliJ](../../open-metadata-resources/open-metadata-t
 Lombok Plugin
 
 Egeria makes use of Project Lombok. If using JetBrains IntelliJ IDEA ensure the IDEA has the required plugin configured. See https://projectlombok.org/setup/intellij for more information.
+Also before running a Maven build please choose "Don't detect" from the "Generated sources folders" dropdown in Preferences -> Build, Execution, Deployment -> Build Tools -> Maven -> Importing. This will avoid triggering a duplicate classes build error caused by delomboked sources folder being added as source folder for the Maven module.
```
planetf1, line 25: Do you know what the impact to gradle is - our long term goal is still to switch

pom.xml
```diff
@@ -3587,6 +3605,7 @@
                 <configuration>
                     <additionalJOption>-J-Xmx1536m</additionalJOption>
                     <quiet>true</quiet>
+                    <sourcepath>${project.build.directory}/delombok</sourcepath>
```
planetf1, line 3608: I'm uncomfortable we change the source path for every module away from the defaults  - we have non java code too. In a number of cases in the past going with non-defaults have caused subtle issues elsewhere in the process

I see some references (google!) to keeping lombok'd code in a separate build tree within the component - this means that components without lombok are not affected *at all* (the delombok process will do nothing), whilst the lombok'd code is appropriate delomboked before the rest of the usual process applies. This seems to cause less changes to the overall build process /filesystem layout, and is really just adding in lombok in a less intrusive way

Even with a couple of components using lombok this seems a more natural fit, but it's particularly important if we want to use further in future

pom.xml
```diff
@@ -3342,6 +3342,24 @@
 
         <!-- Default compiler options - enable additional lint checks -->
         <plugins>
+            <plugin>
```
planetf1, line 3345: Since this is hard to review fully (for example myself, @mandy-chessell , @davidradl  have all seen slightly different behaviour),I think we should try out changes just on a limited set of modules until we are comfortable with how they work. We can then be more free to experiment until we get a good solution (which this may be)

planetf1 (changes requested):

pom.xml
davidradl: I agree, is there a way of organising this so that java builds without Lombok are not effected. Maybe the ' separate build tree option'

developer-resources/tools/IntelliJ.md
davidradl: I think this will be something that many people would fall over. If we can have it working with no settings changes - that would be ideal.

pom.xml
alexandra-bucur: The sourcepath tag above is inside the config of the javadoc plugin and it indicates the place where the plugin should search for java files in order to generate javadoc. sourceDirectory inside the build tag is left untouched so the java build will not be affected, nor the non java code files. 

Also, taken from Delombok documentation: "To keep any changes to your code style to a minimum, delombok just copies a source file directly without changing any of it if the source file contains no lombok transformations." This means that only the modules that really use Lombok will be touched by the plugin. 
That can also be seen in the console when running a mvn clean install command: [INFO] --- lombok-maven-plugin:1.18.20.0:delombok (default) @ open-metadata-deployment ---
[WARNING] Skipping Delombok; no source to process.

In this case I think there is no overload of the build or risks regarding sources files or javadocs.

pom.xml
alexandra-bucur: It would be great if you could compile a list of strange behaviours in order to be able to address them. I think that even without this PR the Maven "Don't detect" setting mentioned in the first comment of the PR would solve the "duplicate class" issue starting with a clean project structure.

developer-resources/tools/IntelliJ.md
alexandra-bucur: @planetf1 I tried really fast to import Egeria as a Gradle project in Intellij, but there seem to be some issues while building it after the import. Do you have a working gradle project import?

Otherwise, the gradle build runs successfully as can be seen here too. Anyway, I think the issue is in fact Maven + Intellij + Lombok related.

@davidradl I would like to have it working without a mandatory Intellij setting. Unfortunately, it looks like Intellij has a bug regarding adding the sources in the classpath when running with Lombok: https://youtrack.jetbrains.com/issue/IDEA-184689. The bug is in the "Submitted" state and this is a workaround. I do hope they will fix it soon so we can forget about the settings change.

pom.xml
planetf1: My concern is that if we go with the current config wide-scale across the project (an approach I did initially support!) we may hit various errors as we're moving away from the default location for source as assumed by maven/plugins. I know that isn't precise, but last time round we had problems that we couldn't fully describe. Maybe with time. Egeria has so many modules, and the overall build process is non trivial.

I also don't know what else the 'don't detect' may change. Usually with intellij we would have auto-import on, and all of intellijs settings around maven etc come from importing the pom (rather than making specific intellij config), so what else does that setting do? Does it affect the import of other modules

As such I think a better initial step is to keep the scope narrow/limited. This seems to me to be through
 * limiting the projects (ie plugin at module level) we use lombok on until a little time/changes have passed (in master) so we are comfortable with them (we also have multiple non-java modules). That gives us more confidence before rolling out more widely
 * Trying to stick with defaults for 'regular' java code, and adding any moves/special steps only for lombok -- for example here I'm thinking about things like having lombok source files in a seperate directory (this seems common practice?) so that only they are processed via the plugin?

pom.xml
planetf1: Thanks for clarifying the location of the source tag - I had missed that! Agreed that's reduced impact

We should also double check that with the plugin in place, that the source jar we push to maven central (or snapshot repo) contains appropriate source for debugging that intellij can use, as well as javadoc. (ie testing of the dev/debug scenario outside core egeria, but using egeria artifacts)

I still think it makes sense to do this in two stages - propose an approach for the small number of modules using lombok, merge, ask people to try it out in their environment (ie make code changes , build, debug), give it a period of time (weeks?) and then a new discussion/proposal on applying to all components (ideally only java ones)

developer-resources/tools/IntelliJ.md
planetf1: The issue dates from 2018. Could you raise the issue with jetbrains (add to the issue perhaps) to get clarification of expected behaviour?

On gradle - I haven't had time to explore, but I have been able to import egeria as a gradle project directly and without doing further config. However I've kept it in a seperate dir tree as otherwise intellij seems to get confused about whether the project is maven or gradle. But note that in that dir tree do a mvn clean install first (CLI). Why? Because only maven downloads some derby/gaian stuff that the build needs. that is addressed in a PR about to be merged.

planetf1:
Doing a little more reading ...
 * https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000763864-Can-t-permanently-exclude-delombok-from-generated-source seems to suggest that as long as delombok is outside generated sources, you should be ok (which is the case with your proposal), implying the config change isn't needed - which isn't what we appear to see
 * https://stackoverflow.com/questions/37801851/how-to-get-rid-of-duplicate-class-errors-in-intellij-for-a-mavenized-project-usi reports the above doesn't work, and suggests using clean, or the workaround (same as you found)
 * https://github.com/awhitford/lombok.maven/issues/7 (older article) talk through a few options - nothing particularly new here
 * In terms of where to put the source - https://github.com/awhitford/lombok.maven/issues/7 talks about moving lombok'd code to src/main/lombok or similar but indicates it causes issues in any case
 
 In terms of maven I don't think we can do any better than you have proposed. I agree the scope is limited in terms of delombok is ONLY being used here to support javadoc .
 
 I will try and check
  * source jars - I suspect the source jars are unaffected, ie they contain lombok code. I'm not sure what intellij will make of these when trying to build a project that is consuming one of our artifacts.
   * built javadoc - try javadoc.io with one of our lomboked & non-lomboked modules
   * gradle - check in intellij

developer-resources/tools/IntelliJ.md
planetf1: Worked with @alexandra-bucur on more checks today.
 - If we're going to use lombok and maven I don't think we can escape from this issue - which is really intellij being a little over zealous in looking for generated code . Having researched further I don't think we can do any better, and @alexandra-bucur 's proposal is the best we have.
 - We're not seeing this issue in gradle which is building cleanly at CLI and in INtelliJ. Javadoc & sources are correctly generated including aggregated javadoc . No configuration is needed (this is good news for when we switch to gradle). This is probably due to the better gradle/intellij integration.

pom.xml
planetf1: having researched/tested/discussed further I now support the proposal in the PR including going back to a change across all projects. If any further issues arise an Issue should be raised. Other ideas all seem to have their own set of issues and are not as good.

planetf1 (approved):

planetf1:
further checks/discussions complete. Supporting the proposal as the best option to enable lombok.

pom.xml
planetf1: now agree we don't need to back off in this way.

Issue #5226, opened 2021-05-26
Correct the readme for the community calls


davidradl:
https://github.com/odpi/egeria/blob/master/README.md says - 

Developers call - one hour - Discussion on code development - not minuted -

10 am (UK); 11 am (Amsterdam/CET); 12pm (Bucharest)

Zoom Conference https://zoom.us/j/523629111

Community call - one hour - Demos, meet-ups and other activities going on in the Egeria community

8am (EST); 1pm (UK/GMT); 2pm (Amsterdam/CET); 3pm (Bucharest)

Zoom Conference https://zoom.us/j/523629111

I notice that the developers call does not mention time zone, but the community call does. The UK call is incorrect during the summer as the calls are not being run at 1PM GMT but at 1PM BST. 

The definitive source is "https://lists.lfaidata.foundation/g/egeria-technical-discuss/calendar" Which will be undated with any reschedules or cancellations.

PR #5227, opened 2021-05-26
#5226 correct call details in readme

davidradl:

Updated because:
-  the time is wrong for the community call for UK. 
- the developers call incorrectly references the zoom call 
- for the community calls, there is no pointer to the definitive call schedule which is the Egeria LF&AI calendar


README.md
```diff
@@ -43,27 +43,29 @@ The governance of the project is described in more detail in the
 
 The Egeria team run three types of calls and all are welcome
 
-* **Developers call** - one hour - 
-  Discussion on code development - not minuted -
+* **Developers call** 
+   * one hour 
+   * Tuesday mornings (UK time)
```
lpalashevski, line 48: Developers community call is on Thursdays right? and on Tuesdays it is focus group call.

planetf1: Yes our public meeting is Thursday am.

davidradl: My bad -corrected

README.md
```diff
@@ -27,52 +27,53 @@ delivered in each release and our future plans.
...
+Calls are scheduled for one hour on Zoom Conference [https://zoom.us/j/523629111](https://zoom.us/j/523629111).
 
+[Egeria Calendar](https://lists.lfaidata.foundation/g/egeria-technical-discuss/calendar) holds the latest timings and information for these calls.
+
+* **Community call** 
+   * Every Wednesday.
+   * Demos, meet-ups and other activities going on in the Egeria community.
+   * *Egeria Community Call (every Wednesday)* in the
+   [Egeria Calendar](https://lists.lfaidata.foundation/g/egeria-technical-discuss/calendar).    
+   * Zoom Conference [https://zoom.us/j/523629111](https://zoom.us/j/523629111).
```
lpalashevski, line 53: Do we also want to have the zoom link per event? If so lest add it also for the other two calls (or just remove it here and have it on top).

README.md
```diff
@@ -27,52 +27,53 @@ delivered in each release and our future plans.
...
-bound to any corporate affiliation.
+This project aims to operate in a transparent, accessible way for the benefit of the Egeria community.
+All participation in this project is open and not bound to any corporate affiliation.
 
-To understand how to join and contribute see the 
+To understand how to join and contribute, see the 
 [Community Guide](./Community-Guide.md).
 
-All participants are bound the ODPi's [Code of Conduct](https://github.com/odpi/specs/wiki/ODPi-Code-of-Conduct).
+All participants are bound to the ODPi's [Code of Conduct](https://github.com/odpi/specs/wiki/ODPi-Code-of-Conduct).
```
planetf1, line 36: Oops. I guess we haven't updated to reference linux foundation yet?

README.md
planetf1: I think this is done with the latest update

planetf1 (approved):
Minor comment about LFAI, but agree with the important updates on call information

planetf1:
Note: The code scan is jumping in with many issues probably caused by a large merge last night. I can't see why they are attributed to this PR, and I note the check is optional. I suggest we ignore them for now for this particular proposal

README.md
davidradl: I am not sure what we should point to., Will raise an issue.

PR #5231, opened 2021-05-27
de-omas, al-omas and ols : Expanded to accept LineageMapping between Referenceable entities

lcpopa:


open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/AssetContextHandler.java
```diff
@@ -164,7 +164,7 @@ public RelationshipsContext buildAssetContext(String userId, EntityDetail entity
      * @throws OCFCheckedExceptionBase checked exception for reporting errors found when using OCF connectors
      */
     public Map<String, RelationshipsContext> buildColumnContext(String userId, String guid) throws OCFCheckedExceptionBase {
```
popa-raluca, line 166: This method builds the context for a TabularColumn/RelationalColumn, and doesn't need to be modified. The lineage mappings relationships are published in publishLineageMappingRelationshipEvent anyway(by calling publishLineageRelationshipEvent(lineageRelationship, eventType)) - so there is no need for this change

popa-raluca (changes requested):
The sample postman collection DE endpoints should be modified to reflect the new mapping added on the CopyColumnsFlow. Also, the lineage mappings .md file needs to be updated to reflect the new changes

popa-raluca (approved):

popa-raluca (approved):

PR #5295, opened 2021-06-02
Asset Catalog OMAS - verify that the entity is known before retrieving the entity details

popa-raluca:


open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/AssetCatalogHandler.java
```diff
@@ -121,6 +121,10 @@ public AssetDescription getEntityDetails(String userId, String assetGUID, String
         invalidParameterHandler.validateUserId(userId, methodName);
         invalidParameterHandler.validateGUID(assetGUID, GUID_PARAMETER, methodName);
 
+        if (repositoryHandler.isEntityKnown(userId, assetGUID, assetTypeName, methodName, GUID_PARAMETER) == null) {
+            return null;
+        }
+
         EntityDetail entityByGUID = commonHandler.getEntityByGUID(userId, assetGUID, assetTypeName);
```
mandy-chessell, line 128: This is very inefficient since the entity is being retrieved twice.  What is the purpose of the change?

PR #5309, opened 2021-06-02
June 2021 dependabot updates (part 2)

planetf1:
Additional dependabot wrapup for June 2021


planetf1:
Unfortunately jena cannot be updated as version 4 requires java 11.
(to be returned to)

PR #5312, opened 2021-06-02
Support three topics for a cohort

mandy-chessell:
This PR includes:
 * The code to allow OMAG Servers to form a cohort that uses 3 dedicated topics - one for each type of cohort event.
 * Updates to the admin guide to explain how to configure the cohort for different topic structures
 * Change of default userId for CreageGovernanceZonesSample and inclusion of this sample into the Governance Program OMAS FVT
 * New types for area 4: Threat and GovernanceDefinitionScope.


open-metadata-test/open-metadata-fvt/access-services-fvt/governance-program-fvt/pom.xml
```diff
@@ -73,6 +73,12 @@
             <artifactId>governance-leadership-sample</artifactId>
         </dependency>
 
+        <dependency>
```
planetf1, line 76: harmless, but we *should* be ok here as long as the assembly (which I think is the aim of what we test) is dependent on the sample.

Where is the actual code that executes the tests (I may have missed in lots of files)? Ideally I think that would be invoked via gradle or in pom via surefire plugin, similar to how we launch the other fvts?

planetf1: Ah. I see the test is in https://github.com/mandy-chessell/egeria/commit/c6dfbf297b838a33ff3358f2074cf7415e8bc6eb

PR #5321, opened 2021-06-04
fix some Sonar blocker bugs

wbittles:
fixes the sonar blocker bugs in the CSVFileStoreConnector and the KafakaOpenMetaDataTopicConnector


planetf1:
It's great to be correcting some of these issues.

Out of interest what did you use as source?
Our sonarcloud scan at https://sonarcloud.io/organizations/odpi/projects is not up to date (though still has useful info)

wbittles:
@planetf1 it was actually the link you added to the postgres connector readme that alerted me to these.
I thought I'd try the low hanging fruit first , before cracking on and fixing a few more.
The bugs I fixed are valid resource leaks as is , so I don't think moving java versions will have an impact.

planetf1:
@wbittles ah got it - yes a cut & paste. That was indeed a link to sonarcloud. 
The build is now fixed, so you should see an updated scan at https://sonarcloud.io/summary/new_code?id=odpi_egeria within 20-30 minutes

wbittles:
@planetf1 do we know why those two scans are so dramatically different , especially as the underlying issues are still present ?

planetf1:
> @planetf1 do we know why those two scans are so dramatically different , especially as the underlying issues are still present ?

Different technologies/companies, different rulesets is probably a big part of it -- but 'probably' in that we've not done a thorough comparison. There are also other scans we've seen before - both from commercial tools and more easily 'snyk' (for public repos anyone can run this). 

The bigger issue is for developers spending time on fixing, and balancing that with other work. This is where I think the github scan works best (for newly introduced issues)- it inserts a review comment directly into a PR so is more easily seen and addressed. The other scans require more effort to go review?

INterestingly the github scan is based on LGTM - a company github acquired, and it also offers it's own UI which may be better/easier for a full scan.

I'd be very happy to arrange/join a session for anyone interested to talk through what we've tried in the past and see if we can come up with an approach to do better? We can also use one of our regular calls.

wbittles:
@planetf1 
So I'm just comparing  the results of a Sonar java version 8 scan , and a Sonar version 11 scan.
In the version 8 scan the defect I fixed were listed , and in the Sonar version 11 scan they are missing ?

The biggest concern is that they are valid defects and the defect I looked at where all potential resource leaks and need to be fixed.
I'll have a look at the integrated report.

planetf1:
@wbittles agree that is worth understanding! For sonarcloud we use a free cloud service, and the rules can change from time to time (which we don't control). A possibility, though generally you'd expect detection to improve of valid issues. I think it's worthy of a discussion for sure. I'll can try and make sure all the scans are up to date and viewable - perhaps snyk, lgtm, github reports, sonar & we can chat over next week.

wbittles:
@planetf1 yes let's do that . I checked the for any codeql alerts on this change and it looked totally clean.

planetf1 (approved):

Issue #5305, opened 2021-06-02
Swap the isatypeof type ends around 


davidradl:
The isatypeof relationship type has the ends the wrong way round. The direction of relationship should point from end1 to end2. So in this inheritance case the UML arrow should point from the child to parent.

Currently TermISATypeOFRelationship is defined with end1 as supertypes and end2 as subtypes.

It should be the other way round. 

we are also assessing whether we need ISA as well as TermISATypeOFRelationship

davidradl:
@cmgrote If we were to change this would this cause an issue for any code you are aware of? 
@mandy-chessell as discussed

Any concerns?
@planetf1 
@CDaRip2U 
@lpalashevski 

If not we may need to deprecate and introduce a newly named relationship with the ends thew other way round

Issue https://github.com/odpi/egeria/issues/5304 sets the spine object classifications appropriately when the relationship is created. We are holding off from implementing this for isatypeof until this issue is resolved one way or another

cmgrote:
Are you talking about changing the UML diagram or the relationship itself?

If the latter, is there a way to do this via a TypeDefPatch?

If not, wouldn't this be a breaking change to the types (not backwards compatible, so any existing repository with such relationships will be broken), and therefore is not a permitted change to the type?

davidradl:
@cmgrote thanks for your feedback and questions.

 yes the relationship itself
yes this would be via a TypeDefPatch
yes it would be a breaking change - if anyone is using it / be impacted by it. Pragmatically if It is not being used then we could update it in a TypeDefPatch . I was hoping to assess the impact by tagging everyone.   

If we see there is an impact then this would be a breaking change, so we would deprecate and create a newly named type - maybe isatype.

cmgrote:
I am surprised we are considering a breaking change to the types -- it seems to me that there are countless previous examples where we wanted to make a breaking change to a type to "fix" one oversight or another, but did not explicitly because we cannot be sure we have visibility over all users that might already be using the type.  So I don't see how this situation is different (?)

Changing code in connectors or proxies is easy enough, but this doesn't address the migration of existing content in a repository.  We'd need upgrade utilities in all the persistent repos (Graph, Crux, etc) to apply such breaking changes, and would presumably need a major release to be able to switch all OMASes that interact with the type to only use the newer version of it (not have to handle previous versions), or what would happen in a cohort that had different repositories some of which have the old relationship going in the opposite direction and others in the new direction (?)

davidradl:
[...]

Issue #4903, opened 2021-03-12
Gradle: Subject Area FVT fails


planetf1:
In #4098 the gradle build is being extended to run FVTs. Other FVTs in the test suite are running, but I see an odd area with Subject Area FVT that requires further investigation.

I suspect it is a classpath issue - need to review code and debug, so opening up this issue to investigate later & to allow the other tests to be checked in (currently gradle does not run FVTs)

**Once fixed this test should be reenabled in gradle**

The failure is seen when using the groovy script to initialize subject area OMAS:
```
=== Checking platform at https://localhost:10448 is available (12 attempts remaining) ===
200
Egeria OMAG Server Platform (version 2.8-SNAPSHOT)

=== Configuring server: serverinmem ===
200
{"class":"VoidResponse","relatedHTTPCode":200}
=== Enabling Subject Area OMAS: serverinmem ===
200
{"class":"VoidResponse","relatedHTTPCode":200}
=== Starting server: serverinmem ===
200
{"class":"SuccessMessageResponse","relatedHTTPCode":500,"exceptionClassName":"org.odpi.openmetadata.adminservices.ffdc.exception.OMAGConfigurationErrorException","exceptionCausedBy":"java.lang.IllegalStateException","actionDescription":"initialize","exceptionErrorMessage":"OMAG-ADMIN-500-004 The Subject Area OMAS service detected an unexpected java.lang.IllegalStateException exception with message zip file closed during initialization","exceptionErrorMessageId":"OMAG-ADMIN-500-004","exceptionErrorMessageParameters":["Subject Area OMAS","java.lang.IllegalStateException","zip file closed"],"exceptionSystemAction":"The system is unable to start the service in the OMAG server.","exceptionUserAction":"This is likely to be either an operational or logic error. Look for other errors in the audit log.  Validate the request.  If you are stuck, raise an issue."}
```
```
Fri Mar 12 09:36:50 GMT 2021 servergraph Startup OMAS-SUBJECT_AREA-0001 The Subject Area Open Metadata Access Service (OMAS) is initializing a new server instance
Fri Mar 12 09:36:50 GMT 2021 servergraph Exception OMAS-SUBJECT_AREA-0005 The Subject Area Open Metadata Access Service (OMAS) is unable to initialize a new instance in server java.lang.IllegalStateException; the zip file closed exception occurred with error message: {2}
[...]

PR #5351, opened 2021-06-10
#5305 Replace is-a-type-of relationship

davidradl:


davidradl:
I have this as a draft PR, until the relationship and end names have been agreed.

planetf1:
I note the maven build failed
Note that with gradle currently the subject area tests are not running - see https://github.com/odpi/egeria/issues/4903

Issue #5350, opened 2021-06-10
missing PathVariable in the path


rychagova:
I'm not sure, i'm newbee, it looks like there's no {setGUID} in the path to create a new valid value definition.
`@RestController
@RequestMapping("/servers/{serverName}/open-metadata/access-services/asset-owner/users/{userId}")

public class ValidValuesOnboardingResource{

@PostMapping(path = "/valid-values/new-definition")

    public GUIDResponse  createValidValueDefinition(@PathVariable String                 serverName,
                                                    @PathVariable String                 userId,
                                                    @PathVariable String                 setGUID,
                                                    @RequestBody  ValidValuesRequestBody requestBody)`

mandy-chessell:
Agreed - looks wrong to me

mandy-chessell:
Would you like to fix it, or should I?

rychagova:
I would!

planetf1:
Closing as this is now merged. Thanks @rychagova for spotting and fixing

PR #5357, opened 2021-06-11
per Issue #5350

rychagova:
https://github.com/odpi/egeria/issues/5350


planetf1 (changes requested):
Hi, I notice your PR failed on our DCO check (see https://egeria.odpi.org/developer-resources/why-the-dco.html)

There are tips at https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md to correct this retrospectively.

Hope that helps.

I'll enable actions for you too for the build validation, but the DCO change is needed before any merge can happen.

Thanks!

Issue #5336, opened 2021-06-08
Change LocalDateTime to Date or Long for Asset Lineage OMAS publishEtities 


lpalashevski:
Based on recent discussions, we established standard for working with dates-times in Egeria java code.

https://github.com/odpi/egeria/blob/master/developer-resources/Developer-Guidelines.md#working-with-date-and-time

Following this, we need to adjust asset-lineage omas implementation (api/server/client) to use Date (or Long).

lpalashevski:
Fixed.

PR #5381, opened 2021-06-22
[al-omas] Changing LocalDateTime to Date

lcpopa:

### DIFF
--- a/open-metadata-implementation/access-services/data-engine/data-engine-api/build.gradle
+++ b/open-metadata-implementation/access-services/data-engine/data-engine-api/build.gradle
@@ -13,7 +13,6 @@ dependencies {
     implementation project(':open-metadata-implementation:common-services:ffdc-services')
     implementation project(':open-metadata-implementation:frameworks:audit-log-framework')
     implementation project(':open-metadata-implementation:repository-services:repository-services-apis')
-    implementation 'org.apache.commons:commons-collections4'
 }
 
 description = 'Data Engine OMAS API'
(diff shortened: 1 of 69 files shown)


lcpopa:
#5336

open-metadata-implementation/access-services/asset-lineage/asset-lineage-spring/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/server/spring/AssetLineageResource.java
```diff
@@ -54,7 +54,7 @@
     public GUIDListResponse publishEntities(@PathVariable String serverName,
                                             @PathVariable String userId,
                                             @PathVariable String entityType,
-                                            @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE_TIME) LocalDateTime updatedAfterDate,
+                                            @RequestParam(required = false) @DateTimeFormat(iso = DateTimeFormat.ISO.DATE) Date updatedAfterDate,
```
lpalashevski, line 57: I think we should preserve the date time format, otherwise it looks to me that if we switch the format to DATE we are going to cut-off the time information. The assumption is based on the format I see in the docs https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/format/annotation/DateTimeFormat.ISO.html#DATE

Please check this.

lpalashevski (changes requested):
Potential issue with the Date format used in rest resource parameter. See inline comment.

lpalashevski (approved):

planetf1 (approved):

Issue #4046, opened 2020-10-31
Dockerhub pull request limits - additional registry/docs


planetf1:
See https://docs.docker.com/docker-hub/download-rate-limit/ & some useful analysis at https://www.openshift.com/blog/mitigate-impact-of-docker-hub-pull-request-limits

Need to consider
 - effect on build process
 - effect on consumers
 - whether we need to save base images (as those images will have restrictions)
 - use of other registries (github, azure, quay.io) and/or premium subscriptions

tbc....

planetf1:
Dockerhub is now offering exemption for qualifying open source projects -> see https://www.docker.com/blog/expanded-support-for-open-source-software-projects/
Will see if we can get this in place cc: @jmertic

planetf1:
During an attempt to test the 2.10 release #5211 , the following was noted when trying to setup egeria:
```
 Warning  Failed          19s               kubelet            Failed to pull image "docker.io/odpi/egeria:2.10": rpc error: code = Unknown desc = Error reading manifest 2.10 in docker.io/odpi/egeria: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
 ```
 
Following https://www.docker.com/blog/checking-your-current-docker-pull-rate-limits-and-status/ we see:
```
$ curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/ratelimitpreview/test/manifests/2.10 2>&1
```
and decoding at jwt.io :
```
{
  "access": [
    {
      "type": "repository",
      "name": "odpi/egeria",
      "actions": [
        "pull"
      ],
      "parameters": {
        "pull_limit": "100",
        "pull_limit_interval": "21600"
      }
    }
  ],
  "aud": "registry.docker.io",
  "exp": 1622154518,
  "iat": 1622154218,
  "iss": "auth.docker.io",
  "jti": "fTZgVmovxlHuKuoo9K1Y",
  "nbf": 1622153918,
  "sub": ""
}
```
```
curl --head -H "Authorization: Bearer $TOKEN" https://registry-1.docker.io/v2/odpi/egeria/manifests/2.10 2>&1
HTTP/1.1 200 OK
content-length: 19568
content-type: application/vnd.docker.distribution.manifest.v1+prettyjws
docker-content-digest: sha256:e3c61ab251ec404b05d0dc92b136a7823b245690989e1b20badf30c7c6d0d792
[...]

Issue #3580, opened 2020-09-07
Switch to UBI base images for egeria docker image(s)


planetf1:
See #3560

Recently we discovered the jre image used for egeria was not being regularly updated by the openjdk team. This could expose anyone running the code to more bugs/security issues - and most recently to incompatabilities.

Work is underway to build an egeria operator, and redhat produce freely usable 'UBI' standard images including a few standard frameworks. As of July 2020 java is included & supported. This makes for a more standard base OS/jre for anyone putting the egeria operator into production

The base image will become RHEL rather than alpine which means the necessary changes ie apk->yum but the amount of work should be small.

Aiming to do this alongside the operator work in the 2.3 timescale

planetf1:
More info on the redhat UBI image : https://developers.redhat.com/blog/2020/06/25/introducing-the-red-hat-build-of-the-openjdk-universal-base-images-now-in-red-hat-enterprise-linux-8-2/

planetf1:
The image is licensed under:   https://www.redhat.com/en/about/red-hat-end-user-license-agreements#UBI - specifically the current text is at https://www.redhat.com/licenses/EULA_Red_Hat_Universal_Base_Image_English_20190422.pdf . 

The image help is accessed via ` docker run registry.access.redhat.com/ubi8/openjdk-11 cat /help.md`

The relevant environment variables that would help us to run our apps include:
```
* JAVA_APP_DIR - location of the application
* JAVA_ARGS - arguments for the application
* JAVA_CLASSPATH (or can be set via the file JAVA_APP_DIR/classpath)

Optionally
* JAVA_LIB_DIR - for extra libraries - ie server/lib for the main app
* JAVA_MAIN_CLASS - could be useful if running something different

Some functionality we scripted - but no longer needed
* JAVA_DEBUG (true/false) - to enable debugging - port is 5005 by default but can be overriden - unnecessary 
For persistant storage we also need to consider
* JAVA_DATA_DIR - location of data files when we add persistance (a volume could be placed over here)

The intent should be to run egeria in a similar way to today -- run egeria chassis itself by default, UI if required (or other tools)

The image runs as user 185 (different to our choice today - but I see no reason to change from default)
Ports 8080, 8443 are open by default - we'll stick with 8443 (UI) and 9443 (chassis)
The launch script is /usr/local/s2i/run by default.

**Comment by planetf1** [comment]
Simple changes made to work with UBI.

Results is we can start up the egeria chassis as before - ie no parms
```
[...]

PR #5393, opened 2021-06-25
Fixes static UI portion of lab chart

cmgrote:
Primarily fixing the UI chassis pod, by overriding the CMD with the appropriate jar file and truststore locations

(@planetf1 not sure if it makes sense to merge this into master here, for the imminent release; or to wait to move it to the separate charts repository?  Change is reasonably small, and does fix something (the static UI) that I don't think works at all in the current chart (?))


planetf1:
Interesting as it worked on last release. As in asset search. Will review. Here should focus for current release. We'll have to port over to new repo later if needed

planetf1:
What did you find failed? I will retest in a while

planetf1:
Are you going to https://xxxx:443 ? Where xxxx is the address of the Nginx proxy as exposed?

planetf1 (changes requested):
I've restested the lab chart (openshift) 
 - deploy the chart (all defaults)
 - wait until pods all ready
 - go to jupyter (with an 'exposed' service via alb) & run config/start/data catalog notebooks
 - go to https://<addressofnginxservice>
 
 I was able to login as faithbroker/admin, and was able to navigate to the asset search screen
 I don't see any backend errors
 
 I wasn't able to get search results. However I don't have good test case information/guidance/notebooks for this part.
 
 Did your problem occur prior to this point? 
 Is your environment different? Perhaps there's some other reason for failure?
 Are you using current docker images?
 Is this in master?

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/etc/default.conf.template
```diff
@@ -3,31 +3,24 @@
...
     }
 
+    # Set the static UI as the default location to load
     location / {
-        proxy_pass ${UI_STATIC};
-        proxy_set_header Host $http_host;
-        proxy_ssl_verify       off;
-        proxy_ssl_session_reuse on;
-        proxy_ssl_server_name on;
+        proxy_pass              ${UI_STATIC};
```
planetf1, line 22: see overall review comment. this seems to work on openshift (insofar as I test). We need to understand what the difference is if it doesn't work for you?

planetf1 (changes requested):

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-ui.yaml
```diff
@@ -72,6 +72,10 @@ spec:
           envFrom:
             - configMapRef:
                 name: {{ include "myapp.fullname" . }}-configmap
+          command:
+            - "/bin/bash"
+            - "-c"
+            - "java -jar /deployments/user-interface/ui-chassis-spring-*.jar --server.ssl.trust-store=/deployments/truststore.p12"
```
planetf1, line 78: the UBI8 openjdk image is configured differently.... We set JAVA_APP_JAR to the application that is to be run. We do not set a command to be executed. This follows the standard pattern for that image

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/values.yaml
```diff
@@ -44,7 +44,7 @@ imageDefaults:
   registry: docker.io
   namespace: odpi
   tag: latest
-  pullPolicy: Always
+  pullPolicy: IfNotPresent
```
planetf1, line 47: For *master* I set this value to Always intentionally, since the master code is always changing, and otherwise the image gets cached.
If that's inappropriate you can override.
It gets set to 'IfNotPresent' when the version number is locked prior to release

planetf1 (changes requested):

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/values.yaml
planetf1: See https://github.com/odpi/egeria/issues/4046 also for the reference to dockerhub limits. You can authenticate to dockerhub to increase the limits, or wait a while. I think this is clearer than accidentally using invalid code.

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/templates/egeria-ui.yaml
planetf1: There's a little more on the UBI image captured in the issue originally used to create it https://github.com/odpi/egeria/issues/3580

planetf1: The way the environment is setup also means the ssl parm shouldn't be needed - at least to use default certs (we may still have work to do to inject custom certs into the deployment)

open-metadata-resources/open-metadata-deployment/charts/odpi-egeria-lab/etc/default.conf.template
```diff
@@ -3,31 +3,24 @@
 server {
 
     listen                443 ssl;
-    #listen                 80;
```
planetf1, line 6: We only intend to listen on port 443

planetf1:
I've checked the ui notebook (which has little in it) and it does still say https://xxxx:8443 - which is wrong

The 2.8 release notes contained a statement
```
 In this release any routing needs to be setup externally, for example by placing a nginx proxy in front of both the ui
 chassis and static content server. This is now done by our docker-compose environment & helm charts so to access
 the UI you need to go to the nginx proxy. Further summary information can be found in the documentation for those assets.
 ```
 Unfortunately the chart README.md does not clarify this (in part as there's little about the UI in total).
 
 So I do think some update docs at a minimum would be useful? But I think the code is working as I intended.... in so far as I know ;-)

planetf1 (changes requested):

cmgrote:
I was under the impression the static UI wasn't accessible from other conversations, so started getting this working outside Helm initially (missed the UBI specifics there just noticed it wasn't starting the UI process).  But seems I fundamentally misunderstood that the static UI is not accessible -- sounds like it has been all along?  So no use for this PR, closing.

planetf1:
Happy to chat through this next week. At least perhaps we can improve the docs/explanation

Issue #5208, opened 2021-05-20
Migrate to Java 11


planetf1:
Egeria currently required Java 8, and enforces Java 8 language constructs, but is built for both Java 8 and 11, our containers run with Java 11, and some developers are using further versions up to 16.

Some minor dependencies we are using are no longer supporting Java 8, and other software stacks are tending to move to Java 11 which is the current Java LTS and has free support until 2024-2027 depending on distribution 

Note that Java 17 is the next LTS from Sep 2021, but feel it's more appropriate to go with v11 for the next year or so at a minimum

See https://en.wikipedia.org/wiki/Java_version_history for information on service dates

I therefore propose we
 - Adopt java 11 as minimum required language level (for build and run)
 - Build and run primarily on Java 11 but also run a build (with any workarounds needed/or optionally) on Java 16 initially, with a goal of covering Java 17 in September to proactively address any breakages before next LTS
 
 This could co-incide with a switch in major version of Egeria to version 3
 
 cc: @mandy-chessell @lpalashevski @cmgrote 
 
 (see #5091 for an outstanding issue with Java 16, but this does not affect 11)

planetf1:
#2377 previously proposed making use of module-info files which adds better support for a contract between modules providing capabilities & those consuming. See https://www.baeldung.com/java-9-modularity

This would aid in preventing unintended use & improving clarity in our APIs. It may be a little disruptive. It can be done after going to java 11, although worth considering if a new release 3 is the right time to try and do this. It would however require a fair bit of work in adding correct module info files for producer/consumer

planetf1:
No objections. Discussed on developer call 2021-06-22 & agreed that we will plan to make this change for the release after the one we are about to do (2.11) . This will occur at the beginning of that release, so in the next week or so, and we will update the version to 3.0 at that time

planetf1:
Note that our egeria containers already use Java 11 runtime (redhat openjdk, based on UBI8), as does the gradle build. these will be left as is

Work plan:
 - [ ] Ensure 2.11 release notes indicate last version for Java 8
 - [ ] Slack post reminder before starting migration
 - [ ] Update main build scripts for Java 8 to use Java 11 (this covers release, publishing maven artifacts) - use adoptopenjdk (*vs zulu8 currently). 
 - [ ] leave gradle build using Java 11, but make distro consistent (adopt)
 - [ ] Update language level settings in maven/gradle to use language level 11
[...]

PR #5403, opened 2021-06-30
#5208 update docs on java/gradle to be consistent with Java 11 switch

planetf1:

Fixes: #5208 

Updates docs following Java 11 switch


planetf1:
Any further comments - suggest opening an issue/pr - for now at least the basic info for new users of master will be in place

Issue #5380, opened 2021-06-22
Release 2.11


planetf1:
Create 2.10 release 

- [x] slack post
- [x] agree in team call 
- [x] Create branch
- [x] Reassign any issues not being worked on to the next release
- [x] Update master
- [x] Ensure any remaining fixes are merged 
- [x] Update branch version 
- [x] Remove release notes from branch 
- [x] Update image versions for compose & charts
- [x] update release notes & rename 
- [x] Verify notebooks 
- [x] Check polymer UI 
- [x] Check node (ps) UI 
- [x] Run CTS
- [x] Check samples 
- [x] Check swagger doc 
- [x] Verify odpi-egeria-lab chart
- [x] Verify egeria-base chart
- [x] verify docker-compose 
- [x] create release candidate (manual github action)
- [ ] Verify RC packaging on MC
- [x] Fix up maven artifacts as needed
- [ ] Create final github releases (launches pipeline)
- [ ] Verify artifacts all on maven central 
- [ ] Release on maven central
- [x] ensure any fixes ported to master

Discussed in team call 2021-06-22. Awaiting some delete related changes from @davidradl and @lpalashevski .current default is to branch on 2021-06-29,  but will review again on 2021-06-24

Note that this will be the last 2.x release, the new release will be 3.0 and also plan to include Java 11 support

planetf1:
Reviewed today, agreed to wait for latest changes as per @lpalashevski with target of branch Tue 2021-06-29

planetf1:
Agreed developer call 2021-06-29 to go ahead once last PRs merged (now complete)

planetf1:
CTS (inmem) working well with
```
      "testCaseCount": 5245,
        "testFailedCount": 0,
        "testPassCount": 5245,
```
```
	name	description	profilePriority	conformanceStatus
0	Metadata sharing	The technology under test is able to share met...	MANDATORY_PROFILE	CONFORMANT_FULL_SUPPORT
1	Reference copies	The technology under test is able to save, loc...	OPTIONAL_PROFILE	CONFORMANT_FULL_SUPPORT
2	Metadata maintenance	The technology under test supports requests to...	OPTIONAL_PROFILE	CONFORMANT_FULL_SUPPORT
3	Dynamic types	The technology under test supports changes to ...	OPTIONAL_PROFILE	UNKNOWN_STATUS
[...]

PR #5406, opened 2021-06-30
#5380 Add release notes for 2.11

planetf1:

See #5380

- Adds release notes for 2.11 into the release notes README 
- adds reference to planned java switch

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-client/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/client/nodes/categories/SubjectAreaCategoryClient.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-client/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/client/nodes/categories/SubjectAreaCategoryClient.java
@@ -39,7 +39,7 @@ public class SubjectAreaCategoryClient<C extends Category> extends AbstractSubje
      */
 
     public List<Category> getCategoryChildren(String userId, String guid, FindRequest findRequest) throws InvalidParameterException, PropertyServerException, UserNotAuthorizedException {
-      return getCategoryChildren(userId,guid, findRequest, false, false);
+      return getCategoryChildren(userId,guid, findRequest, false, true);
     }
     /**
      * Get the Category's children categories
@@ -77,7 +77,7 @@ public class SubjectAreaCategoryClient<C extends Category> extends AbstractSubje
 
     public List<Term> getTerms(String userId, String guid, FindRequest findRequest) throws InvalidParameterException, PropertyServerException, UserNotAuthorizedException {
 
-        return getTerms(userId,guid,findRequest, true,false);
+        return getTerms(userId, guid, findRequest, false,true);
     }
     /**
      * Get the terms owned by this Category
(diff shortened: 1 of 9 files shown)


release-notes/release-notes-2-11.md
```diff
@@ -147,7 +147,10 @@ in the [Data Connectors](https://github.com/odpi/egeria-data-connectors) GitHub
 ## Known Issues
 
 * It is recommended to use a chromium-based browser such as Google Chrome or Microsoft Edge, or Apple Safari for the Egeria React UI. Some parts of the UI experience such as Dino currently experience problems with Firefox. See [odpi/egeria-react-ui#96](https://github.com/odpi/egeria-react-ui/issues/96) .
-* When running the 'Understanding Platform Services' lab, ensure you run the 'egeria-service-config' notebook first and do not restart the python kernel before running this lab. See [#4842](https://github.com/odpi/egeria/issues/4842) .
+
+## Special note on Java Support
+
+* This is expected to be the last release supporting Java 8. The next release will commence the 3.x series of releases and require Java 11.
```
davidradl, line 153: I wonder do we want to talk about 3 being a new version to highlight the change

release-notes/release-notes-2-11.md
```diff
@@ -147,7 +147,10 @@ in the [Data Connectors](https://github.com/odpi/egeria-data-connectors) GitHub
 ## Known Issues
 
 * It is recommended to use a chromium-based browser such as Google Chrome or Microsoft Edge, or Apple Safari for the Egeria React UI. Some parts of the UI experience such as Dino currently experience problems with Firefox. See [odpi/egeria-react-ui#96](https://github.com/odpi/egeria-react-ui/issues/96) .
-* When running the 'Understanding Platform Services' lab, ensure you run the 'egeria-service-config' notebook first and do not restart the python kernel before running this lab. See [#4842](https://github.com/odpi/egeria/issues/4842) .
+
+## Special note on Java Support
```
davidradl, line 151: I am not sure why but I do not seem to be able to add review comments on the 2.11 release notes md file  - I am wondering if there is something incorrect with it.

release-notes/release-notes-2-11.md
davidradl: I wonder if we need more detail that 'require Java 11'.
- I run with java 15 and that works. Maybe saying we now target compiles at java 11 bytecodes would be better
- maybe mention both at runtime and compile time.
-

planetf1: I do say it will commence the 3.x series of releases? I think that's sufficient?
On Java version, I've written more in the Java.md on release 3.0 which I hope is a little clearer (though we can refine that accordingly!) and where we have more space to elaborate. I wanted to add a simple reference here as a 'heads-up'. With that goal I think saying require Java 11 is sufficient EVEN THOUGH I agree technically - We require Java 11 as a min, and we build everything inc. maven artifacts/docker with Java 11. We also build with Java 16 which is kinda ok but we have one glitch with private symbols. In summary I'm inclued to just adopt the KISS principle in this statement

release-notes/release-notes-2-11.md
planetf1: this file is the 2.11 release notes md file?

release-notes/release-notes-2-11.md
planetf1: That PR for master is at https://github.com/odpi/egeria/pull/5403

Issue #5407, opened 2021-06-30
Restart KafkaOpenMetadataEventConsumer if thread dies


jnhagelberg:
We have been having problems where, for whatever reason the `KafkaOpenMetadataEventConsumer` thread dies.  When this happens, the sync processing completely stops.  In order for our product to continue functioning when this happens, we need the `KafkaOpenMetadataEventConsumer` thread to be restarted if it dies.

There is already similar logic for restarting the `KafkaOpenMetadataEventProducer` if that thread dies.  The mechanism needs to be extended to also apply to the consumer.

jnhagelberg:
FYI, I'm working on a PR for this.

mandy-chessell:
What are the reasons for the thread dying - should there be some additional error handling in place to prevent it dying?

jnhagelberg:
@wbittles What we saw was that that thread died during what appeared to be a transient out of memory issue.  We are looking at that, but to us this is really about resiliency.  If the thread dies, we are pretty much dead in the water, so we need to do everything we can to keep it alive.  In our scenario, the memory issue resolved itself but the kafka consumer thread remained dead.

jnhagelberg:
@mandy-chessell ^^

wbittles:
@jnhagelberg 

Resiliency shouldn't be valued over functional integrity, there is little value in restarting a thread if it's behaviour is indeterminate.
My understanding is that OOM is a Java JVM Error , where the JVM has already decided that it's in an error state and that the JVM needs to be shut down. 
My fear is that this JVM Error has been caught and translated into an Exception which is allowing Egeria to ignore the state of the JVM, and Egeria is in effect just limping along in an error state.
To resolve OOM errors usually involves allocating more memore via the JVM command line options and it's unclear how we are going to avoid the need to shut the JVM and down and restart with an increased heap ?

planetf1:
@wbittles - Completely agree... In fact in this case I would suggest that the JVM must be restarted - for example in a k8s environment this could be to intentionally let the server 'die' (shut itself down) and exit, and k8s will restart it elsewhere - clean environment, new jvm etc

jnhagelberg:
@planetf1 @wbittles I agree that we should probably restart the JVM.  However, to this is more about resiliency.  We cannot guarantee that the JVM will be restarted.  If it does not restart we need to make a best effort attempt to keep functioning. 
[...]

PR #5408, opened 2021-06-30
opdi/egeria#5407 - Restart KafaOpenMetadataEventConsumer thread if it does unexpectedly

jnhagelberg:
Signed-off-by: Jeff Hagelberg <jnhagelberg@us.ibm.com>

See odpi/egeria#5407

### DIFF
--- a/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/ffdc/OMAGAdminErrorCode.java
+++ b/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/ffdc/OMAGAdminErrorCode.java
@@ -375,6 +375,11 @@ public enum OMAGAdminErrorCode implements ExceptionMessageSet
                "The server fails to start because this value is needed to operate successfully.",
                "Add a value for this property to the configuration document and restart the server."),
 
+    AUDIT_LOG_DESTINATION_NOT_FOUND(404, "OMAG-ADMIN-404-100",
+                       "The {0} audit log destination connection name does not exist, so the requested {1} operation cannot proceed",
+                       "The audit log destination is not changed.",
+                       "Amend the request so it refers to an audit destination log connection name that exists."),
+
     UNEXPECTED_EXCEPTION(500, "OMAG-ADMIN-500-001",
             "Method {1} for OMAG server {0} returned an unexpected exception of {2} with message {3}",
             "The system is unable to configure the OMAG server.  No change was made to the server's configuration document.",
--- a/open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/AdminServicesRESTClient.java
+++ b/open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/AdminServicesRESTClient.java
@@ -11,6 +11,7 @@ import org.odpi.openmetadata.adminservices.ffdc.exception.OMAGNotAuthorizedExcep
 import org.odpi.openmetadata.adminservices.rest.*;
 import org.odpi.openmetadata.adminservices.rest.ConnectionResponse;
 import org.odpi.openmetadata.commonservices.ffdc.rest.*;
+import org.odpi.openmetadata.frameworks.connectors.ffdc.PropertyServerException;
 
 
 /**
@@ -126,7 +127,7 @@ class AdminServicesRESTClient
 
 
     /**
-     * Issue a POST REST call that returns a VoidResponse object.  This is typically a create
+     * Issue a DELETE REST call that returns a VoidResponse object.  This is typically a delete
      *
      * @param methodName  name of the method being called.
      * @param urlTemplate  template of the URL for the REST API call with place-holders for the parameters.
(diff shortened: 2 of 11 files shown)


planetf1:
@wbittles you may want to review?

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
```diff
@@ -58,15 +64,32 @@
...
+    
+    private class KafkaConsumerExecutor extends ThreadPoolExecutor {
+        KafkaConsumerExecutor() {
+            super(1, 1, Long.MAX_VALUE, TimeUnit.MICROSECONDS, new LinkedBlockingQueue<>(1));
+        }
+
+
         @Override
         public void afterExecute(Runnable r, Throwable t) {
             super.afterExecute(r, t);
```
mandy-chessell, line 87: What happens if the server is shutting down - should it check that the connector is still active before restarting the threads?

mandy-chessell (changes requested):
This may be ok but I can not see any logic that checks that the connector has is still active before restarting the threads.  Please confirm that this case is handled.

jnhagelberg:
@mandy-chessell This uses the exact same logic that is used to restart the producer thread.  If the server is shutting down, the thread will be killed by the JVM, so that should not be an issue.  I will add logic to check whether or not the connector is active before restarting the thread.

jnhagelberg:
@mandy-chessell I have pushed changes to address your review comments.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
jnhagelberg: Fixed

mandy-chessell:
> @mandy-chessell This uses the exact same logic that is used to restart the producer thread. If the server is shutting down, the thread will be killed by the JVM, so that should not be an issue. I will add logic to check whether or not the connector is active before restarting the thread.

The JVM does not exit when the server shuts down.  The JVM only exits when the platform shuts down.  If the producer thread is restarted without checking whether the connector is active or not then it is wrong too.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
mandy-chessell: Just need to update the comment so it matches the code and then I think it is ok

jnhagelberg: @mandy-chessell I think I addressed you comment.  I'm not exactly sure which comment you were referring to.

jnhagelberg:
@mandy-chessell Can you please take another look?

jnhagelberg:
@mandy-chessell Can you please take another look?

mandy-chessell:
@jnhagelberg Hello Jeff.  We are still debating what the fix should be.   One of the decisions we made last year was to no longer catch Throwable, and to catch Exception instead.  This would leave the Errors such as the out of memory error to be handled by the JVM and typically cause the process (ie the platform) to exit.  There is concern that if a JVM error occurs, the state of the memory and other aspects of the process is unknown which is why it is not safe to continue.

We are still in the transition and so some of the components are still catching Throwable - unfortunately the Kafka topic connector is one of them.  At one point it even wraps a Throwable with a CheckedException.

This connector clearly needs to be updated so it is no longer catching Throwable.  
Once updated then we need to consider what circumstances are OK to restart the thread and what follow on errors will result.

One scenario is that there is an event that is very big and this causes an error when it is received.  When the thread is restarted, does it:
 * try to receive that event, fail and the thread is restarted again - this could create a loop in one of the servers which is detrimental to the other servers on the platform.
 * the error in the old thread causes the event to be lost.  The recovery of the thread is successful but the event is lost.  In most Egeria situations where federated queries are in operation, the loss of an event is no problem.  But if the repositories are relying on the events then the server needs to be able to log that an event is lost.

Another scenario is that something else in the platform is causing th out of memory exception - at this point the server that the kafka connector belongs to may no longer be viable.

planetf1:
In the out of memory case it may not just be the server that isn't viable but in fact the platform (ie the Java process). In such cases it may be more appropriate for any supervising logic (such as k8s) to restart the platform entirely.

jnhagelberg:
@mandy-chessell @planetf1 I just got back from vacation.  What is the current state of this?  Are there any additional changes I need to make?  It seems like we are in agreement that it can be merged.

mandy-chessell (approved):

jnhagelberg:
Thanks!!

Issue #5461, opened 2021-07-05
Coco Pharma example security for 'admin' user


cmgrote:
At the moment the polymer-based UI only seems to show menu options other than "asset catalog" for a single user: `admin`.  Yet if the notebooks are followed including configuring security for the users, then most (all?) of the menu options for the `admin` user are actually broken (i.e. asset catalog shows no types populated in the drop-down, which means no searches can be run; type explorer returns an error about not being able to access the server; and the same for repository explorer).

Disabling the security configuration as part of running the lab Notebooks means that all of these various menu options then work fine for `admin`, so it seems like there is something missing in the security configuration?  Either:

- Giving access to this `admin` NPA to access all of those various pieces of functionality (via the security connector?)
- Giving some other user (like `garygeeke`) access to the other menu options of the UI besides only the asset catalog (not sure where / how this is configured)?

planetf1:
I think there are two distinct pieces of functionality affecting what can be seen/done
 a) There is specific code in the polymer UI which controls which menu items are available to which users  @bogdan-sava should be able to describe this more.
 b) The coco pharma security plugin governs what is permissible at the backend for the server's services. This plugin is only a sample, but is of course relevant for the coco/labs use case
 
 Agree it would be good to have consistency and to specify what users can be used in the coco use case

cmgrote:
Security configuration seems to be defined primarily in: `CocoPharmaServerSecurityConnector`

bogdan-sava:
Indeed.
In UI content is role based, and all content is configured only for "admin" role.
The username is then sent on API request to backend which suppose to enforces security using SecurityConnector and responds accordingly.

content showed to roles is configured in application.properties by default like this:

```
role.visibleComponents.USER=asset-catalog,about,asset-lineage,vertical-lineage
role.visibleComponents.ADMIN=*
```

meaning for role USER asset-catalog,about,asset-lineage components are visible
while for the role ADMIN all content is visible


I'll change these role names to 'COCO_PHARMA_USER' and 'COCO_PHARMA_ADMIN' and reassign sample users.
I will assign `garygeeke` to COCO_PHARMA_ADMIN, so he will have access to all content.


Maybe later on will configure roles to personas matching [personas](https://github.com/odpi/data-governance/tree/master/docs/coco-pharmaceuticals/personas) , but not sure how to name the roles and what content should be available for those

mandy-chessell:
[...]

Issue #5460, opened 2021-07-05
Should asset catalog OMAS be explicitly excluding subtypes in search results?


cmgrote:
At the moment, the asset catalog OMAS search method (`searchByType`) appears to be intentionally excluding subtypes in its search results:

```
// filter out the entities that are retrieved with the enterprise search.
// metadataCollection.findEntitiesByProperty includes in the search the subtypes of the provided type
if (!typesFilter.contains(entityDetail.getType().getTypeDefGUID())) {
    continue;
}
```

This means that with the default search types used (if none are explicitly configured) of `GlossaryTerm`, `Asset` and `SchemaElement` it is probably only `GlossaryTerm` that will ever return any results (the others are unlikely to have instances that exist at these super-type levels, but rather at some sub-type level like `DataFile`, `CSVFile` or `RelationalColumn`).

This means that the default behaviour of the UI makes it basically impossible to ever see any search results -- despite the fact that the performance hit of finding the results (enterprise connector federating queries to the cohort and retrieving all those subtype results back) is already done at the point at which they're explicitly excluded.

Is this really the intended behaviour?  If so, why?

(^^ @lpalashevski as discussed)

lpalashevski:
Right. This is intended behaviour but not from asset catalog perspective :) I agree that is limiting functionallity as it is now. It was added to enhance user experiance but it is done in a wrong way, the idea was to do some smart filtering in the backend but not affect the existing functionallity (that means do it on the UI API and not asset catalog). So it was added as an intermediate step (or quick fix) before the UI gets enahnced and we allow smarter filtering input (like have two options lists one for assets and the other for related subtypes or similar).

I am also labeling it as "bug". Lets discuss the solution on the call tomorrow.

lpalashevski:
Addressed in #5486

lpalashevski:
Fixed with #5486

PR #5462, opened 2021-07-05
Adds an initial draft lab for lineage

cmgrote:
Having it operate correctly will currently depend on #5460 and #5461

### DIFF
--- a/open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
+++ b/open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
@@ -258,7 +258,7 @@
     "\n",
     "### Register an external tool\n",
     "\n",
-    "We'll begin by registering an external data platform that was used to create the data files and processing that we will see in the lineage. Gary Geeke will carry out this step as an administrative user against cocoMDS2 where governance is done."
+    "We'll begin by registering an external data platform that was used to create the data files and processing that we will see in the lineage. Peter Profile will carry out this step against cocoMDS1 in the Data Lake platform."
    ]
   },
   {
@@ -297,7 +297,7 @@
     "}\n",
     "\"\"\"\n",
     "\n",
-    "registerTool(\"cocoMDS2\", corePlatformURL, \"garygeeke\", json.loads(engineDoc))"
+    "registerTool(\"cocoMDS1\", dataLakePlatformURL, petersUserId, json.loads(engineDoc))"
    ]
   },
   {


planetf1:
Brilliant to have this under way :-) One minor note (albeit draft) - strip out the results in the cells when saving to git

cmgrote:
Output should be stripped now...  Only change to existing lab is to update the Asset Catalog OMAS on cocoMDS2 with some additional configuration (previously it had no additional configuration properties) -- this is necessary at the moment for the Polymer-based UI to actually work (per #5460).

Any reason to delay this merge, given it's only fixing something that was not previously working + adding a new work-in-progress notebook (marked as such)?

mandy-chessell:
Why has cocoMDS2 been chosen? I would have thought the Egeria UI users would be on cocoMDS4 since this is the data lake users' access point?  The governance zones are set up with this assumption.  Is there a problem with cocoMDS4?

The governance team would be on cocoMDS2 but I would have expected them to use the react ui since it is where we are adding the editing of glossary terms/assets etc

cmgrote:
> Why has cocoMDS2 been chosen?

This is where the combination of OMASes needed by the polymer-based UI are all running: Asset Catalog OMAS, Asset Lineage OMAS, etc.  I don't think AL OMAS was running in cocoMDS4 (but maybe I missed it?)

For some reason TEX was also not working against cocoMDS4, while it is against cocoMDS2...

mandy-chessell:
We should make cocoMDS4 work for this use case - since the Egeria UI is meant for the data lake users - if this means enabling the OMASs in cocoMDS4, it makes sense - I am not sure why Tex will not work.

cmgrote:
> We should make cocoMDS4 work for this use case - since the Egeria UI is meant for the data lake users - if this means enabling the OMASs in cocoMDS4, it makes sense - I am not sure why Tex will not work.

Ok, that's done.  It seems that actually neither TEX nor REX works: the difference appears to be that by default the "Enterprise" box is ticked on REX, and that's why REX appears to work.  Unticking this box and trying to use `cocoMDS4` on REX ends up giving the same error on REX as it does on TEX.  (And despite there being an "Enterprise" box on TEX it seems to always be greyed-out -- I can't figure out how to tick it to enable it.)

It appears that using cocoMDS1 on the same platform (x-datalake:9443) works, just that cocoMDS4 does not work...  Any further ideas given this context? 🙏 

(The end-to-end lineage view also does not work with cocoMDS4 even after adding the asset lineage OMAS, so I'm hoping whatever is causing cocoMDS4 to be broken everywhere else also will resolve that -- as it basically doesn't work for anything in the UI at the moment...)

cmgrote:
Ah, just spotted this difference in the cocoMDS1 vs cocoMDS4 config:

- cocoMDS1 configures a repository (`configureMetadataRepository(adminPlatformURL, adminUserId, mdrServerName, metadataRepositoryType)`)
- cocoMDS4 does not (just has a comment stating `# Note: no metadata repository or collection configuration here`)

It seems that various aspects of the UI (i.e. TEX) rely on such an underlying repository (?)  Should we add a repository to cocoMDS4, use a different MDS for the UI, or some other fix?

It seems this UI has been broken for probably the last 3-4 releases, so I'm keen that we at least have it in a somewhat usable state even if we work on a longer-term fix -- I'd rather not wait another 4-6 months before we can use it without needing to hack through all of the notebooks manually every time to reconfigure the default settings into workable ones...

mandy-chessell:
Each of the coco servers supports a different pattern.  So we need cocoMDS4 to run without a reposiotry to flush out errors such as this one with TEX.   We need to fix the TEX in the Egeria UI to support the enterprise client - the react UI version does this.

planetf1:
* I don't see an issue with checking in a new notebook in progress (which is marked as such) as long as we are happy with changes elsewhere not breaking anything. Even for short term use with caveats this is helpful as currently we have ZERO support for trying out lineage in the notebooks.
* Note that we have rex and tex in polymer ui and in react. I've not used them (nor test them) in the former. It would be good to qualify which you refer to when discussing issued
* TEX is not available on cocoMDS4 in the reactUI -- ie it is not configured in the notebooks where we setup the view service for tex. The same is true of REX. 
* The above perhaps suggests you are referring to a different, untested (by me) version of rex and tex. This is a concern if the capability is differing between them. they should either be as identical as their containment allows, or only be in once place?
* You are trailing new ground trying to get the polymer UI working with cocoMDS4 - but it's a good thing to do, just bear in mind you're probably the first.... So I'm not sure I'd say it's only 3-4 releases it's been broken, rather that this integration between the coco pharma environment, and the lab infrastructure (in terms of coco's setup) hasn't been tested with the polymer ui.
* I agree with Mandy that we should not add a repo at cocoMDS4 - the error is elsewhere

planetf1:
I'd also suggest that the user for tex/rex is quite different to asset search or lineage.
The latter are more business oriented, typical of a data lake user, and cocoMDS4 is a good target - but note that it's possible different UI options there should be configurable with different back-end egeria servers. The OMASs for lineage could be different to asset search for example?
Then tex/rex is very much an admin usage typically, or for developers. the closest persona at coco is garygeeke

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "            ]\n",
+    "        },\n",
+    "        \"endpoint\": {\n",
+    "            \"class\": \"Endpoint\",\n",
+    "            \"headerVersion\": 0,\n",
+    "            \"address\": \"server.cocoMDS4.omas.assetlineage.outTopic\"\n",
+    "        },\n",
+    "        \"configurationProperties\": {\n",
+    "            \"producer\": {\n",
+    "                \"bootstrap.servers\": \"lab-kafka:9092\"\n",
```
planetf1, line 173: this works in k8s but will fail when run locally or in docker compose. See the config we use for the Event Bus

So we probably want something like/:
```
    "        \"bootstrap.servers\": eventBusURLroot\n",
```

Also repeated for the consumer beneath

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "            },\n",
+    "            \"local.server.id\": \"39b54be8-229f-4480-8a41-242b1c822669\",\n",
```
planetf1, line 175: should this come from config?

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "            \"consumer\": {\n",
+    "                \"bootstrap.servers\": \"lab-kafka:9092\"\n",
```
planetf1, line 177: variable for event bus as above

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "        \"headerVersion\": 0,\n",
+    "        \"displayName\": \"Lineage Graph Connection\",\n",
+    "        \"description\": \"Used for storing lineage in the Open Metadata format\",\n",
+    "        \"connectorType\": {\n",
+    "            \"class\": \"ConnectorType\",\n",
+    "            \"headerVersion\": 0,\n",
+    "            \"connectorProviderClassName\": \"org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.graph.LineageGraphConnectorProvider\"\n",
+    "        },\n",
+    "        \"configurationProperties\": {\n",
+    "            \"storage.directory\": \"data/servers/open-lineage/lineage-repository/berkeley\",\n",
```
planetf1, line 192: shouldn't the path use   the  'olsServerName' rather than hardcoded 'open-lineage' ?

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "            \"index.search.directory\": \"data/servers/open-lineage/lineage-repository/searchindex\",\n",
+    "            \"storage.backend\": \"berkeleyje\",\n",
+    "            \"gremlin.graph\": \"org.janusgraph.core.JanusGraphFactory\",\n",
+    "            \"index.search.backend\": \"lucene\"\n",
+    "        }\n",
+    "    },\n",
+    "    \"accessServiceConfig\": {\n",
+    "        \"serverName\": \"cocoMDS4\",\n",
+    "        \"serverPlatformUrlRoot\": \"https://lab-datalake:9443\"\n",
```
planetf1, line 201: as with the event bus this needs to come from the relevant variable as we use in the config notebook, so that it works in all environments not just k8s

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+    "        {\n",
+    "            \"jobName\": \"LineageGraphJob\",\n",
+    "            \"jobInterval\": 120,\n",
+    "            \"jobEnabled\": true\n",
+    "        },\n",
+    "        {\n",
+    "            \"jobName\": \"AssetLineageUpdateJob\",\n",
+    "            \"jobInterval\": 120,\n",
+    "            \"jobEnabled\": true,\n",
+    "            \"jobDefaultValue\": \"2021-12-03T10:15:30\"\n",
```
planetf1, line 213: is this an end date or beginning? Could we calculate it instead?

cmgrote:
> TEX is not available on cocoMDS4 in the reactUI -- ie it is not configured in the notebooks where we setup the view service for tex. The same is true of REX. This restriction of where tex/rex can access was done both for usability and security.

This implies that we're using REX and TEX in the reactUI against something other than cocoMDS4...  Did I understand that correctly?  If so, it doesn't match for me the restriction that we must use cocoMDS4 in the polymer UI -- but putting that to one side for now...  Which server are REX / TEX configured to work against in the react UI at the moment?  (Or do they not actually work in any UI at the moment?)

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "e1cd0151-292c-4d32-b141-be61529946e4",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "deployServerToPlatform(adminPlatformURL, \"garygeeke\", \"open-lineage\", corePlatformURL)\n",
+    "activateServerOnPlatform(\"open-lineage\", \"unused\", corePlatformURL)"
```
planetf1, line 247: since we had olsServerName above wouldn't we use the var here

open-metadata-resources/open-metadata-labs/lineage/lineage.ipynb
```diff
@@ -0,0 +1,692 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "3ac395bd-87ad-438c-9fb3-bf039a1f6f32",
+   "metadata": {},
+   "source": [
+    "### Create some example files\n",
+    "\n",
+    "Next, we'll simulate that external tool sharing metadata about some of the files that it has processed: both the files themselves, as well as the fields within them.\n",
+    "\n",
+    "These will be created by Peter Profile against cocoMDS1 in the Data Lake platform."
```
planetf1, line 312: more a scenario comment, but I wonder if this would be more suited to cocoMDS5 (business systems)?

cmgrote:
Yes, there is lots to clean up in the lineage notebook -- that's why I said it's still work-in-progress 😉

planetf1:
_> This implies that we're using REX and TEX in the reactUI against something other than cocoMDS4... Did I understand that correctly?_

Yes, the definition of the view service in the config notebook lists the servers that are valid targets. See cell 23 ie 'Configuring the View Server and View Services' in that config notebook.

That's not to say the list is perfect, or everything will work. It's representative of what APIs will return - so if the user doesn't have permission, you'll see an error etc. Over time the errors have improved but I'm sure they could be improved further

 _If so, it doesn't match for me the restriction that we must use cocoMDS4 in the polymer UI_ 
 The issue here may be that the polymer UI is assuming one server is suitable for everything. In any case REX/TEX (at least in the reactUI) initially go through the view server (cocoView1) to get the list of servers they are allowed to connect to. There is no hardcoded server in the UI.
 
 _-- but putting that to one side for now... Which server are REX / TEX configured to work against in the react UI at the moment? (Or do they not actually work in any UI at the moment?)_
 They definately 'work' though I'm sure there are still bugs/glitches. There may be some open. But all servers with a repo work to some extent. We may not have tested proxies fully.

planetf1:
> Yes, there is lots to clean up in the lineage notebook -- that's why I said it's still work-in-progress 😉 Let's not focus on those bits until we've addressed all these other broader config questions about the cohort and servers against which things should be operating.

Agreed, some of these are minor, just adding feedback. I do think using variables for the platform addresses is the one change I would suggest doing even now, as it makes it easier for others to consume outside k8s, plus we already have an established pattern for it

planetf1:
One additional issue to beware of when using the Polymer UI, is that the static content is found in the 'egeria-ui' docker image.

For the last few months this has been build as version '2.8.0' (this tag comes from the package.json). Previously the 'released' helm charts would use '2.8.0' and the 'master' one would use 'latest'. In particular the former would typically be cached (as IfNotPresent was set). Additionally there's no fix between the egeria/chart version and the ui since this version is remaining static.

For the react UI we are incrementing the release version each release - in fact we set the version in package.json to something like 2.11.0-rc.0 during development, and then bump it to 2.11.0 on release. This prevents any caching confusion. It also keeps the versions aligned between egeria and the react ui. (the versioning startegy for the charts may need tweaking to deal with this subtlety in 3.0)

Just worth mentioning in case you find you are running old code. (At a minimum that tex/rex version not including the dependence on view services - so wondering if other aspects like error handling may not have picked up fixes from the react UI)

cmgrote:
I think there are too many moving pieces / open questions / issues here to try to address in a PR discussion...

Rather than trying to tactically unpick I'll suggest we just close this PR and start over at some point down the road -- probably starting with the storyline we want to expose via the lab (?)

PR #5469, opened 2021-07-07
[ui-chassis-spring] fix for https://github.com/odpi/egeria/issues/5461

bogdan-sava:


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/auth/db/DemoInMemoryInitializingBean.java
```diff
@@ -18,44 +18,47 @@
 @ConditionalOnProperty(value = "authentication.source", havingValue = "db")
 public class DemoInMemoryInitializingBean implements InitializingBean {
 
+    private static final String PASSWORD = "secret";
```
planetf1, line 21: THough long term it would be good to remove even demo passwords from egeria, for now, if the password is being changed, I think we should add to the release notes so that those using the demo ids know what to use?

bogdan-sava: I am creating another PR to remove the samples into a separate module. Eventually move it to another repo?! 
But I need to set a password for sample users.

planetf1: The change is fine ... I'm just wondering how we communicate it to anyone currently using the UI sample. Perhaps that isn't currently documented anyway so nothing to change? (an improvement can be done in future as we refactor). feel free to close this conv if you're ok with what we have currently.

Issue #5162, opened 2021-05-07
Adding new sample security connector for LDAP


aproeve:
Currently in the process of replacing the hard-coded user information to an LDAP directory & issuing LDAP calls.

planetf1:
If you check in any LDAP files you may be able to clean up https://github.com/odpi/egeria-samples/issues/23 too (After the persona docs, ideally the ldap .ldif files should be the authoritative definition of the coco org)

PR #5488, opened 2021-07-15
#5162 LDAPDraftv2

aproeve:
Signed-off-by: Alexcia <aproeve.ptech@gmail.com>

### DIFF
--- a/open-metadata-resources/open-metadata-samples/open-metadata-security-samples/src/main/java/org/odpi/openmetadata/metadatasecurity/samples/CocoPharmaServerSecurityProvider.java
+++ b/open-metadata-resources/open-metadata-samples/open-metadata-security-samples/src/main/java/org/odpi/openmetadata/metadatasecurity/samples/CocoPharmaServerSecurityProvider.java
@@ -18,7 +18,7 @@ public class CocoPharmaServerSecurityProvider extends OpenMetadataServerSecurity
     {
         super();
 
-        Class<?>        connectorClass = CocoPharmaServerSecurityConnector.class;
+        Class<?>        connectorClass = AbstractCocoPharmaServerSecurityConnector.class;
 
         super.setConnectorClassName(connectorClass.getName());
     }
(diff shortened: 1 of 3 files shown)


open-metadata-resources/open-metadata-samples/open-metadata-security-samples/src/main/java/org/odpi/openmetadata/metadatasecurity/samples/CocoPharmaServerSecurityProvider.java
```diff
@@ -18,7 +18,7 @@ public CocoPharmaServerSecurityProvider()
     {
         super();
 
-        Class<?>        connectorClass = CocoPharmaServerSecurityConnector.class;
+        Class<?>        connectorClass = AbstractCocoPharmaServerSecurityConnector.class;
```
mandy-chessell, line 21: I know this is still work in progress but here are some hints.

The connector class can not be an abstract class because this is the name of the class that is used to create a new bean.

There are two approaches to designing the two modes of operation for this connector.
* Have two independent connectors - in which case they each need their own connector provider
* Have a single connector class that has the business rules in it.  When it needs to look up which groups a user is in, it has an interface to the user data.  There are two implementations of this interface - one for the in-memory user maps and one for LDAP.  The connector chooses which one it uses by a property in the connection's configuration properties.

aproeve: Hi Mandy! 

I'm very grateful for your input! I've been working with Nigel & David trying to develop this connector and I heard you spoke on connectors during the last call. I'll check out that call as I did need some guidance on how to go about completing these connector(s). 

I am going to keep your suggestion in mind, specifically the one about creating two connectors. As of right now we're trying to establish full functionality of the LDAP calls & such. Once that is done I'll be moving onto the connector side. 

I'll be posting a couple more of these drafts before I request a merge as there's plenty of testing that needs to be done.

Forgive my sluggishness, this is all very new to me but I'm very dedicated towards creating something useful for you guys.

I'll work more on this, expect a new PR soon!

PR #5490, opened 2021-07-19
Add support for tabular_file_column 

pmarius22:


lcpopa (approved):

lpalashevski (approved):

open-metadata-implementation/access-services/asset-lineage/asset-lineage-server/src/main/java/org/odpi/openmetadata/accessservices/assetlineage/handlers/AssetContextHandler.java
```diff
@@ -88,10 +88,9 @@ public AssetContextHandler(InvalidParameterHandler invalidParameterHandler, OMRS
 
         final String typeDefName = entityDetail.getType().getTypeDefName();
         switch (typeDefName) {
-            case TABULAR_COLUMN:
```
lpalashevski, line 91: Approved with one note: we might need to keep this case and allow context to be created for TabularColumn for some time -- There is a chance we still use this type when creating Data/CSVFile with DataEngine. We need to check this.

lpalashevski (approved):

PR #5509, opened 2021-07-28
handle delete events

pmarius22:

### DIFF
--- a/open-metadata-implementation/access-services/data-engine/data-engine-api/src/main/java/org/odpi/openmetadata/accessservices/dataengine/model/Attribute.java
+++ b/open-metadata-implementation/access-services/data-engine/data-engine-api/src/main/java/org/odpi/openmetadata/accessservices/dataengine/model/Attribute.java
@@ -262,4 +262,25 @@ public class Attribute extends Referenceable {
      */
     private String validValuesSetGUID;
 
+    /**
+     * The name of the type for this schema element
+     * -- GETTER --
+     * Return name of the type of this schema element
+     * @return name of the type
+     * -- SETTER --
+     * Set the name of the type for this schema element
+     * @param typeName the name of the type of data
+     */
+    private String typeName;
+
+    /**
+     * Unique identifier of the type for this schema element
+     * -- GETTER --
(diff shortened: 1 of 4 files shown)


popa-raluca (approved):

open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/graph/LineageGraphConnector.java
```diff
@@ -858,5 +813,23 @@ public LineageVertexResponse getEntityDetails(String guid) {
     public boolean isEntityInGraph(String guid) {
         return !g.V().has(PROPERTY_KEY_ENTITY_GUID, guid).toList().isEmpty();
     }
+
+    /**
+     * commit the graph transaction if graph transactions are enabled
+     * @param g the graph traversal
+     */
+    private void commitTransaction(GraphTraversalSource g) {
+        if (graphFactory.isSupportingTransactions()) {
```
davidradl, line 822: I think this refactoring change is good, as you are refactoring - I wonder if you could put these methods in a place where the other Egeria code issuing commits and rollback like this could call them. I notice that the other callers of commit and rollback  in Egeria do not have the graphFactory.isSupportingTransactions() check. We could then start to accumulate common code that users of the graph call.

marius-patrascu: Hey David, I think it would be a good idea to tackle your proposal in a separate issue/PR. What do you think?

PR #5582, opened 2021-08-10
AssetCatalog publish indexing events on out topic

bogdan-sava:


open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/listenenrs/AssetCatalogOMRSTopicListener.java
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Unpack and deliver an instance event to the InstanceEventProcessor
+     *
+     * @param instanceEvent event to unpack
+     */
+    @Override
+    public void processInstanceEvent(OMRSInstanceEvent instanceEvent)
+    {
+        if (log.isDebugEnabled()) {
```
lpalashevski, line 73: Non-functional (insignificant maybe) optiomization we can do: Since we are using slf4j from what I remember if we use parametrized message for debug the check is performed transparently inside the debug method call. Maybe consider:

             `log.debug("Processing instance event: {} ", instanceEvent);`

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/listenenrs/AssetCatalogOMRSTopicListener.java
```diff
@@ -0,0 +1,209 @@
...
+    private void processEntityDetail(EntityDetail entityDetail){
+            String assetType = getAssetType(entityDetail);
+
+            if ( assetType != null ) {
+                AssetConverter assetConverter
+                        = new AssetConverter(entityDetail, null, repositoryHelper, serviceName, serverName);
+                Asset assetBean = assetConverter.getAssetBean();
+
+                if (assetBean == null || !this.inTheZone(assetBean.getZoneMembership())) {
+                    log.debug("Ignored instance event - Asset not in the supported zones!");
```
lpalashevski, line 139: We probabbly need to also do aduditLog here to stick to the pattern.

lpalashevski (approved):
Approved with some proporsals

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/listenenrs/AssetCatalogOMRSTopicListener.java
bogdan-sava: I didn't want to load the audit log with messages that are saying it is working as expected, in case there are many events for assets not in the supported zones.

Issue #5590, opened 2021-08-12
Enhance the server author fetch platforms to include the registered access services.


davidradl:
currently the ui get the platforms using the admin call then gets the registered access services on the platform in 2 calls. 

I have already extended the fetch platforms in the server author view service to augments the returned server objects with an operational status. I intend to extend the response to include the registered access services. This means the ui only needs to issue one call to the server rather than 2. The platform content is not expected to change under the server author (or at least we are not coding for that in the UI at this time!).

davidradl:
I notice that MetadataAccessPointConfigurationClient has a constructor that takes a serverName.
This server name ends up in the top level super constructor and creates a Spring rest admin client.

For the getRegisteredAccessServices() I see
`  public List<RegisteredOMAGService> getRegisteredAccessServices() throws OMAGNotAuthorizedException,
                                                                            OMAGInvalidParameterException,
                                                                            OMAGConfigurationErrorException
    {
        final String methodName  = "getRegisteredAccessServices";
        final String urlTemplate = "/open-metadata/platform-services/users/{0}/server-platform/registered-services/access-services";
`

This is a platform call so there is no server - so I should not have a supply a server on the constructor in order to make this call. 

I propose moving this call to OMAGServerPlatformConfigurationClient , next to the call that gets all the server configurations.

I will deprecate the existing API call - as the admin API is released. 

@mandy-chessell are you ok with this?

davidradl:
A further consideration is that the intent for the Server author is to author a new server on the omag platform that the view service is connected to. So
- maybe I should only use the access services defined on that server
- maybe I should always assume that all access services are available on the platform targeted by the view server config.     
In that case it simplified the callers job, as they do not need to deal with registered access services 
- if I then look to deploy onto another platform I could check whether the target platform supports all the access services that are in the configuration
- the platform targeted by the view service (in terms of the remote end point it talks to) config may not be actually defined as a platform )

I suggest a minimal design as:
- for the server author view service to check if it is defined as a platform - and not start if it is not - with appropriate message
- for the server author view service to check if all the access services are registered - and not start if it is not - with appropriate message. This would save the caller from having to query it.
- at deployment the caller could query the target platform to check that the selected access services in the configuration to be deployed match the registered ones on the target platform. 

@planetf1 @mandy-chessell Are we ok with this suggested approach?

davidradl:
[...]

PR #5591, opened 2021-08-12
#5590 Introduce enhanced Server Author fetch platforms to include all the registered services. 

davidradl:
# Description
Amend the server author to return a fuller list of platforms.

Fixes # (issue) #5590

# How Has This Been Tested?

React UI

The issue talks of wider considerations in this area that may result in more prs for the issue

In addition to Server author changes:
 
- I have changed the application.properties to remove the default of Debug from the logging. This was confusing me as I had assumed that by default we would have no debug logging - so I assumed the entries were errors.

- I have amended the PlatformServers to add 2 new APIs.

- Dino has been amended to call the new Platform services APIs to get the integration and engine registered services as well as the other registered services. 

- server author has been introduced to the Gradle build as it was missing.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -51,6 +51,8 @@ include(':open-metadata-implementation:adapters:open-connectors:data-store-conne
 include(':open-metadata-implementation:adapters:open-connectors:data-store-connectors:file-connectors')
 include(':open-metadata-implementation:adapters:open-connectors:data-store-connectors')
 include(':open-metadata-implementation:adapters:open-connectors:integration-connectors:files-integration-connectors')
+include(':open-metadata-implementation:adapters:open-connectors:integration-connectors:kafka-integration-connector')
+include(':open-metadata-implementation:adapters:open-connectors:integration-connectors:openapi-integration-connector')
 include(':open-metadata-implementation:adapters:open-connectors:integration-connectors')
 // todo - convert to integration connectors and governance action connectors
 //include(':open-metadata-implementation:adapters:open-connectors:governance-daemon-connectors:view-generator-connectors')
@@ -441,6 +443,8 @@ project(':open-metadata-implementation:adapters:open-connectors:data-store-conne
 project(':open-metadata-implementation:adapters:open-connectors:data-store-connectors:file-connectors').projectDir = file('open-metadata-implementation/adapters/open-connectors/data-store-connectors/file-connectors')
 project(':open-metadata-implementation:adapters:open-connectors:data-store-connectors').projectDir = file('open-metadata-implementation/adapters/open-connectors/data-store-connectors')
 project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:files-integration-connectors').projectDir = file('open-metadata-implementation/adapters/open-connectors/integration-connectors/files-integration-connectors')
+project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:kafka-integration-connector').projectDir = file('open-metadata-implementation/adapters/open-connectors/integration-connectors/kafka-integration-connector')
+project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:openapi-integration-connector').projectDir = file('open-metadata-implementation/adapters/open-connectors/integration-connectors/openapi-integration-connector')
 project(':open-metadata-implementation:adapters:open-connectors:integration-connectors').projectDir = file('open-metadata-implementation/adapters/open-connectors/integration-connectors')
 // todo convert to integration connectors and governance action connectors
 //project(':open-metadata-implementation:adapters:open-connectors:governance-daemon-connectors:view-generator-connectors').projectDir = file('open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/view-generator-connectors')
--- a/open-metadata-distribution/open-metadata-assemblies/build.gradle
+++ b/open-metadata-distribution/open-metadata-assemblies/build.gradle
@@ -33,6 +33,8 @@ dependencies {
     implementation project(':open-metadata-implementation:adapters:open-connectors:configuration-store-connectors:configuration-encrypted-file-store-connector')
     implementation project(':open-metadata-implementation:adapters:open-connectors:configuration-store-connectors:configuration-file-store-connector')
     implementation project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:files-integration-connectors')
+    implementation project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:kafka-integration-connector')
+    implementation project(':open-metadata-implementation:adapters:open-connectors:integration-connectors:openapi-integration-connector')
     implementation project(':open-metadata-implementation:adapters:open-connectors:discovery-service-connectors')
     implementation project(':open-metadata-implementation:adapters:open-connectors:governance-action-connectors')
     implementation project(':open-metadata-implementation:adapters:open-connectors:event-bus-connectors:open-metadata-topic-connectors:inmemory-open-metadata-topic-connector')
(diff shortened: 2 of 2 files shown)


open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/MetadataAccessPointConfigurationClient.java
```diff
@@ -71,13 +71,20 @@ public MetadataAccessPointConfigurationClient(String adminUserId,
...
+     * This call is deprecated, as the MetadataAccessPointConfigurationClient client is for configuring a particular
+     * server, not interrogating platform information.
+     *
+     * Please use the call getRegisteredAccessServices in the platform configuration client
+     * OMAGServerPlatformConfigurationClient.
      *
      * @return list of access service descriptions
      * @throws OMAGNotAuthorizedException the supplied userId is not authorized to issue this command.
      * @throws OMAGInvalidParameterException invalid parameter.
      * @throws OMAGConfigurationErrorException unusual state in the admin server.
```
mandy-chessell, line 84: This call is needed - do not deprecate

open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/OMAGServerPlatformConfigurationClient.java
```diff
@@ -278,4 +281,30 @@ public Connection getPlatformSecurityConnection() throws OMAGNotAuthorizedExcept
...
+
+    /**
+     * Return the list of access services for this platform.
+     *
+     * @return list of access service descriptions registered on this platform
+     * @throws OMAGNotAuthorizedException the supplied userId is not authorized to issue this command.
+     * @throws OMAGInvalidParameterException invalid parameter.
+     * @throws OMAGConfigurationErrorException unusual state in the admin server.
+     */
+    public List<RegisteredOMAGService> getRegisteredAccessServices() throws OMAGNotAuthorizedException,
```
mandy-chessell: I do not see the value of adding this call here -if it is not accompanied by equivalent calls for the other types of registered services - engine, view, integration

open-metadata-implementation/view-services/server-author-view/server-author-view-server/src/main/java/org/odpi/openmetadata/viewservices/serverauthor/handlers/ServerAuthorViewHandler.java
```diff
@@ -185,6 +186,10 @@ public ServerAuthorViewHandler(String localServerUserId, String metadataServerUR
                         // do not have a description of the server yet.
                         platform.addStoredServer(storedServer);
                     }
+                    //`/open-metadata/platform-services/users/${userId}/server-platform/registered-services/access-services`;
+                    MetadataAccessPointConfigurationClient metadataAccessPointConfigurationClient = new MetadataAccessPointConfigurationClient(userId, "",resourceEndpoint.getResourceRootURL());
```
mandy-chessell: Strange to be only retrieving the access services.  Seems to be missing requests for retrieving other registered services needed for configuring other types of OMAG Server?

mandy-chessell (changes requested):

open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/MetadataAccessPointConfigurationClient.java
davidradl: OK - I am unsure what to put as the serverName in the constructor of this class. As the call is a platform orientated call without an associated server.
It ends up calling `AdminServicesRESTClient(String serverName,
                            String serverPlatformURLRoot) throws OMAGInvalidParameterException
    {
        final String  methodName = "RESTClient(no authentication)";

        this.serverPlatformURLRoot = serverPlatformURLRoot;

        RESTClientFactory factory = new RESTClientFactory(serverName, serverPlatformURLRoot);

        try
        {
            this.clientConnector = factory.getClientConnector();
        }
        catch (Exception error)
        {
            throw new OMAGInvalidParameterException(OMAGAdminErrorCode.NULL_LOCAL_SERVER_NAME.getMessageDefinition(serverName, error.getMessage()),
                                                    this.getClass().getName(),
                                                    methodName,
                                                    error);
        }
    }`
 The Exception and call seems to want a valid serverName.  Am I missing something?

open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/OMAGServerPlatformConfigurationClient.java
davidradl: I have removed it now - I found the PlatformServices Client.

open-metadata-implementation/view-services/server-author-view/server-author-view-server/src/main/java/org/odpi/openmetadata/viewservices/serverauthor/handlers/ServerAuthorViewHandler.java
davidradl: Yes I agree - I started duplicating the capability of the existing code, I realised the PlatformServices Client that Dino uses was a better place to get this information

open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/MetadataAccessPointConfigurationClient.java
mandy-chessell: It is behaving correctly for the usecase it is designed for - ie configuring a server.  In this case the server name is the server being configured.  Requesting the registered access services is just one of the sequence of calls used to create a metadata access point and/or metadata server.

Use the platform services client if not configuring a specific server.

settings.gradle
```diff
@@ -256,6 +256,10 @@ include(':open-metadata-implementation:view-services:dino-view:dino-view-api')
 include(':open-metadata-implementation:view-services:dino-view:dino-view-server')
 include(':open-metadata-implementation:view-services:dino-view:dino-view-spring')
 include(':open-metadata-implementation:view-services:dino-view')
+include(':open-metadata-implementation:view-services:server-author-view:server-author-view-api')
+include(':open-metadata-implementation:view-services:server-author-view:server-author-view-server')
+include(':open-metadata-implementation:view-services:server-author-view:server-author-view-spring')
+include(':open-metadata-implementation:view-services:server-author-view')
```
planetf1, line 262: Great so far, but we also need project definitions ie similar to these we use for dino:
```
project(':open-metadata-implementation:view-services:dino-view:dino-view-server').projectDir = file('open-metadata-implementation/     view-services/dino-view/dino-view-server')
```
You'll need to also add those to settings.gradle for the above projects.

We shouldn't need the top level directory if empty (server-author-view) but we still have them for other projects, so ok to leave for consistency (I'll clean this area up later when I get back to more gradle work)

planetf1 (changes requested):

open-metadata-implementation/admin-services/admin-services-client/src/main/java/org/odpi/openmetadata/adminservices/client/MetadataAccessPointConfigurationClient.java
davidradl: ok will do

davidradl: done

settings.gradle
davidradl: ok done

Issue #5581, opened 2021-08-10
A Small Error in the Open MetaData Types Documentation


Kelukin:
Documentation URL: https://egeria.odpi.org/open-metadata-publication/website/open-metadata-types/0020-Property-Facets.html

The ReferenceableFacet's stereotype should be a **relationship** instead of an **entity** in the UML Model.

mandy-chessell:
Thanks for pointing these out.  I will update them

mandy-chessell:
They should be fixed now ...

PR #5594, opened 2021-08-14
Connector categories and new connectors

mandy-chessell:
# Description

This PR has two features:
* Additional of connector categories and extensions to the connector type
* Two new integration connectors - one for cataloging topics for Apache Kafka and the other for cataloguing APIs though the Open API Specification

The testing of the two connectors uncovered a problem with the data manager's getXXXByName and findXXX.  Tey included the search string in the path.  If the search string contains a URL then this confused the server receiving the request.  This PR has changed these calls to send the search string in the request body.

There is also some of the documentation for the IT Infrastructure OMAS is in place along with the updates to the open metadata types that this work uncovered.

Fixes # (issue)
#5581

# How Has This Been Tested?

Using the labs to run the 3 cohorts and loading the new metadata through the data connector archive and running the two new connectors to exercise the Data Manager OMAS API.

The OpenAPI connector is only cataloging APIs and APIOperations - it is extracting the parameters but not yet converting them to calls to the API Integrator OMIS.

### DIFF
--- a/open-metadata-implementation/access-services/it-infrastructure/it-infrastructure-server/build.gradle
+++ b/open-metadata-implementation/access-services/it-infrastructure/it-infrastructure-server/build.gradle
@@ -16,6 +16,7 @@ dependencies {
     implementation project(':open-metadata-implementation:common-services:ffdc-services')
     implementation project(':open-metadata-implementation:common-services:repository-handler')
     implementation project(':open-metadata-implementation:common-services:generic-handlers')
+    implementation project(':open-metadata-implementation:common-services:metadata-security:metadata-security-server')
 
     implementation 'com.fasterxml.jackson.core:jackson-annotations'
 


open-metadata-distribution/open-metadata-assemblies/src/main/assemblies/egeria-omag.xml
```diff
@@ -66,6 +66,8 @@
                 <include>org.odpi.egeria:discovery-service-connectors:*</include>
                 <include>org.odpi.egeria:governance-action-connectors:*</include>
                 <include>org.odpi.egeria:files-integration-connectors:*</include>
+                <include>org.odpi.egeria:kafka-integration-connector:*</include>
+                <include>org.odpi.egeria:openapi-integration-connector:*</include>
```
planetf1, line 70: The gradle assembly uses the build.gradle file in this directory to compose an equivilent assembly

new components are added by
a) Adding a dependency
```
 dependencies {

     // Dependencies for the main server assembly - this is modeled on the pom.xml
     // In future we may be able to select by type or label
     implementation project(':open-metadata-implementation:adapters:open-connectors:repository-services-connectors:audit-log-connectors:audit-log-  console-connector')
     ```
b) Adding the particular library ie:
```
            into('server/lib') {
                 // All our connectors, plus discovery sample & the security plugin
                 from { project(':open-metadata-implementation:adapters:open-connectors:repository-services-connectors:audit-log-connectors:audit-  log-console-connector').jar }
                 ```
                 
The gradle assembly is not currently used, but it is built and I have tried to keep it up to date. When we finish the gradle work there will be more checks needed to ensure we have the same content as the maven built version

FYI: @mandy-chessell

mandy-chessell:
OK - thanks for the reminder - will add the updates to gradle

PR #5599, opened 2021-08-18
fixed 5592 to prevent jvm crash when kafka brokers are not reachable

guptaneeru:
# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change. Note that there should always be a related issue alongside a PR as that enables appropriate discussion.

Fixes # (issue)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration. Consider any more risky areas that you need advice on.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ buildscript {
 plugins {
         //id "nebula.lint" version "16.9.0"
         //id "nebula-aggregate-javadocs"
-        id "io.freefair.lombok" version "6.1.0-m3"
+        id "io.freefair.lombok" version "6.1.0"
     }
 
 /* run './gradlew aggregateJavadocs' at top level to build all docs & output to build/docs/javadoc */
@@ -41,7 +41,7 @@ apply plugin: 'nebula-aggregate-javadocs'
 allprojects {
 
     group = 'org.odpi.egeria'
-    version = '3.1-SNAPSHOT'
+    version = '3.2-SNAPSHOT'
 
     // This enforces version checking but is slow to process
     //apply plugin: 'nebula.lint'
@@ -76,19 +76,19 @@ allprojects {
     ext {
         logbackVersion = '1.2.5'
         jacksonVersion = '2.12.2'
-        nettyVersion = '4.1.66.Final'
+        nettyVersion = '4.1.67.Final'
         luceneVersion = '8.9.0'
         gremlinVersion = '3.4.12'
         groovyVersion = '3.0.7'
         janusVersion = '0.5.3'
-        springdataVersion = '2.5.3'
+        springdataVersion = '2.5.4'
         springwebVersion = '5.3.9'
         springldapVersion = '2.3.4.RELEASE'
-        springsecurityVersion = '5.5.1'
-        springbootVersion = '2.5.3'
-        mockitoVersion = '3.11.2'
+        springsecurityVersion = '5.5.2'
+        mockitoVersion = '3.12.4'
+        springbootVersion = '2.5.4'
         junitjupiterVersion = '5.7.2'
-        tomcatVersion = '9.0.50'
+        tomcatVersion = '9.0.52'
         httpcoreVersion = '4.4.14'
     }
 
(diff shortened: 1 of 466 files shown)


guptaneeru:
@planetf1 can you please take a look here.

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
```diff
@@ -420,8 +420,8 @@ int getNumberOfUnprocessedEvents() {
         boolean getRunningBrokers(Properties connectionProperties ) {
 
             boolean found = false;
-            try  (AdminClient adminClient = KafkaAdminClient.create(connectionProperties))
-            {
+            AdminClient adminClient = KafkaAdminClient.create(connectionProperties)
```
planetf1, line 423: If adminClient is null ie the create fails, we will end up exiting the retry loop.
However I can't see any compelling failures that might lead to this so probably ok.
Are you comfortable with KafkaAdminClient behaviour?

planetf1:
You need to sign off the commit as per the DCO. See https://github.com/apps/dco

planetf1 (changes requested):
- Confirm/update behaviour if KafkaAdminClient.create() fails
- Sign DCO

open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
guptaneeru: Handled NPE.

planetf1: Thanks

planetf1 (approved):

planetf1:
Hi @guptaneeru , thanks for making the updates for the now, but there is now a syntax issue that needs resolution before we can merge.

We will be branching this week. this update looks valuable to include in the release.

planetf1:
If you click on 'checks' you should be able to see the details of what failed. You can click each one for more details which should show the exact error.

If it's a code error you should see it when you build locally. I suggest you do a 'mvn clean install' from the top directory so you can fix locally. (We also have a gradle build foot which you can run './gradlew build'

planetf1:
Your error is in /Users/jonesn/IdeaProjects/egeria.maven/open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java at line 423 - the semicolon is missing at the end of this line.

planetf1:
The DCO needs updating.
Also have you done any testing of the changes in both normal and error cases?
You can take the assembly from the build output if it helps.
The fix looks good by inspection but should test some kafka scenario before merging (our FVTs don't currently as they are single-server not cohort based)

guptaneeru:
Nigel, let me think about some ways to manually test it locally. One of the challenge is WKC uses 2.8 and this build might not be compatible with WKC code. I will try.

planetf1 (changes requested):
I think we need one further change. Other than that the code seems to make sense to me having checked hopefully thoroughly. 

Also the DCO is still incomplete - and without this we cannot incorporate your commits.

If you are having trouble I am happy to recode & merge as needed - could you perhaps take a look in the next few days. I'll redo if I don't hear as the issue you raised is valuable to address.

planetf1:
No activity, so closing & replaced with #5684

PR #5620, opened 2021-08-31
Ols use remote graph

pmarius22:
updated code to be compatible with a remote janusgraph server


popa-raluca (approved):

davidradl (commented):
I have had a quick look , the pr title is 'update the OLS code to be compatible with remote janus graph'. I am not sure what changes in the code actually do this  as there are many cosmetic changes. I suggest adding a comment around the code that is required for remote janus - so that it is really obvious to the reviewer what is the important code is.

marius-patrascu:
Hi, David

All the changes in the code are related to using the remote instance of JanusGraph. From my testing, I noticed that the vertex.property() method does not retrieve all the properties of a node and thus the changes for using the valueMap method. Also, not all instances of jaunsGraph support graph variables so we opted instead for creating a node to store the asset lineage update time.
The only changes that are not related to the use of remote graph are those for spacing, those were done automatically by my IDE.

Thanks,
Marius

PR #5621, opened 2021-09-01
HTTP request fix: making room for request body on DELETE

lcpopa:

Param for request body was not existent. Added it, since some DELETE requests have one and instead of placing the object as the body, it's string representation was placed within the url.


lcpopa:
#5634

popa-raluca (approved):

PR #5672, opened 2021-09-03
#5666 update and delete audit log destination APIs

### LINKED ISSUE #5666: Introduce more APIs around the audit log destination authoring
In the admin services there is a call 
```
@PostMapping(path = "/audit-log-destinations/connection")
public VoidResponse addAuditLogDestination
```
 I could introduce update and delete calls  as a rest call and in the admin client. I would then also expose these APIs in the server author view.

 It would then allow the server author user to supply a name and description for their audit log destination - the calling code would map the audit log destination  type (console , default etc) to the appropriate class for the connection. 

As to what to use as the resource identifier:
-there is a Connection open type which extends referencable so it could have a guid. but this is not set in the Connection objects.
      
I see in `ConnectorConfigurationFactory` for example that the connection for console is created like this 
`public Connection getConsoleAuditLogConnection(List<String> supportedSeverities)
    {
        final String destinationName = "Console";

        Connection connection = new Connection();

        connection.setDisplayName(destinationName);
        connection.setConnectorType(getConnectorType(CONSOLE_AUDIT_LOG_STORE_PROVIDER));

        setSupportedAuditLogSeverities(supportedSeverities, connection);

        return connection;
    }`

So even though the connection is a Referencable there is no guid - I assume because it is not persisted in the repository. 

I suggest that the best identifier for me to use in the API would be the display name the server code would only look for this in the CONSOLE_AUDIT_LOG_STORE_PROVIDER type.

davidradl:

# Description

Introduce an update and a delete API to the admin API and equivalents in the server author to allow the server author UI to have a page where audit logs can be identified and  named by the user. The update and delete allows the UI user fine grain control of the auditlog destination authoring.    

Fixes #5666

# How Has This Been Tested?

Using postman through the server author view  I have 
- created a console type connection with name test1 with all the supported severities. 
- updated test1 to change its name to test2 and change the supported severities to only contain 4 
- I then deleted test2
- I have tested the update with a audit log destination name that does not exist and get the expected error
- I have tested the delete with a audit log destination name that does not exist and get the expected error

In the admin API , I have called the audit log destination connection  'connection' so it is consistent with the parameter name of the existing create. For the name of it I am calling it connectionName
In the server author view API, I have called the connection, auditLogDestination and auditLogDestinationName - which is hopefully more useful to the view service caller.    
I have corrected some typos and missing javadoc as well. 

I have not policed for duplicate audit log destination connection names, so would handle this separately if required in another issue. So a delete or update currently will just handle one of a particular name. I was thinking if we want to police for duplicate names we could:
- reject configurations with duplicate audit log destination connection names at start up and every admin rest call that alters them.
- check in the ui and in the server author view service calls. 
- it could be confusing if 2 names are the same with different type or supported severities.
- delete could just delete every audit log destination connection with the requested name. Working from the end of the list would allow be to remove them all from the list in one for loop. 
- we could police for duplicate supported severities - though having 2 the same in the list would have no operational effect as they have no properties.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,31 +113,6 @@ open-metadata-implementation/access-services/subject-area/subject-area-server/sr
 # Ignore fvt Generated output
 open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test/src/**
 
-# Ignore Presentation Server generated output and downloaded content
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/dist/**
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/node_modules/**
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/package-lock.json
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/.vscode
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/*.log
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/yarn.lock
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/node/**
-open-metadata-implementation/user-interfaces/presentation-server/nodejs/.env
-open-metadata-implementation/user-interfaces/presentation-server/node/**
-open-metadata-implementation/user-interfaces/presentation-server/package-lock.json
-
-
-open-metadata-implementation/user-interfaces/presentation-server/cra-client/node_modules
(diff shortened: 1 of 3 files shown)


davidradl:
I will merge at the end of UK Tuesday 7th of September - unless I hear any feedback indicating I shouldn't.

PR #5680, opened 2021-09-10
#5229 check consumer/producer arent null when handling disconnect

### LINKED ISSUE #5229: CohortTopicConnector is not able to gracefully shut down when the Kafka connection initially failed
`CohortTopicConnector` should be able to gracefully shut down when the Kafka connection initially failed. Right now, it is giving a `NullPointerException`

### Initialize cohortTopicConnector
In Egeria2.4:
The logic of OMRSCohortManager.initialize() is:
```java
        try {
            ....
            cohortTopicConnector.registerListener(cohortEventListener, cohortName);
            cohortTopicConnector.start();
            this.cohortTopicConnector = cohortTopicConnector;
            this.cohortEventListener = cohortEventListener;
        catch (Throwable  error){}
```
If kafka connection fails to initialize, cohortTopicConnector.start() would go straight to catch block without calling this.cohortTopicConnector = cohortTopicConnector; In this case, we have a null cohortTopicConnector.

In Egeria 2.8:
The logic of OMRSCohortManager.initialize() changes to:
```java
        try {
            ....
            this.cohortTopicConnector = cohortTopicConnector;
            this.cohortEventListener = cohortEventListener;
        catch (Throwable  error){}
```
The first two lines move to `OMRSCohortManager.setSecurityVerifier()` method, and this method will only be executed in `OMRSMetadataHighwayManager.setSecurityVerifier()` when cohortManager != null, after OMRSCohortManager is already initialized.
```java
    public void setSecurityVerifier(OpenMetadataEventsSecurity securityVerifier){
        if (cohortTopicConnector != null)
                cohortTopicConnector.registerListener(cohortEventListener, cohortName);
                cohortTopicConnector.start();
   }
```
Even though Kafka connection fails to initialize, we still get a not null cohortTopicConnector here.

### When shutting down cohortTopicConnector
```java
            if (cohortTopicConnector != null) {
                cohortTopicConnector.disconnect();  
            }
```
In Egeria 2.4, since we have a `cohortTopicConnector=null`; `cohortTopicConnector.disconnect()` won't be excuted.
However, In Egeria 2.8, even though Kafka connection fails to initialize, we still have a not null cohortTopicConnector. cohortTopicConnector.disconnect() will be executed and `consumer.safeCloseConsumer()` will throw an NPE. 

Wrapping consumer.disconnect() and producer.disconnection() with a try catch block in `KafkaOpenMetadataTopicConnector.disconnect()` can fix it.

planetf1:

# Description

Checks for producer/consumer not being null in the Kafka topic connector when handling a connector disconnect

Fixes #5229

# How Has This Been Tested?

in progress... (hence draft)

See discussion in #5229


mandy-chessell (approved):
Agreed - useful fix

Issue #5687, opened 2021-09-13
Intermittent failures in Data Engine OMAS FVT


planetf1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Seeing the error below in some environments, building direct from master (3.2-SNAPSHOT - 13 Sep)
```
[INFO] Running org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT
[ERROR] Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.165 s <<< FAILURE! - in org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT
[ERROR] org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT.registerExternalTool(String, DataEngineClient, RepositoryService)[1]  Time elapsed: 0.068 s  <<< ERROR!
java.lang.NullPointerException
	at data.engine.fvt@3.2-SNAPSHOT/org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT.registerExternalTool(SoftwareServerCapabilityFVT.java:55)

[ERROR] org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT.registerExternalTool(String, DataEngineClient, RepositoryService)[2]  Time elapsed: 0.085 s  <<< ERROR!
java.lang.NullPointerException
	at data.engine.fvt@3.2-SNAPSHOT/org.odpi.openmetadata.accessservices.dataengine.fvt.SoftwareServerCapabilityFVT.registerExternalTool(SoftwareServerCapabilityFVT.java:55)

[INFO]
[INFO] Results:
[INFO]
[ERROR] Errors:
[ERROR]   SoftwareServerCapabilityFVT.registerExternalTool:55 NullPointer
[ERROR]   SoftwareServerCapabilityFVT.registerExternalTool:55 NullPointer
[INFO]
[ERROR] Tests run: 14, Failures: 0, Errors: 2, Skipped: 0
```

The exception appears to occur on
```
        assertEquals(softwareServerCapability.getEngineType(), entity.getProperties().getPropertyValue(DEPLOYED_IMPLEMENTATION_TYPE).valueAsString());
```

### Expected Behavior

Build works correctly

### Steps To Reproduce

trying to establish cause.
The error occurs even after clearing out ~/.m2
[...]

PR #5689, opened 2021-09-14
DE OMAS - Add connector type, connection and endpoint for files and databases using common handlers

alexandra-bucur:

### DIFF
--- a/open-metadata-implementation/access-services/subject-area/subject-area-server/src/test/java/org/odpi/openmetadata/accessservices/subjectarea/server/mappers/graph/TestRelationshipTypeMapper.java
+++ b/open-metadata-implementation/access-services/subject-area/subject-area-server/src/test/java/org/odpi/openmetadata/accessservices/subjectarea/server/mappers/graph/TestRelationshipTypeMapper.java
@@ -16,7 +16,7 @@ public class TestRelationshipTypeMapper {
             RelationshipTypeMapper.mapOMASRelationshipTypeToOMRSRelationshipTypeGuid("operation", null);
             assertTrue(false, "Expected null to fail");
         } catch (InvalidParameterException e) {
-            e.printStackTrace();
+           // we expect to get here
         }
         for (RelationshipType relationshipType : RelationshipType.values()) {
             if (relationshipType == RelationshipType.Unknown) {
@@ -24,7 +24,7 @@ public class TestRelationshipTypeMapper {
                     RelationshipTypeMapper.mapOMASRelationshipTypeToOMRSRelationshipTypeGuid("operation", relationshipType);
                     assertTrue(false, "Expected unknown to fail");
                 } catch (InvalidParameterException e) {
-                    e.printStackTrace();
+                    // we expect to get here
                 }
             } else {
                 String guid = RelationshipTypeMapper.mapOMASRelationshipTypeToOMRSRelationshipTypeGuid("operation", relationshipType);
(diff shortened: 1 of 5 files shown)


open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/converters/ConnectorTypeConverter.java
```diff
@@ -0,0 +1,100 @@
...
+
+            if (returnBean instanceof ConnectorType) {
+                ConnectorType bean = (ConnectorType) returnBean;
+
+                if (entity != null) {
+
+                    InstanceProperties instanceProperties = new InstanceProperties(entity.getProperties());
+
+                    bean.setQualifiedName(this.removeQualifiedName(instanceProperties));
+                    bean.setDisplayName(this.removeDisplayName(instanceProperties));
```
davidradl, line 50: I notice that you are removing properties from the instanceProperties in the entity. This is modifying a method parameter, I assume if the caller then addresses the entity it will no longer have all its instance properties. It seems a bit dangerous to have this side effect. It would be safer to clone the instance properties to a new object and remove properties from that object which only has this method scope.

alexandra-bucur: I was under the impression that the constructor of InstanceProperties does a deep copy and it's safe to use. This happens on line 47. Am I missing something?

alexandra-bucur: @davidradl This pattern is inherited from OpenMetadataAPIGenericConverter and it's widely used inside a series of converters. The purpose of it is to remove all properties from the deep copy that can be assigned to proper model members and have all the others inside the extendedProperties map.

popa-raluca (approved):

planetf1:
This build is failing in the exact same place I had *locally* - which for me was addressed by #6575 (I used an earlier version in draft). Yet the PR/merge builds had passed. I would recommend we try and get 6575 merged first ?

planetf1:
^^ @lcpopa

planetf1:
See also https://github.com/odpi/egeria/issues/5687

planetf1:
I notice the build is still failing in maven on data engine fvt cc: @lcpopa 
Oddly the gradle build *is not* failing - yet it seems to be running the fvts ....

planetf1:
I pulled this PR and tried building it on my mac, and linux systems. In both cases data Engine OMAS FVT failed in maven

alexandra-bucur:
@planetf1 It seems it's related tot the ConnectionAndEndpointFVT. It was indeed developed by @lcpopa, but with the current changes the test needed a serious update. I had to change the groovy script in order to load the ConnectorTypes archive. I tried to use content-packs/DataStoreConnectorTypes.json and a relative path to it inside the script.

The maven build here fails because of: 
`[INFO] Tue Sep 21 14:24:44 UTC 2021 servergraph Exception OCF-FILE-OPEN-METADATA-ARCHIVE-STORE-CONNECTOR-0002 Supplementary information: log record id c6b42840-3f72-4e18-a07d-0610ef80bdbb java.io.FileNotFoundException returned message of /home/runner/work/egeria/egeria/../../../../content-packs/DataStoreConnectorTypes.json (No such file or directory) and stacktrace of 
[INFO] java.io.FileNotFoundException: /home/runner/work/egeria/egeria/../../../../content-packs/DataStoreConnectorTypes.json (No such file or directory)`

When I run it locally through Maven, it works, so I guess I need to change it somehow so it stays the same no matter where it's running.

planetf1:
The FVTs currently run (cwd) in their own directories ie ${project.build.outputDirectory} in part so that any output files etc are localized to one fvt and they can't contaminate each other. Also so that for gradle they can be run in parallel

As a consequence when the jar is launched we specifically set the loader path & find the jar via a relative path back to the unpacked assembly. 

I think in your case you want to find content-packs/xxx.json and of course as a relative path this is not found

Since the API to load the content pack allows a fully qualified path, I think you should just be able to use `fvt.distdir`variable to locate it?

planetf1:
I see you've fixed this in maven now ... :-) My local issue was something different - intermittent data engine failures in my local environment that I'm still seeing

PR #5789, opened 2021-10-04
Dependabot 20211004 - dependency updates for October 2021

planetf1:
# Description

Aggregated dependabot fixes as detected 1 Oct 2021

# How Has This Been Tested?

build/fvt

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,6 +217,7 @@ allprojects {
 	        runtimeOnly("javax.servlet:javax.servlet-api:4.0.1");
             testImplementation("junit:junit:4.13.2")
             testImplementation("org.glassfish:javax.json:1.1.4")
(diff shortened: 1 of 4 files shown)


planetf1:
In testing a few anomolies were noted
 - config file connector & subject area client were using a hardcoded version of jupiter, and with incorrect scope - this is corrected (in gradle build)
 - the cassandra libraries are causing some issues with more recent versions of dependencies. Specifically the health metric plugin is failing at startup as it is trying to connect to a cassandra host (when we don't use/need cassandra by default). it is therefore disabled by default
 - the hibernate validator used by janusgraph cql support is also pulled forward to current versions as it was causing issues with the java validation API. (and the above suggests the previously this wasn't working, this change may have fixed it, and hence caused the above runtime issue)

Issue #2391, opened 2020-01-10
Add additional architecture (arm64) to docker images for pi


planetf1:
In 1.11 we wish to support egeria cohorts including many iot devices - we will use some actual raspberry pis in this scenario. Therefore targetting the build of the images a couple of releases before this (1.9) to ensure things are in place.

Whilst we can now run the egeria code locally on the pi without containers, it would be nice to also have docker images available, and indeed it is possible to run under k3s (a lightweight k8s implementation).

The docker repo doies allow for multiple architectures, and there's a useful article on cross building with 'buildx' at https://blog.heyal.co.uk/docker-bham-jan/

Raising this issue to track
 - ensuring we can run docker images locally on the pi4
 - getting appropriate images published to the same registries as used for x86_64
 - automating through our build infrastructure

This is currently a longer term, would be nice issue .... :-)

planetf1:
Still hoping to do this

planetf1:
Closing, I don't think we will do this until the point it's a hard requirement.

planetf1:
@tcnt you mentioned you're interested in looking at egeria on pi ... so assigning the issue you're way
Happy to help with the tweaking of ci/cd actions as needed

tcnt:
To build Egeria images for raspberry PI (ARM64 architecture) is no problem.
I can build all Egeria images without code change using
docker buildx build \
--push -t \<docker id\>/\<image name\> \\
--platform linux/amd64,linux/arm64 .
where image name is 

- egeria
- egeria-configure
- jupyter
- egeria-ui
- egeria-react-ui

The challenge is (and for Macbook as well) to get a proper Kafka image as Bitnami which is currently used in the helm-charts on Kubernetes is not supporting ARM64 at the moment as mentioned on https://github.com/bitnami/charts/issues/7305

I have a working version for Kafka 3.0.0 on Kubernetes running on Raspberry PI which is based on https://hub.docker.com/r/moukoublen/kafka. 
But this is not an offical supported image, and also meant to be used on docker directly (not Kubernetes).
I created my own preliminary helm chart to get it working, but need to revise a bit more so scaling is properly supported.
Also I can test using the KRaft option in Kafka so zookeeper is no longer needed.
But having said that: Kafka is done on my personal Docker Id.

So creating the egeria images is not a problem. We could add the ARM64 architecture to the build.
But I assume we do not support Kafka by building our own KAfka images inside the Egeria project.
[...]

PR #5947, opened 2021-11-25
Add support for ARM64 Image Build

tcnt:
Signed-off-by: Theo Thomas <tcn.thomas@googlemail.com>

# Description

Removed building images using Maven.
Building Images is now done using Docker directly in the workflow.
Using docker directly enables us to build for multiple architectures.
This is the first step related to https://github.com/odpi/egeria/issues/2391  to enable the use of ARM64 images

Fixes # (issue)

# How Has This Been Tested?

Building the images for AMD64/ARM64 (Raspberry PI) on my own Docker repo.
Tested lab exercises on Kubernetes Cluster on Raspberry PI using images from my own Docker Repo.

### DIFF
--- a/.azure-pipelines/nexus-iq-clm.yml
+++ b/.azure-pipelines/nexus-iq-clm.yml
@@ -9,19 +9,9 @@ trigger:
     include:
       - master
 
-schedules:
-  - cron: "0 11 * * 6"
-    displayName: Weekly CLM Scan
-    always: true
-    branches:
-      include:
-        - master
-
 variables:
   - group: Artifactory
   - group: NexusIQ
-  - name: CLM_VERSION
-    value: '2.8.1-01'
 
 stages:
   - stage: CLM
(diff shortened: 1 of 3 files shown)


tcnt:
@planetf1 
In egeria-ui and egeria-react-ui I see that for release only version is used for tag and for master/main latest and version is used in the tag.

In the release workflow for egeria  always the latest and version is used as tag.
As far as I could see this is similar to what we used when using maven for docker build.
Is this done on purpose in the release for egeria or should it be aligned with the other 2 repos?

planetf1:
> @planetf1 In egeria-ui and egeria-react-ui I see that for release only version is used for tag and for master/main latest and version is used in the tag.

egeria-ui doesn't have a release workflow currently, so the merge build uses ver+latest for main, and ver for !main
The same applies to egeria-react-ui


> In the release workflow for egeria always the latest and version is used as tag. As far as I could see this is similar to what we used when using maven for docker build. Is this done on purpose in the release for egeria or should it be aligned with the other 2 repos?

Egeria has a specific release workflow -- so the decision I took there was the build the docker image as part of that process.
Therefore the merge build only builds for main

Ideally we'd push pre-release images to a staging area, or with an alternate tag automatically to facilitate testing - but that requires more steps.

I would go with the same approach to tags for this PR on the architecture -- but we can certainly discuss improvements to the overall release process in another issue or PR as needed?

planetf1:
Thanks for proposing the change!

planetf1 (approved):
Reviewed and approved

These changes look good for adding the arm v8 architecture alongside intel/amd 64. 
This also simplifies the build by no longer being dependent on the spotify docker plugin

In a future PR we can provide some docs on rebuilding the docker image to help others incorporate into their CI/CD process, but it makes sense to validate the process first.

Inspected, conditions look good, but will merge and will check build output

PR #6026, opened 2021-12-09
[server-chassis-spring] Activate startup servers after application start

bogdan-sava:
# Description

Move activation of startup server list from spring context refresh event, after application is starting.
Also make sure that CSV of servers order is preserved for server activation order

Fixes # (issue)

# How Has This Been Tested?

Running server list and test the port application is open when servers begin activate.

Removed unnecessary toString calls


lpalashevski (approved):

lpalashevski:
nice I like the addition that the server start up sequence order from the configuration is preserved.

PR #6041, opened 2021-12-16
Asset Catalog - search types name restriction and validation

### LINKED ISSUE #6004: [Enhancement] Incorrect types supplied on a asset catalog omas search result in internal errors on the server

### Current Behavior

If you issue `../servers/mds1/open-metadata/access-services/asset-catalog/users/Fred/search?searchCriteria=abc`
with a body of 
{
	"class": "SearchParameters",
	"status": "ACTIVE",
	"limit": 0,
	"pageSize":100,
    "entityTypes": ["aaa"]
}

it fails with an internal server error. 
it fails for "entityTypes": ["GlosSaryTerm"] but works "entityTypes": ["GlossaryTerm"]

the type name need to be correct including the case.

### Expected Behavior

the java client and server should match the types without taking the case into account. If the type is not known then an appropriate error should be returned to the call saying something like "A catalog asset search has been issued with invalid  entityTypes {0}" user response " please supply valid types. You can find a list of valid types by ....".

### Alternatives

document the current behaviour without changing the code

### Any Further Information?

_No response_

- [X] I can work on this

alexandra-bucur:
# Description
There was a NPE hidden under and Internal Server Error that appeared with non existing entity types that is fixed in this  PR. Also, there is an exception thrown when an entity type in request list does not exist, even when it's something related to upper/lower case. I added this detail to the user documentation.

Fixes #6004 

# How Has This Been Tested?
Manually

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/factory/GraphFactory.java
+++ b/open-metadata-implementation/adapters/open-connectors/governance-daemon-connectors/open-lineage-connectors/open-lineage-janus-connector/src/main/java/org/odpi/openmetadata/openconnectors/governancedaemonconnectors/openlineageconnectors/janusconnector/factory/GraphFactory.java
@@ -15,7 +15,6 @@ import org.janusgraph.core.JanusGraph;
 import org.janusgraph.core.JanusGraphFactory;
 import org.janusgraph.core.schema.JanusGraphManagement;
 import org.odpi.openmetadata.frameworks.auditlog.AuditLog;
-import org.odpi.openmetadata.frameworks.connectors.properties.ConnectionProperties;
 import org.odpi.openmetadata.governanceservers.openlineage.ffdc.OpenLineageException;
 import org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.graph.LineageGraphConnectorProvider;
 import org.odpi.openmetadata.openconnectors.governancedaemonconnectors.openlineageconnectors.janusconnector.graph.LineageGraphRemoteConnectorProvider;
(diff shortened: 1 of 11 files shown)


open-metadata-implementation/access-services/asset-catalog/asset-catalog-api/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/exception/AssetCatalogErrorCode.java
```diff
@@ -49,12 +49,17 @@
     LINKING_RELATIONSHIPS_NOT_FOUND(404, "OMAS-ASSET-CATALOG-404-015 ",
             "There is no intermediate relationships that connect the {0} with the {1} in OMAS Server {2}",
             "The system is unable to retrieve the linking relationship.",
-            "Check that the unique identifiers of the assets are correct."),
+            Constants.CORRECT_UNIQUE_IDENTIFIER),
```
davidradl, line 52: I do no think we should use constants in our messages. These messages should be localisable and a constant makes that very difficult to do.

open-metadata-implementation/access-services/asset-catalog/asset-catalog-client/docs/user/asset-catalog-omas-client.md
```diff
@@ -48,7 +48,9 @@ Once you have an instance of the client, you can use it to:
 
 * Returns the sub-graph that represents the list of assets that in neighborhood of the given asset - [getAssetsFromNeighborhood]()
 
-* Return a list of assets matching the search criteria without the full context [searchByType]()
+* Return a list of assets matching the search criteria without the full context [searchByType](). Entity types are case 
+insensitive, so proper values should be given in the request, otherwise an EntityNotKnownException will be thrown. If
```
davidradl, line 52: how does the user know what a proper value is? Can we add checking client side to as well to reject the request before it reaches the server. Maybe the client gets the supported types and then checks against those.

alexandra-bucur: I added details to the user documentation about the way to find out the supported types.

davidradl: @alexandra-bucur it would be better if the client could issue a call to the omas to get a list of supported types, the server side OMAS can expose any new types in a controlled way.  I recommend that this list also include any child types. For example if you support Topic you should also support KafkaTopic and any custom child types that have been added. I suggest this is not an enumeration as it could change in time; it should probably return a sanitised OMAS friendly version of the typedefs in question. The caller can then be 'metadata-driven' and use the attribute definitions of each type to inform its processing.

alexandra-bucur: @davidradl I though of this one to be the user documentation. Do you think that sentences added in this file are not enough?

lcpopa (approved):

davidradl (approved):

PR #6042, opened 2021-12-16
Update the commit process for janusgraph connector

pmarius22:

# Description

Reduced the number of transactions that occur while saving data for OLS and generated new traversals for each call. 

# How Has This Been Tested?

Tested on local environment


lpalashevski (approved):
Approving the current changes with minor comment. 

I also did broader review of the state of the connector and captured further improvement ideas in #6053. Lets discuss and these separately.

Issue #6044, opened 2021-12-17
[BUG] getNestedAttribute is not performing correctly


davidradl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

I have a tree of nested attributes. If I issue the call getNested attributes, I get back all of the schema attributes that are connected to the supplied guid, including the parent. 

### Expected Behavior

I have a tree of nested attributes. If I issue the call getNested attributes, I get back all of the schema attributes that are connected to the supplied guid, excluding the parent. 

### Steps To Reproduce

set up a Topic, EventType and some nested schema attributes. got example nested attributed  A with child B with child C. Issue getNestedAttribute on B, and you expect to see only C in the response not A and C.  

### Environment

```markdown
- Egeria:
- OS:
- Java:
- Browser (for UI issues):
- Additional connectors and integration:
```

### Any Further Information?

In SchemaAttributeHandler method getNestedSchemaAttributes it calls this.getAttachedEntities. There appears to be no relationship end checking.  It should only return children, i.e. connected schema attributes to the nestedAttributes and not the parentAttribute end.

mandy-chessell:
I have extended the SchemaAttributeHandler to pass the end identifier of the relationship that it should navigate.  This means that only children are returned.

PR #6046, opened 2021-12-20
DE OMAS - add capability for upserting Topics with EventTypes and EventSchemaAttributes

popa-raluca:
# Description

* New endpoints added for upserting the Topic structure: Topic, EventTypeList, EventType, EventSchemaAttribute
* New endpoints added for deleting the Topic structure
* Updated REST & event clients with corresponding methods
* Update FVTs for the new endpoints

Fixes # (issue)

# How Has This Been Tested?

Tested locally with updated Postman collection for upserting assets & unit tests & FVTs


davidradl:
Hi @popa-raluca @lpalashevski , I have just finished working on integration connectors that do this sort of work - updating and creating topics, EventTypes and schema attributes. The APIs are already present in Data Manager OMAS to do all of this. Note that there is an issue in the generic handler around accessing nested schema Attributes https://github.com/odpi/egeria/issues/6044 . @mandy-chessell  is currently working on a fix for this.

lpalashevski:
> Hi @popa-raluca @lpalashevski , I have just finished working on integration connectors that do this sort of work - updating and creating topics, EventTypes and schema attributes. The APIs are already present in Data Manager OMAS to do all of this. Note that there is an issue in the generic handler around accessing nested schema Attributes #6044 . @mandy-chessell is currently working on a fix for this.

Thanks for the info @davidradl  

Although this looks like duplicate work, it is for a reason. We want to have API interface that is more specific and tailored for Data Engine use cases. This gives us opportunity to abstract (or automate) part of the model during assets creation (for example creating data file by specifying basic properties and column names). 
In practice, there are cases where keeping the "public" interface model simple and scoped to the domain of the OMAS encourages adoption and usage.  FYI we already do the same for files and databases although the Data Manger was the first to support these asset types. Adding topics is just logical continuation of the work. 

The above applies for the simpler caller integration scenarios we are trying to support for in-house build data platforms. For more automated integration scenarios we are still going to look at the dedicated integration daemons using Data Manager and re-use functionality.

lpalashevski (approved):

PR #6052, opened 2021-12-23
GE OMAS client - Expose linkElementsAsPeerDuplicates from ReferenceableHandler

alexandra-bucur:
# Description
Added new endpoint to the Governance Engine OMAS, including client changes, to expose the functionality of marking 2 entities as duplicates using the Peer Duplicate Link relationship and the Known Duplicate classification in a single call.
# How Has This Been Tested?
Manually using the POC we have for Deduplication Governance Action Connector.


open-metadata-implementation/access-services/governance-engine/governance-engine-api/src/main/java/org/odpi/openmetadata/accessservices/governanceengine/api/MetadataElementInterface.java
```diff
@@ -493,6 +493,25 @@ String createRelatedElementsInStore(String            userId,
...
+     *
+     * @param userId caller's userId
+     * @param metadataElement1GUID unique identifier of the metadata element at end 1 of the relationship
+     * @param metadataElement2GUID unique identifier of the metadata element at end 2 of the relationship
+     * @throws InvalidParameterException the unique identifier's of the metadata elements are null or invalid in some way; the properties are
+     *                                    not valid for this type of relationship
+     * @throws UserNotAuthorizedException the governance action service is not authorized to create this type of relationship
+     * @throws PropertyServerException there is a problem with the metadata store
+     */
+    void linkElementsAsPeerDuplicates(String userId,
```
mandy-chessell, line 508: Why are the properties (eg steward) not included in the request?

mandy-chessell (changes requested):
The server-side implementation is very elegant.  I like the way inheritance is used on the handler.  A couple of comments:
* There are properties on the duplicate relationship that it would be useful to be able to set up from the governance action.
* There also need to be a change to the RemediationGovernanceContext (in the GAF) to add the new method in order to make it visible to a remediation governance action service.

open-metadata-implementation/access-services/governance-engine/governance-engine-api/src/main/java/org/odpi/openmetadata/accessservices/governanceengine/api/MetadataElementInterface.java
alexandra-bucur: @mandy-chessell Does it look like you thought about the changes? :)

popa-raluca (approved):

Issue #6111, opened 2022-01-05
Fix multiple SLF4J bindings in gradle build


planetf1:
When running from the gradle build, we find some slf4j duplicate bindings which should be corrected
These issues do not occur from the maven build
Some corrections over scoping are likely needed

```
jonesn:egeria/ (master) $ egeria start                                                                                             [10:43:46]
Starting multiple egeria platforms in bg
==> Successfully started `zookeeper` (label: homebrew.mxcl.zookeeper)
==> Successfully started `kafka` (label: homebrew.mxcl.kafka)
no main manifest attribute, in ./open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/user-interface/ui-chassis-spring-3.5-SNAPSHOT.jar
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/jonesn/IdeaProjects/egeria.gradle/open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/server/lib/graph-repository-connector-jar-with-dependencies-3.5-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/jonesn/IdeaProjects/egeria.gradle/open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/server/server-chassis-spring-3.5-SNAPSHOT.jar!/BOOT-INF/lib/logback-classic-1.2.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/jonesn/IdeaProjects/egeria.gradle/open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/server/lib/graph-repository-connector-jar-with-dependencies-3.5-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
[...]

PR #6116, opened 2022-01-09
Add Archive Manager OMES

mandy-chessell:
# Description

This pull request is primarily adding support for the Archive Manager OMES which supports a new type of engine called an Archive Engine.  Within the Archive Engine are Archive Services and they run in the engine host.  They are responsible for the dynamic maintenance of an open metadata archive.  Examples of its use include:

* Running a dynamic backup of all changes to one or more open metadata respository.
* Maintaining a content pack of say a glossary that captures all edits ot the glossary as they are made through the OMASs.

This new service required a number of related changes:
* New open metadata types for the archive engine/service.
* Support for classification events that include an entity proxy rather than an entity detail.  This is because classifications are included in an open metadata archive with an entity proxy.
* Enablement of an optional remote Enterprise OMRS Topic to allow the Archive Services to receive OMRS topic events for the connected repositories.  This added changes to the Admin configuration services and the OMRS enterprise repository services.
* Renaming of Partner OMAS to Partner Service in the engine services - this is because the partner service for the Archive Manager OMES is OMRS rather than an OMAS.
* New open metadata archive store based around a directory where each instance is in its own file.  This is to allow for very large archives that do not fit in memory.  To make this possible, the archive builder function needs to be suppported by the connector rather than using the one supplied by OMRS that builds the archive in memory.  The result is a new builder interface in OMRS.  I also added a common header for OMRS types and instances to make it easier for common functions in the archive management.

There are minor fixes to javadoc, code formatting and replacement of Throwable to Exception that I came across.

# How Has This Been Tested?

At this stage, the focus has been on adding the new service and validating there are no regressions to most particularly the OMRS.  Therefore the FVT and the open metadata labs have been run.

Testing of the new function will occur as examples of the new archive services are written.

### DIFF
--- a/.github/workflows/pr-gradle.yml
+++ b/.github/workflows/pr-gradle.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     name: "Gradle"
     steps:
       - uses: actions/checkout@v2
@@ -24,7 +24,8 @@ jobs:
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
           restore-keys: ${{ runner.os }}-gradle
       - name: Build with Gradle
-        run: ./gradlew build
+        # For a build machine environment, force a more clean build 
+        run: ./gradlew clean build --no-build-cache --no-configure-on-demand --priority normal --no-watch-fs --refresh-dependencies
       - name: Upload Test coverage report
         uses: actions/upload-artifact@v2
         with:
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,8 +7,8 @@ org.gradle.parallel=true
 # Caching
 org.gradle.caching=false
 
-# Watch fileystem - disable if memory constrained
-org.gradle.vfs.watch=true
+# Watch fileystem - disable - too many files swamps OS
+org.gradle.vfs.watch=false
 
 # Default logging output
 org.gradle.console=auto


planetf1:
Having done some further tests in #6122, the only reproduced failure has been by running the build on a clean system

* Installed Ubuntu 20.04 (same as official build?)
* Installed temurin 17 (java 17 from adoptium) [note: official build is running java 11 for gradle]
* Ran a clean clone (albeit pointing to my own fork)
* Ran `./gradlew build` (running the wrapper ensures we are all running the same gradle version)

The same failure then occurs as we see here

```
AnalyticsModelingOMASSynchronizationIT > testSynchronization(String) > org.odpi.openmetadata.accessservices.analyticsmodeling.fvt.execution.AnalyticsModelingOMASSynchronizationIT.testSynchronization(String)[2] FAILED
    org.opentest4j.AssertionFailedError at AnalyticsModelingOMASSynchronizationIT.java:33

8 tests completed, 1 failed

> Task :open-metadata-test:open-metadata-fvt:access-services-fvt:analytics-modeling-fvt:test FAILED
```

It continues to fail on a rebuild, and after the same environment has run a maven (3.8.4) build

A system, previously working, even after running a clean clone, `./gradlew clean`, `git clean -f -d`, `rm -fr ~/.m2` & manually inspecting/checking that the 'build' directories under assemblies & the respective fvts were empty, still continues to work

planetf1:
The stderr from the server is:
```
jonesn@ubuntu-buildtest:~/src/egeria/open-metadata-test/open-metadata-fvt/access-services-fvt/analytics-modeling-fvt/build$ cat chassis-error.log
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/home/jonesn/src/egeria/open-metadata-test/open-metadata-fvt/access-services-fvt/analytics-modeling-fvt/build/../../../../../open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/server/lib/graph-repository-connector-jar-with-dependencies-3.5-SNAPSHOT.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/home/jonesn/src/egeria/open-metadata-distribution/open-metadata-assemblies/build/unpacked/egeria-3.5-SNAPSHOT-distribution.tar.gz/server/server-chassis-spring-3.5-SNAPSHOT.jar!/BOOT-INF/lib/logback-classic-1.2.10.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```
This is known (for gradle) and tracked in #6111

Also the same error is seen in a 'working' gradle build without this PR

There are no abnormal messages from the stdout (chassis.log)

planetf1:
Surprisingly new gradle builds are also now failing (and yet must have worked to get through the PR)
See https://github.com/odpi/egeria/pull/6125 for example .....

PR #6129, opened 2022-01-18
add logback-test.xml to asset-consumer-fvt

bogdan-sava:

# Description

added logback-test.xml in order to debug the fvt fails


planetf1 (commented):
Was this added to try and debug https://github.com/odpi/egeria/issues/6130 (more is explained there)?
As far as I can see the asset consumer FVT is already logging INFO level messages?

Issue #6137, opened 2022-01-21
Enhancement - add quick build option 


planetf1:
Sometimes it is desirable to do a quick build
 - in a quick dev/test loop
 - for a demo
 - for a tutorial
 
 In these cases various flags can be set to make the build super quick. For example I use a progressive set of aliases to gradually go from slowest/best checks to fastest/least checks
 
I propose this is added as a profile to the maven build. Initially this won't be documented except in tutorial due to the risk of a bad environment, missing dependency checks etc -- though the PR process will always check these

Gradle is already much quicker, though needs clarity around skipping tests

- [x] Maven : Add quick profile
- [x] Maven : Document quick profile 
- [x] Gradle : Add quick(er) option for build
- [ ] Gradle : Document quick profile

cc: @mandy-chessell

planetf1:
When adding a profile for maven, using it to set various properties is easy -- and this can influence behaviour of our plugs such as skipping javadoc, dependency checks, source build, unit tests etc.

However FVT is harder, since one profile cannot activate/deactivate another, and the properties used to govern activation are evaluated early on, before any other profile can change them. Additionally a profile activation/deactivation cannot be controlled by multiple properties (and/or). Then we observe that our FVTs do not purely use the normal test plugin & structure, but actually have code built as normal source, and make use of other plugins. Nor can modules be deactivated in a profile (though they can be selectively activated). 

-- will try and figure out if this is possible via a single parameter without lots of refactoring ....

planetf1:
Notes on performance

Machine A - RHEL, 16GB RAM, 4 cpu
Machine B - MacOS, 16GB ram, 4 CPU, 2016 mbp

Machine A:
 - Maven 3:24 (with m2 already populated) `mvn clean install -Pquick -DskipFVT`
 - Gradle 1:22 `gradle build -x test -x javadoc`


 Machine B
  - Maven 03:09 `mvn clean install -Pquick -DskipFVT`
  - Gradle 0:57  `gradle build -x test -x javadoc`

 Note on gradle
  - actually ran `gradle clean build -x test --no-build-cache` to avoid my local cache & delete prebuild targets to force full compile (but m2 still used)
  - you wouldn't normally use clean. added just for worse case to force a build
   - selection execution of tests based on code changed still needs adding to benefit incremental use cases -- this timing comparison shows the potential benefits from a properly coded gradle build ie best of both worlds, if we can define rules about under what conditions fvts are required
 
 
[...]

PR #6138, opened 2022-01-21
#6137 add -Pquick option

planetf1:

# Description

Adds `-Pquick` maven build option

Skips many build checks

See #6137 for details

# How Has This Been Tested?

Build

1. Currently you must use this option in conjunction with `-DskipFVT` otherwise FVTs will fail

Still looking at how to resolve this user impact - see #6137 for discussion

Leaving as draft..


planetf1:
I'm inclined to stick with this simple option for the first PR, it at least reduces the number of parks
needed
I'd rather not do further changes just prior to release
Gradle is already quick, and probably where we should focus moving forward
cc: @mandy-chessell

PR #6287, opened 2022-03-01
Remove entity proxy validation on classify/reclassify/declassify 

popa-raluca:
# Description

This PR covers two issues:
1. Ignore exception on classifyEntity when the entity is already classified - to address the race condition that can occur when two different servers are adding the latest change classification at the same time
2.  Remove entity proxy validation on classify/reclassify/declassify methods in LocalOMRSMetadataCollection
The previous validation was retrieving the entityDetail of an entity, that was later used in the validation methods of the securityVerifier.
With these changes the entitySummary of an entity is used instead. I created new methods in the securityVerifier classes, that use EntitySummary as a parameter instead of EntityDetail, but I was wondering if it would be better to modify the existing methods.  @mandy-chessell , any thoughts on this?

Fixes # (issue)

# How Has This Been Tested?

Tested locally with the setup for the deduplication.


lcpopa (approved):

mandy-chessell:
I agree - lets update the original methods in the security connector to pass EntitySummary on the classification methods.

mandy-chessell (changes requested):
All looks good - I agree with suggestiond to update original classification methods in the security connector to pass EntitySummary rather than EntityDetail

PR #6290, opened 2022-03-02
Removes duplicate typedef attributes and adds initial relationship super/subtype support to CTS

### LINKED ISSUE #6211: [BUG] Same attribute at multiple levels of a typedef hierarchy?

### Current Behavior

In 3.5 it looks like we introduced a new `SoftwareCapability` entity with a `name` attribute that inherits from `Referenceable`. We then changed the existing `SoftwareServerCapability` to have this `SoftwareCapability` as its super type. However, the `SoftwareServerCapability` itself also has a `name` attribute. For various repositories (Graph and XT) this introduces ambiguity and could cause edge-case failures on how `name` as a property should be interpreted (searched) and stored (against `SoftwareServer` or `SoftwareServerCapability`). At a minimum it is likely to cause false negative search results in these repositories (where a search should actually find a result it may not).

### Expected Behavior

I vaguely recall we had a similar discussion some time ago during some other typedef update, but couldn't find the detailed issue again (if there was one).

My fundamental question was whether we have a way of removing the `SoftwareServerCapability`'s `name` property, as presumably the `SoftwareServer`'s `name` property should be used unambiguously (?)

### Steps To Reproduce

_No response_

### Environment

_No response_

### Any Further Information?

_No response_

cmgrote:
# Description

Removes duplicate typedef attributes from:

- SoftwareServerCapability
- SoftwareServerSupportedCapability

Fixes #6211 

Adds initial support for relationship super types and subtypes to the CTS (search case only for now).

Partially addresses #6292

# How Has This Been Tested?

- CTS ✅ 
- Full build (including FVTs) ✅ 

Please confirm this is how we treated these overlaps in the past 🙏

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/integration-connectors/files-integration-connectors/src/main/java/org/odpi/openmetadata/adapters/connectors/integration/basicfiles/BasicFilesMonitorIntegrationConnectorBase.java
+++ b/open-metadata-implementation/adapters/open-connectors/integration-connectors/files-integration-connectors/src/main/java/org/odpi/openmetadata/adapters/connectors/integration/basicfiles/BasicFilesMonitorIntegrationConnectorBase.java
@@ -32,7 +32,7 @@ public abstract class BasicFilesMonitorIntegrationConnectorBase extends FilesInt
     String  templateQualifiedName = null;
     boolean allowCatalogDelete    = false;
 
-    private String            fileDirectoryName     = null;
+    private String            fileDirectoryName = null;
     private FileFolderElement dataFolderElement = null;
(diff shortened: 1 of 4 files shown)


cmgrote:
Confirmed that CTS still passes (along with all other build-based tests like FVT). Not sure if there are other tests to run (manually) e.g. for OMASes that might use these objects?

cmgrote:
@planetf1 @mandy-chessell to discuss in tmw's dev call -- think we should cherry-pick these commits over to 3.6 release if we're happy with them, as I'm now able to get the CTS to run through without any failures (and without needing to change the relationship typedef discussed in the release issue).

Issue #6311, opened 2022-03-11
[BUG] Some classifications not returned from findEntityXXX requests


mandy-chessell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The federated findEntityXXX repository requests supported by the enterprise repository services do not return classifications for entities if they are stored in a repository attached to an entity proxy.  These methods do return all classifications attached to home entities and reference copies.

The reason for the difference is that the entity proxy can not be returned on the findEntityXXX methods because they return EntityDetail objects.

The methods that retrieve a single entity (isEntityKnown, getEntitySummary, getEntityDetail) will successfully return all classifications because the executors for these requests are making requests for a specific entity and so receive EntityProxyOnlyException if the entity is known but only a proxy is available.  This then allows getHomeClassifcations() to be called on that repository to retrieve any locally homed classifications from that repository.

### Expected Behavior

All classifications for an entity should be returned from the federated query no matter how they are stored.

The federating query mechanism needs to minimise calls to the cohort member repositories and allow the query to execute in parallel.

### Steps To Reproduce

The entities needs to be stored in a repository that is using the adapter pattern (ie running in a repository proxy) and does not have an event mapper nor supports the Anchors classification.  The lack of an event mapper means that events relating to its metadata will not be sent over the cohort topic(s) and so no reference copies from this repository will be created in the other cohort members.  The lack of support of the Anchors classification means that this classification needs to be stored in another repository - and due to the lack of reference copies, the classification will be attached to an entity proxy.

This repository needs to be part of a cohort with a native repository so there is somewhere to store the Anchors classifications.  

Then issue a query to find these entities through an OMAS to engage the federated query.  Beforethis fix, the anchors classification is missing, after this fix, the Anchors classification is returned.

### Environment

```markdown
- Egeria: 3.7 SNAPSHOT or earlier
```

mandy-chessell:
This fix requires the findEntityXXX executors to operate in two phases.  The first phase to retrieve the entities and the second phase to retrieve any stranded home classifications.

The first phase visits each repository and retrieve none, or a list of entity details.  These are assembled in the accumulator where old/duplicated versions are discarded and the classifications are merged into a common list.
The second phase needs to visit each repository **for each entity returned** by the first phase and retrieve any homed classifications for that entity.  Any retrieved classifications are added to the accumulator.

[...]

Issue #6158, opened 2022-01-28
[BUG] Broken links in swagger docs


planetf1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

The following sections have broken links from generated swagger UI at /swagger-ui.html

* Administration Services - Operational
* Catalog Integrator OMIS
* Conformance Suite
* Dino OMVS
* Glossary View OMAS
* Platform Services
* Search Integrator OMIS

### Expected Behavior

Links should be correct

### Steps To Reproduce

go to swagger endpoint and view

### Environment

```markdown
- Egeria:
- OS:
- Java:
- Browser (for UI issues):
- Additional connectors and integration:
```

mandy-chessell:
@planetf1 Which links are bad? the same list as at the top of this issue?

mandy-chessell:
I had a look at the code and although the URL for `Administration Services - Operational` is coded correctly, the swagger parser is mangling it.  I have changed it to not refer to a section within the operations guide.

planetf1:
Very minor, but we might get a minor improvement with the long labels -> https://github.com/odpi/egeria/issues/6158#issuecomment-1081713323

planetf1:
I checked the swagger docs as part of 3.9 testing and raised more specific issues for any remaining problems

Issue #6155, opened 2022-01-27
[Enhancement] Add module status to each of Egeria's services' component description


mandy-chessell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Please describe the new behavior that that will improve Egeria

Each module in Egeria has a component description that is registered with the audit log so it can be associated with each log record that it produces.  This description includes:

* Component Id - integer
* Component Name
* Component Description
* Component's wiki page URL

Lists of component descriptions are maintained as Enums in the egeria code - for example:
* [Access Services](https://github.com/odpi/egeria/blob/master/open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/registration/AccessServiceDescription.java)
* [OMRS Components and connector types](https://github.com/odpi/egeria/blob/master/open-metadata-implementation/repository-services/repository-services-apis/src/main/java/org/odpi/openmetadata/repositoryservices/auditlog/OMRSAuditingComponent.java)

In addition to registering with the audit log, these component descriptions are used to support platform services that list registered services.  They also appear in the audit log report showing the hierarchy of components running in a server.

This enhancement proposes to add a status code to this description that has one of 4 values:
* In development
* Tech Preview
* Supported
* Deprecated

This then becomes our official status for each component.  It would provide a way to create reports of the status of all of our components for TSC review - or if an organization is going into production and they want to understand the risks in their deployment.

We would still publish the status of a component in the webpage.  IE [https://odpi.github.io/egeria-docs/services/omrs/](https://odpi.github.io/egeria-docs/services/omrs/)

### Alternatives

Today it is necessary to follow the web URL link to the component's page to read its status from the top of the page.  This makes it hard to automate the generation of reports showing the status of components in a particular deployment environment.

### Would you be prepared to be assigned this issue to work on?

- [X] I can work on this

mandy-chessell:
I am thinking that a 5th status of "sample" would be a good addition.  This would distinguish between the supported runtime and a sample.

mandy-chessell:
Discussion on the community call suggested using "Stable" instead of Released/Supported

Issue #6299, opened 2022-03-03
Move graph repository connector to new github repository


planetf1:
Move the graph repo connector to a new repository

Outline of tasks
- [x] #6300

planetf1:
The open lineage connector is also based on Janusgraph.

We should consider whether to 
 - leave as is
 - move to same repo as graph repo connector (as they are all janus based)
 - move to it's own new repo (as it's a very distinct connector)
 
 cc: @mandy-chessell @lpalashevski

planetf1:
Note that this month dependabot recommended gremlin 3.5.2 -> 3.6.0 -- this caused a compile error (iterators). Given the pending move, and testing concerns I've left as-is. I do suggest we move soon - it should be a relatively easy task to move source & perform a build. Obviously anyone integrating will have to pick up from a new location -- and we may need additional work for the openlineage notebook to work (it may be a good time to move the notebooks too).

PR #6316, opened 2022-03-14
Federated query, component status and URLs

mandy-chessell:
# Description

This PR has the following fixes:
* correction of federated query to retrieve stranded home classifications after a search (#6311)
* new ComponentDevelopmentStatus enum assigned to each module (#6155)
* fixes to component description URLs to use new URL
* fixes to swagger URLs (also includes creation of pages on new site for Glossary View OMAs and Search Integrator OMIS) (#6158)
* Movement of connector statistics support from Integration connectors to all types of connectors

# How Has This Been Tested?

FVT and coco labs

### DIFF
--- a/.github/workflows/pr-gradle.yml
+++ b/.github/workflows/pr-gradle.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     name: "Gradle"
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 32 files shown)


planetf1:
The verification failed in FVTs with a similar error to that we've seen before. It has only ever occurred on graph repo, and appears to be an odd issue where an entity is created, but then not there when queried. Previously it looked as it was OS version dependent, but it's not clear that was really the case.

Co-incidentally, the graph has been removed from the fvt under #6300 so a rebuild should work

In #6299 we need to reestablish similar tests for graph connector -- and also crux

PR #6320, opened 2022-03-14
#6307 remove lombok from global scope: only use where required

### LINKED ISSUE #6307: Restrict lombok scope to only those modules where it is required

### Please describe the new behavior that that will improve Egeria

A few modules in Egeria make use of lombok.
We have the lombok plugin enabled globally, across all modules, for both maven and gradle.
'delombok' is explicitly used to enable javadoc generation to work

This causes some confusion - for example 'find in files' ends up finding repeated hits of java source - one for the deblombok'd version, one for the regular. We end up with *ALL* of our java source duplicated, which takes time & space. I also keep hitting https://github.com/odpi/egeria/issues/5570 but can't pin down the cause - reducing the scope should help. 
 
 I propose we look at removing global enablement of lombok, and instead customize only those subprojects requiring to use lombok with the necessary lombok dependencies/plugin & delombok/javadoc steps
 
 This restricts any confusion over use of lombok to those modules where it is actually used, rather than the 100s of modules where it is not. (Currently we have delombok active in 322 modules)
 
 I'd also like to open this issue up for discussion - do you find lombok confusing? How does it affect you? Or do you think we should stick with current global enablement across egeria

### Alternatives

_No response_

### Any Further Information?

See previous discussion in https://github.com/odpi/egeria/issues/5207 where we solved the most crucial issues, and left at global scope.

- [X] I can work on this

planetf1:

# Description

Removes use of the lombok plugin/dependency except in modules where required
Ensures both the lolbok plugin & modified javadoc build are pushed to those modules
maven & gradle

Fixes #6307

# How Has This Been Tested?

Build only


planetf1:
Tested - only delombok directories after execution are:
```
./open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-api/target/delombok
./open-metadata-implementation/access-services/data-engine/data-engine-api/target/delombok
./open-metadata-implementation/access-services/data-engine/data-engine-server/target/delombok
./open-metadata-implementation/access-services/glossary-view/glossary-view-api/target/delombok
./open-metadata-implementation/access-services/asset-catalog/asset-catalog-api/target/delombok
./open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/target/delombok
./open-metadata-implementation/access-services/asset-lineage/asset-lineage-api/target/delombok
./open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/target/delombok
```
(maven)

PR #6324, opened 2022-03-16
Ensure federated query (#6311)

mandy-chessell:
Signed-off-by: Mandy Chessell <mandy.e.chessell@gmail.com>

# Description

1. Change the find executors so they always return false - this causes them to traverse all repositories
2. Update integration connectors to supply a component description for the audit log
3. Change the admin services ServerStatus to ServerServicesStatus to avoid name clash with platform services ServerStatus
4. Update the methods in the SoftwareServerPlatform beans from setSoftwareServerXXX to setSoftwareServerPlatformXXX

Fixes # (issue)

# How Has This Been Tested?

FVT - hoping that @alexandra-bucur will test with config that does not emit events from one of the repositories.

### DIFF
--- a/open-metadata-implementation/repository-services/repository-services-implementation/src/main/java/org/odpi/openmetadata/repositoryservices/enterprise/repositoryconnector/executors/GetAllTypesExecutor.java
+++ b/open-metadata-implementation/repository-services/repository-services-implementation/src/main/java/org/odpi/openmetadata/repositoryservices/enterprise/repositoryconnector/executors/GetAllTypesExecutor.java
@@ -99,7 +99,7 @@ public class GetAllTypesExecutor extends CloneableRepositoryExecutorBase
             accumulator.captureGenericException(methodName, metadataCollectionId, error);
         }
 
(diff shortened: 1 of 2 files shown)


planetf1:
@mandy-chessell The underlying issue has been fixed & tested in master now, so you should be able to merge with master & the checks should work ok. (see https://github.com/odpi/egeria/issues/6326 )

alexandra-bucur:
@mandy-chessell I will test it as soon as it gets in master. I cannot pull code from your fork.

alexandra-bucur:
I tested the search and it works fine, but the publishing of the glossary term through Asset Lineage is not working.  It relies on the GetRelationshipsForEntityExecutor which was not changed. Along with it I saw that there are another 11 executors without this new logic, but I am not aware of the impact of their behaviour.

mandy-chessell:
I have updated getRelationshipsForEntity and getAllTypeDefs

alexandra-bucur:
It works fine now. Thank you!

PR #6345, opened 2022-03-27
Egeria code

mandy-chessell:
# Description

This pull request focuses on the event management for the following OMASs:

* Asset Consumer
* Asset Manager
* Asset Owner
* Community Profile
* Data Manager
* Digital Architecture
* IT Infrastructure
* Security Manager
* Stewardship Action

Asset Owner, Digital Architecture and Stewardship Action did not have any event management and now do.

The rest were not supporting the new classification events that use entity proxies which was the primary reason for the change.  Also most did not support events for rehominng, reidentifying and retyping nor refresh.

The OMASs all now use the generic handlers to verify if the event should be sent - this includes the security call and the asset zone check.  (Before many had their own implementation of the zone chack and omitted the security check.

The types of entities and relationships that Data Manager sends events for has been expanded from files and databases to include APIs.

Asset Manager now sends events for relationship changes in addition to its previous support for entities and classifications.

Fixes # (issue)

# How Has This Been Tested?

Using the Coco Labs


mandy-chessell:
THere is a new npa account suppored by the security connectors called generalnpa.  This is for use by the dojos and other demos.

PR #6412, opened 2022-04-01
remove jakarta annotation api dependency

bogdan-sava:

# Description

The dependency is used only for @PostConstruct annotation, while we are using initialisation bean anyway.


planetf1:
Good news if we can remove it. Better than updating

PR #6431, opened 2022-04-19
Design model and connector archives

mandy-chessell:
# Description

This PR includes
* tests and fixes for the Discovery Engine OMAS and related function.  Thisi s still work in progress.
* the inclusion of open metadata archives from the samples repository into the distribution so they can be used in the open metadata labs
* updates to the open connector archive write to use the new helpers.

Fixes # (issue)

# How Has This Been Tested?

Running the archives and FVT

The aim was to move all of the archive writers to use the new helper utilities.   This was coded but the design model arcives were rolled back to their original content while we decide what to do with the CloudInformationModel archive (this project is no longer active).


mandy-chessell:
This PR also includes the removal of the Cloud Information Model Archive Writer.

PR #6445, opened 2022-04-26
Increase gradle daemon memory for build from 4GB to 5GB to address oc…

planetf1:
…casional daemon dying - probable cause

## Related Issue(s)

Increases memory for gradle daemon from 4GB to 5GB. Seen some occasional failures in the FVTs where the daemon exits which I think is caused by being out of space.

## Testing

Observing build failures

No

## Additional notes

No


planetf1:
Build fails in same way after this change so looks as if cause is something else. closing this PR for now

PR #6447, opened 2022-04-26
Created endpoints to retrieve assets by type GUID or Name

MihaiIliescu:
Signed-off-by: Cristian-Mihai.Iliescu <cristian-mihai.iliescu@ing.com>

## Description

Created endpoints to retrieve assets by type GUID or Name

## Testing

Local tests done.

### DIFF
--- a/.github/workflows/merge-gradle.yml
+++ b/.github/workflows/merge-gradle.yml
@@ -24,7 +24,7 @@ jobs:
         name: Checkout source
         if: ${{ github.repository == 'odpi/egeria' && github.ref == 'refs/heads/master'}}
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           java-version: '11'
           distribution: 'temurin'
@@ -37,7 +37,7 @@ jobs:
         run: './gradlew aggregateJavadoc'
       - name: publish
         if: ${{ github.repository == 'odpi/egeria' && github.ref == 'refs/heads/master'}}
-        uses: JamesIves/github-pages-deploy-action@v4.2.5
+        uses: JamesIves/github-pages-deploy-action@v4.3.3
         with:
           branch: gh-pages
           folder: build/docs/javadoc
--- a/.github/workflows/merge-maven.yml
+++ b/.github/workflows/merge-maven.yml
@@ -23,7 +23,7 @@ jobs:
         # Only for a merge
         if: ${{ github.event_name == 'push' && github.repository == 'odpi/egeria' }}
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           # Java 11 is used for merge builds (PRs do check Java latest)
           java-version: '11'
@@ -112,7 +112,7 @@ jobs:
       # --
       # Mostly for verification - not published to the release itself for now
       - name: Upload assemblies
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: Assemblies
           path: open-metadata-distribution/open-metadata-assemblies/target/*.gz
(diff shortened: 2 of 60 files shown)


open-metadata-implementation/access-services/asset-catalog/asset-catalog-spring/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/server/spring/AssetCatalogEntityResource.java
```diff
@@ -136,6 +136,39 @@ public AssetListResponse searchByType(@PathVariable("serverName") String serverN
...
+     * @param typeGUID         the assets type GUID to search for
+     * @return                 list of assets by type GUID
+     */
+    @GetMapping(path = "/assets-by-type-guid/{typeGUID}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+
+    public AssetListResponse getAssetsByTypeGUID(@PathVariable("serverName") String serverName,
+                                                 @PathVariable("userId") String userId,
+                                                 @PathVariable("typeGUID") @NotBlank String typeGUID) {
+        return assetService.searchByTypeNameOrGUID(serverName, userId, null, typeGUID);
```
davidradl, line 169: I wonder why this method is called a search as there is no search criteria.
I wonder why searchByTypeNameOrGUID - why not getByIdentifier

open-metadata-implementation/access-services/asset-catalog/asset-catalog-spring/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/server/spring/AssetCatalogEntityResource.java
```diff
@@ -136,6 +136,39 @@ public AssetListResponse searchByType(@PathVariable("serverName") String serverN
...
+    @GetMapping(path = "/assets-by-type-name/{typeName}",
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    public AssetListResponse getAssetsByTypeName(@PathVariable("serverName") String serverName,
+                                                 @PathVariable("userId") String userId,
+                                                 @PathVariable("typeName") @NotBlank String typeName) {
+        return assetService.searchByTypeNameOrGUID(serverName, userId, typeName, null);
+    }
+
+    /**
+     * Return a list of assets by asset type GUID without any additional search criteria
```
davidradl, line 156: I think that we need to be explicit in the comments here as to why we are expecting a list of assets when we are looking for a type by name or guid, when there is only one. I assume the reason for this is that it includes subtypes , if so we should say that in the comments / javadoc

open-metadata-implementation/access-services/asset-catalog/asset-catalog-client/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/AssetCatalog.java
```diff
@@ -172,6 +174,34 @@ public AssetListResponse searchByType(String userId,
         return assetResponse;
     }
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public AssetListResponse searchByTypeNameOrGUID(String userId,
```
davidradl, line 181: I do not think is great to have a method 'something1 or something2' then working out which one it is is by a null parameter.  I think it would be much cleaner to have separate methods one for get by type name and one for get by type guid. and call common code in the methods.

open-metadata-implementation/access-services/asset-catalog/asset-catalog-spring/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/server/spring/AssetCatalogEntityResource.java
MihaiIliescu: I'll include subtypes in the description

bogdan-sava (approved):

Issue #5290, opened 2021-06-01
Save additional build artifacts to improve ability to debug build


planetf1:
A few failures in PR/merge builds could be more quickly resolved if some additional files were saved
 - RAT output files ie /home/runner/work/egeria/egeria/open-metadata-resources/open-metadata-deployment/target/rat.txt (maven) - example at https://github.com/odpi/egeria/runs/2712634195?check_suite_focus=true
  - Log files from attempts to launch FVT (maven & gradle) - example at https://github.com/odpi/egeria/actions/runs/893883662
  
Will review other files to archive ...

planetf1:
Closing as not planned

PR #6448, opened 2022-04-27
Git6304

davidradl:


planetf1:
The gradle build is failing. First indication is `The error message was I/O error on GET request for "https://localhost:10448/servers/serverinmem/open-metadata/access-services/subject-area/users/garygeeke/categories": Remote host terminated the handshake; nested exception is javax.net.ssl.SSLHandshakeException: Remote host terminated the handshake',` which suggests that the server chassis has exited. Subsequently there is an indication the gradle background daemon has disappeared, but this could be consequential.

Unfortunately the logs are not kept (see #5290). It could have been due to a server exception in egeria, or possibly a heap issue (github workers are limited to 7GB max, and so default for running the jar is probably 3.5GB). If the issue only occurs on the build infra the latter is more likely.

planetf1:
See #6450

Issue #6130, opened 2022-01-18
Gradle PR build failing


planetf1:
The gradle PR build is failing with errors such as:
```
AssetConsumerOMASCreateTagIT > testCreateTag(String) > org.odpi.openmetadata.accessservices.assetconsumer.fvt.execution.AssetConsumerOMASCreateTagIT.testCreateTag(String)[1] STANDARD_OUT
7785
    AssetGUID: 4358845d-1017-495c-b944-d9b8a33604da
7786
    SchemaTypeGUID: acd45be1-5e35-4415-92bc-82a00567b89b
7787
    Asset2GUID: 8d9c1bde-c967-4743-b6d0-063aacd2b2a4
7788

7789
    11:11:01.864 [Test worker] ERROR o.o.o.c.ffdc.RESTExceptionHandler - FFDC Response: TagResponse{tag=null, exceptionClassName='org.odpi.openmetadata.frameworks.connectors.ffdc.InvalidParameterException', exceptionCausedBy='null', actionDescription='getTag', relatedHTTPCode=400, exceptionErrorMessage='OMAG-COMMON-400-021 The identifier 87b16829-7bad-4181-8ab5-36df81c54703 of the InformalTag passed by newUserId to method getTag of service Asset Consumer OMAS is not recognized by the serverinmem server', exceptionErrorMessageId='OMAG-COMMON-400-021', exceptionErrorMessageParameters=[87b16829-7bad-4181-8ab5-36df81c54703, InformalTag, newUserId, getTag, Asset Consumer OMAS, serverinmem], exceptionSystemAction='The system is unable to process the request because the identifier is invalid.', exceptionUserAction='Update the caller to provide a correct asset identifier.', exceptionProperties={parameterName=guid}}
7790
    11:11:01.883 [Test worker] ERROR o.o.o.c.ffdc.RESTExceptionHandler - Detected Invalid Parameter Exception in REST Response
7791
    org.odpi.openmetadata.frameworks.connectors.ffdc.InvalidParameterException: OMAG-COMMON-400-021 The identifier 87b16829-7bad-4181-8ab5-36df81c54703 of the InformalTag passed by newUserId to method getTag of service Asset Consumer OMAS is not recognized by the serverinmem server
```

or 
```
[...]

PR #6523, opened 2022-05-10
Added possibility to see end2end lineage for processes.

MihaiIliescu:
Signed-off-by: Cristian-Mihai.Iliescu <cristian-mihai.iliescu@ing.com>

## Description

Added possibility to see end2end lineage for processes.
Added test for it.

## Testing

Tested with local lineage graph


planetf1:
I don't see any particular reason why the gradle FVT failed. We have intermittent failures - see #6130 . Something timing/sequence related perhaps as we're failing to find an entity. Rerunning the verification here

bogdan-sava (approved):

PR #6606, opened 2022-06-03
Avoid overriding entity details when saving classifications received on entity proxy

alexandra-bucur:

## Description
Environment: 2 platforms, one with Data Engine, one with Governance Engine. Context: Data Engine creates an entity, Governance Engine deduplicates it, the entity gets classified. Classifying it and sending the event on the cohort topic with an entity proxy overrode the entity in both graphs. The entity was not found in enterprise search anymore. Search entities are now available through search in both graphs.

## Testing

Manual with above described scenario.


mandy-chessell (approved):
Looks good

PR #6608, opened 2022-06-06
git6607 all event mapper to be associated with a native repository

davidradl:
Signed-off-by: David Radley <dgradley@gmail.com>


mandy-chessell (changes requested):
This fix is invalid on multiple levels.  The event mapper is only valid in a repository proxy.  Removing the validation that ensures this is the case means that an administrator may configure an event mapper in a native repository and the native event support expected in a Metadata Access Store is not executed.  Similarly, changing the method name is misleading.

Note - the local repository mode of OPEN_METADATA_NATIVE means the the local repository is for a Metadata Access Store.  It is therefore also required to be a native repository implementation too.  It does not mean that the repository implementation happens to be a native implementation that is running in an arbitrary OMAG server type.

PR #6626, opened 2022-06-16
odpi/egeria-charts#155 pin jupyter image version

planetf1:

## Description

Pins Jupyter lab to the version we used for release 3.9

See related issue -- it's possible some recent changes in Jupyter are breaking the lab environment (ssl/certs). This may not be the other issue, but it's prudent to pin the version to that tested - at least until we make a long term improvement (see end)

## Related Issue(s)

See odpi/egeria-jupyter-notebooks#3

## Testing

Manually configured prior version for testing

none

## Additional notes

Long term we should move to using a regular jupyter image, rather than extending - if we can.
This makes it easier to change the version being used at deployment time


planetf1:
Retested after merge: This back-leveling of the Jupyter container addresses the breakage seen with ssl/certs (python 3.10). 
Will revisit versions after refactoring the notebooks/containers for jupyter

PR #6633, opened 2022-06-27
Sonar scan fixes sniff test

wbittles:
## Description

This PR fixes the 
2 high level "Security Hotspot" alerts   
4 low level security hotspots
30+ of "incidental alerts", these may get lost in the 20K "code smells"

## Testing

automated tests only
N/A

## Additional notes

this PR is just a sniff to test to try and assess
The scale of the CodeQL issues
Check the reliability of the cloud scan as a metric on progress made

@mandy-chessell  I know the indentation is wrong on your files, but I can only see the difference in white space when the code is pushed remotely.
If I can't figure out how to fix it I'll pull those changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,7 @@ plugins {
 allprojects {
 
     group = 'org.odpi.egeria'
-    version = '3.10-SNAPSHOT'
+    version = '3.11-SNAPSHOT'
 
     // Mostly java, so default to this for now
     apply plugin: 'java'
--- a/developer-resources/License-Example-Files/License_for_POM_Files.xml
+++ b/developer-resources/License-Example-Files/License_for_POM_Files.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Add module name here</name>
--- a/developer-resources/Release-Process.md
+++ b/developer-resources/Release-Process.md
@@ -64,7 +64,7 @@ are GPG signed and distributed to [Maven Central](https://oss.sonatype.org).
 #### 4. Update master version & cleanup notes
 * `git checkout master`
 * `git pull upstream master`
-* Edit all files (command line or IDE) to replace 'x.y-SNAPSHOT' with the next version, ie change '1.3-SNAPSHOT' to '3.10-SNAPSHOT'. Most of the changes are in pom.xml files, however some code & documentation also has references to our versions and all need modifying. 
+* Edit all files (command line or IDE) to replace 'x.y-SNAPSHOT' with the next version, ie change '1.3-SNAPSHOT' to '3.11-SNAPSHOT'. Most of the changes are in pom.xml files, however some code & documentation also has references to our versions and all need modifying. 
 * If using an IDE like IntelliJ make sure you have all hits by searching again as by default only a limited number of hits are shown - see https://youtrack.jetbrains.com/issue/IDEA-157855 for Intellij advice.
 * Commit
 * Now remove all the release notes from the 'release-notes' directory other than README.md - so users will always get directed to the latest in master
--- a/developer-resources/tools/HTTPie-Curl.md
+++ b/developer-resources/tools/HTTPie-Curl.md
@@ -11,7 +11,7 @@ It uses commands such as:
 
 ```bash
 $ curl --insecure -X GET https://localhost:9443/open-metadata/platform-services/users/test/server-platform/origin
-Egeria OMAG Server Platform (version 3.10-SNAPSHOT)
+Egeria OMAG Server Platform (version 3.11-SNAPSHOT)
 ```
 
 Note that Egeria is using https, so if you have not replaced the provided self-signed certificate, ensure to add '--insecure' to any requests 
--- a/open-metadata-conformance-suite/docs/platform-workbench/test-cases/platform-origin-test-case.md
+++ b/open-metadata-conformance-suite/docs/platform-workbench/test-cases/platform-origin-test-case.md
@@ -37,7 +37,7 @@ open metadata services are activated.
       "successfulAssertions" : [ "Origin descriptor retrieved from platform." ],
       "unsuccessfulAssertions" : [ ],
       "discoveredProperties" : {
-        "Repository origin id" : "Egeria OMAG Server Platform (version 3.10-SNAPSHOT)"
+        "Repository origin id" : "Egeria OMAG Server Platform (version 3.11-SNAPSHOT)"
       }
 }
 ```
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-api/pom.xml
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-api/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>open-metadata-conformance-suite</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Open Metadata Conformance Suite API</name>
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-client/pom.xml
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-client/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>open-metadata-conformance-suite</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Open Metadata Conformance Suite Client</name>
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/pom.xml
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>open-metadata-conformance-suite</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Open Metadata Conformance Suite Server-side</name>
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-spring/pom.xml
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-spring/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>open-metadata-conformance-suite</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Open Metadata Conformance Suite with Spring</name>
--- a/open-metadata-conformance-suite/pom.xml
+++ b/open-metadata-conformance-suite/pom.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
-        <version>3.10-SNAPSHOT</version>
+        <version>3.11-SNAPSHOT</version>
     </parent>
 
     <name>Open Metadata Conformance Suite</name>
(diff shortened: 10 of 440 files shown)


open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
```diff
@@ -45,19 +45,19 @@ public class CategoryFVT {
...
-    private Set<String> createdCategoriesSet = new HashSet<>();
+    private final Set<String> createdCategoriesSet = new HashSet<>();
 
     public static void main(String[] args) {
         String url;
         try {
             url = RunAllFVTOn2Servers.getUrl(args);
             runWith2Servers(url);
         } catch (IOException e1) {
-            System.out.println("Error getting user input");
```
planetf1, line 56: In general we are using System.out.println across our 'interactive' applications and FVTs with the expectation of the I/O always going to console.

However we are seeing too much clutter in our build logs from FVTs, and ensuring any output generally uses logging (or audit) is the right approach. And our FVTs should really be run purely by the test runners like surefire, failsafe (for fvt)

But a main() that implements interactive tty (which is the intent here), and in a test case (vs library/production code) should probably be an exception - but only in the 'test driver' aspect ie main() or tightly related to that interaction. Indeed this part of the FVT is using a buffered input stream to query the URL if needed. Given that I think the change to log causes more obfuscation

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CheckSerializationFVT.java
```diff
@@ -44,11 +44,12 @@ public static void main(String[] args) {
             String url = RunAllFVTOn2Servers.getUrl(args);
             runWith2Servers(url);
         } catch (IOException e1) {
-            System.out.println("Error getting user input");
```
planetf1, line 47: similar comments on the other FVTs in the use of log vs stdout -- but only this outer wrapper. I won't mark the others up.

planetf1 (changes requested):
Thanks for the proposals
* Agree with many of the changes which clean up the code
* Disagree with the change to stdout (but understand the reasons for the proposal) - what do you think? 
* Suggest leaving the first unknown variables in with referenced guids

Interested in your comments also @davidradl

planetf1:
Gradle PR build failed with FAILURE: 

`Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)`

This is likely resource exhaustion, and an ongoing issue for which I've just submitted another attempt for a fix. See #6130.

I've also restarted the gradle build for this PR

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIMapper.java
```diff
@@ -746,11 +746,23 @@
...
 
-    public static final String NOTE_LOG_AUTHOR_RELATIONSHIP_TYPE_GUID  = "8f798c0c-6769-4a2d-b489-12714d89e0a4";
+    /*
+    The below constants which contain the pattern
+    ..AUTH..="GUID"
+    looks exactly like
+    myAuth = SECURITY_SECRET
+    which is considered a security risk.
+    As both constants are currently unused I've commented them out assuming they are intended to be used in the future.
+    If the contsnats are re enabled in the future they will raise 2 HIGH Security Hotspot errors on SONAR cloud.
```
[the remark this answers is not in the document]
wbittles: If the code was actually being used then I agree we should just turn that rule off in this case.
However the fact that code wasn't being  used meant that I would like the option to track the "unused", rather than the original security alert , but I also wanted rid of the High sev alert.

open-metadata-implementation/integration-services/search-integrator/search-integrator-api/src/main/java/org/odpi/openmetadata/integrationservices/search/ffdc/SearchIntegratorAuditCode.java
```diff
@@ -40,6 +40,13 @@ public enum SearchIntegratorAuditCode implements AuditLogMessageSet {
             "The integration connector is running but does not have a context.  This is a timing issue in the integration daemon.",
             "Gather information about the connector's configuration, the types of metadata it was integrating, the audit log messages " +
                     "from the integration daemon and its partner metadata server.  Then contact the Egeria community to get help."),
+
+    REGISTER_CATALOG_ISTENER("OMIS-SEARCH-INTEGRATOR-0004",
```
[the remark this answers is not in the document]
wbittles: fixed thank you

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
wbittles: I hoped that the output provided by the build tool (maven/gradle) would be good enough and we could simply redirect them to log. The only thing that actually needs to change is the call to printStackTrace everything else is an incidental change so isn't important enough to change just yet.

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIMapper.java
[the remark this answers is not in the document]
mandy-chessell: The security alert is clearly invalid in this case and needs to be flagged in the tool rather than making changes to the code.  This way the tool vendors improve their rules.

As new types are defined contributors are supposed to add the name and GUID into this mapping file so that they are ready for use in the generic handlers.  This is to increase the chances that the GUID values are right and discourage each OMAS from creating their own definitions. 

I am not sure it is helpful to flag the values that are not used in this list - it just creates more work when they are used (in fact they are likely to stay flagged as unused as they are used which is more misleading).  If it is a real problem then I woudl suggest putting one suppressWarnings at the top of the file.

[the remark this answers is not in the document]
wbittles: Thanks @mandy-chessell for responding ,
another issue with this code is that it's also wrongly being confused for a High Sev Security Hotspot.

A constant containing AUTH being assigned a Hex string , looks exactly like a developer has a hard code auth token in the source code. 

`public static final String NOTE_LOG_AUTHOR_RELATIONSHIP_TYPE_GUID  = "8f798c0c-6769-4a2d-b489-12714d89e0a4";
`
in this case that is wrong and could be turned off either locally or centrally and as I don't know how to change the the rules centrally , I've pushed a suggest fix that just turns off that specific alert. If you and Nigel find it acceptable I will open a new PR which turns off warnings for  only the 2 lines that are causing this alert to be raised, which  also happen to be our Highest Sev Security alerts in this scan.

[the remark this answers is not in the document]
mandy-chessell: The tool is creating a false positive since clearly this is setting a variable to a string that is in no way related to security.  

Ideally we need to get the scanning tool fixed.  If we can configure the tool so that the false posititive is recorded and this is passed back to the developers of the tool then there is a chance they will fix it.  If we just turn off the scanning then it will come back when an external party runs their own scan.

If there is no way to configure the tool then disabiling the rule as you suggest is the best we can do.

open-metadata-implementation/integration-services/search-integrator/search-integrator-api/src/main/java/org/odpi/openmetadata/integrationservices/search/ffdc/SearchIntegratorAuditCode.java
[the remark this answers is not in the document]
planetf1: I think it's a good idea to check if there's recent activity in the module, and if so consider if it might have an impact . Less so in my view for small, isolated change. New events etc - more so as that affects behaviour. I think this helps build community and engagement.

regardless, Github should automatically assign the code owner a review. That should be the formal way of ensuring they are notified and have the opportunity to review. Since I don't see any reviews assigned I suspect the CODEOWNER is not set (cc: @bogdan-sava ) so I would consider updating /CODEOWNERS accordingly

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
planetf1: There is some history in the way these FVTs came about. they started as interactive applications - and so the main() allowed for the code to be run directly at the cli, or in IntelliJ. If appropriate parms were not passed, or perhaps always, the code prompted for input using an input stream. unless redirected this uses stdin, so in that context of interactive tty, using stdout seems entirely reasonable, and using log less so.

That changes when we get to the actual test case implementations as these will either be called from this code (which could control output) or most likely from a test runner - via maven/gradle. Once there I strongly agree, we should use logging, and this has big benefits to improving the quality of output from the build by hiding noise, yet making it easy to capture.

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CheckSerializationFVT.java
planetf1: closing as it's covered by other conversations

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
planetf1: I agree with the removal of the stack trace.

If we wanted to cleanup the FVT behaviour more, I would remove this interactive prompting for input parms & the use of input streams/stdout, probably even remove the main, and instead expect the code to always be launched from a test runner or IDE. But that would be a broader community question

davidradl (approved):

davidradl:
@wbittles The changes look good to me, once the blocking conversations are resolved.

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
planetf1: I think in general removing calls to printStackTrace is appropriate - in our runtime libraries they should be managed through our auditog framework if needed. Inadvertent disclosure is a potential security issue.

I'm less worried about FVTs in not being operational code, but it is best practice, so I would also agree with those changes. The one caveat is to be sure that if someone runs the FVTs as a java application via main() that we don't lose the exceptions completely - ie they should still be logged, or we at least report they are in the tool.. or otherwise update docs to say how they work. can comment more when the pr is ready for review

davidradl: I agree with both your points. The main thing is that we can debug an error s in the FVT when they occur in the build. We have moved to using audit logging - that writes to the console for this. I agreed with Nigel - that as a developer if I run it locally as an app to debug it outside of the build - it would be necessary to see all Exceptions.

wbittles:
Thanks for all the input , the Sonar Security Hotspot alerts have now been resolved via another PR.

PR #6636, opened 2022-06-28
#6130 remove gradle memory override, update defaults for devs & build

planetf1:

## Description

Further attempt to stabilize the gradle PR build which is intermittently failing. No error - the process goes away. Believed to me resource exhaustion, probably memory

## Related Issue(s)
See #6130


planetf1:
Latest change with 6GB heap does not prevent the error - indicates this is likely not the cause. We can't really go higher as the github runners are only allocated 7GB ram.

Issue #6638, opened 2022-06-29
[BUG] local.server.id configuration in event bus is not used in access services configs


juergenhemelt:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I configure the `local.server.id` the default event bus this configuration has no effect on the access services created for the metadata server:

```bash
curl -f -k --verbose --basic admin:admin \
  --header "Content-Type: application/json" \
  "${EGERIA_ENDPOINT}/open-metadata/admin-services/users/${EGERIA_USER}/servers/${EGERIA_SERVER}/event-bus?topicURLRoot=itg.egeria" \
  --data @- <<EOF
  {
    "local.server.id": "itg.${EGERIA_SERVER}",
    "producer": {
      "bootstrap.servers": "${KAFKA_ENDPOINT}",
      "acks": "all",
      "security.protocol": "SSL",
      "ssl.keystore.location": "${SSL_KEYSTORE_LOCATION}",
      "ssl.keystore.password": "${SSL_KEYSTORE_PASSWORD}",
      "ssl.truststore.location": "${SSL_TRUSTSTORE_LOCATION}",
      "ssl.truststore.password": "${SSL_TRUSTSTORE_PASSWORD}"
    },
    "consumer": {
      "bootstrap.servers": "${KAFKA_ENDPOINT}",
      "group.id": "${KAFKA_CONSUMER_GROUP_ID}",
      "security.protocol": "SSL",
      "ssl.keystore.location": "${SSL_KEYSTORE_LOCATION}",
      "ssl.keystore.password": "${SSL_KEYSTORE_PASSWORD}",
      "ssl.truststore.location": "${SSL_TRUSTSTORE_LOCATION}",
      "ssl.truststore.password": "${SSL_TRUSTSTORE_PASSWORD}"
    }
  }
EOF
```

The resulting configuration for the access service (e. g. datamanager) is:
```json
        {
            "class": "AccessServiceConfig",
            "accessServiceId": 210,
            "accessServiceDevelopmentStatus": "TECHNICAL_PREVIEW",
            "accessServiceAdminClass": "org.odpi.openmetadata.accessservices.datamanager.admin.DataManagerAdmin",
            "accessServiceName": "Data Manager",
            "accessServiceFullName": "Data Manager OMAS",
            "accessServiceURLMarker": "data-manager",
            "accessServiceDescription": "Capture changes to the data stores and data set managed by a data manager such as a database server, content manager or file system.",
[...]

PR #6639, opened 2022-06-29
#6638 Fix overriding of consumer group id in kafka topic connector

planetf1:

## Description

Fixes overriding of consumer group id for kafka consumers. 

* If no consumer group id is set then the current behaviour continues -- we use the server id as default, both for the default event bus config, as well as during the setup of embedded connections (default)
* If local.server.id is specified in the default event bus config, this is used for all additional connections (*intent* of old behaviour)
* if group.id is specified in the consumer properties, this is used instead (higher priority) (more standard approach)

## Related Issue(s)

Fixes odpi/egeria-docs#484

Will need to be added to release notes, and main docs to clarify behaviour (we do not clearly document this currently)

### DIFF
--- a/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/build.gradle
+++ b/open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/build.gradle
@@ -63,6 +63,16 @@ dependencies {
     testImplementation 'org.mockito:mockito-inline'
     testImplementation 'org.mockito:mockito-core'
     testImplementation 'junit:junit'
+    implementation project(':open-metadata-implementation:admin-services:admin-services-api')
+    implementation project(':open-metadata-implementation:view-services:tex-view:tex-view-server')
+    implementation project(':open-metadata-implementation:view-services:tex-view:tex-view-spring')
+    implementation project(':open-metadata-implementation:view-services:rex-view:rex-view-server')
+    implementation project(':open-metadata-implementation:view-services:rex-view:rex-view-spring')
+    implementation project(':open-metadata-implementation:view-services:dino-view:dino-view-server')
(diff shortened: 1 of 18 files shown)


open-metadata-implementation/adapters/open-connectors/event-bus-connectors/open-metadata-topic-connectors/kafka-open-metadata-topic-connector/src/main/java/org/odpi/openmetadata/adapters/eventbus/topic/kafka/KafkaOpenMetadataTopicConnector.java
```diff
@@ -159,14 +161,19 @@ private void initializeTopic()
...
                  * wants to keep reading from the same list.  Thus it needs to be passed the group.id it used
                  * the last time it ran.  This is supplied in the connection object as the serverIdProperty.
+                 *
+                 * If explicitly set using the standard kafka property, then use that.  Otherwise, use our default.
                  */
-                serverId = (String) configurationProperties.get(KafkaOpenMetadataTopicProvider.serverIdPropertyName);
-                consumerProperties.put("group.id", serverId);
+
+                if (StringUtils.isEmpty((String)consumerProperties.get("group.id"))) {
+                    serverId = (String) configurationProperties.get(KafkaOpenMetadataTopicProvider.serverIdPropertyName);
```
juergenhemelt, line 169: I think this line must be outside of the `if`-block. Otherwise the `serverId` is never set.

planetf1: Agreed! It gets used later.

Issue #6700, opened 2022-07-03
[Enhancement] Provide a consistent generic handlers interface


mandy-chessell:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Please describe the new behavior that that will improve Egeria

A number of features have been consolidated into the common services in order to provide basic metadata management to the [OMASs](https://egeria-project.org/services/omas/):

* Support for external sourced elements that are locked read-only to all but the originator.
* Support for the maintenance and use of Anchors classification to group elements into a logical object for security and integrity management.
* Support for LatestChange classification on anchors.
* Support for effectivity dating.
* Support for the Memento classification that hides metadata elements to queries unless they are for lineage.
* Support for the automatic merging of identified peer duplicates and consolidated duplicates during queries.
* Support for specialized metadata security checks.

Most of this function has been added iteratively to the generic handlers and repository handler.  It involved the addition of new parameters on the method calls.  In order to minimise disruption to the OMASs while these enhancements were going in, the existing method calls were preserved with default values for the new function hard-coded.  These defaults are no longer consistent and are not suitable for all OMASs.  The result is some instability in the behaviour that we were trying to avoid.

For example, some methods default effectiveTime to `null` (any effective time which should only be used by OMASs that support the filtering/management of elements that are not currently effective) and others set it to `new Date()` which means retrieve only those elements that are effective now.  In a particular call path, the value of .

This enhancement proposes to remove all of the methods that default the new parameters so that each OMAS is exposed to the new parameters and can make an appropriate choice for their settings.  Once the OMAS makes a call to the generic handers, these parameters are set consistently throughout the code path.

### Alternatives

Continue the organic migration of the generic handlers API to expose these variables.

### Would you be prepared to be assigned this issue to work on?

- [X] I can work on this

PR #6704, opened 2022-07-06
Generic Handlers - use new method uncheckedLinkElementToElement on create methods

popa-raluca:
This PR solves two issues:
- Don't set up tableCount when creating RelationalTables
- Introduce new method uncheckedLinkElementToElement that doesn't do the check for the relationship existence, and use it in all handlers methods that create entities. 

## Testing
Tested locally & with fvts


open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/RelationalDataHandler.java
```diff
@@ -2265,25 +2265,14 @@ public String createDatabaseTableForSchemaType(String               userId,
                                                                          repositoryHelper);
         }
 
-        /*
-         * The schema type that connects the database schema asset to the tables has been created/retrieved.
-         * Now work out the position of the new table in the database schema type.  This is used to set the element position.
-         * Since this value begins with 0 as the first element, the table count is this table's position.
-         */
-        int tableCount = databaseTableHandler.countSchemaAttributes(userId,
```
lpalashevski, line 2273: Looks like effectiveTime was used in here in countSchemaAttributes call. Not sure if it is better to remove it from the method signature (then all the calls to this will need to be changed as well) or we dismiss codeQL alert.

lpalashevski (approved):
approved, minor comment on codeql alert about unused parameter

mandy-chessell (changes requested):
Please ensure that generic handler methods:
* include the flags forLineage, forDuplicateProcessing, effectiveTime for all methods that issue a retrieve.  
* All update methods should also have isMergeUpdate flag.  
* All create methods should include effectiveFrom and effectiveTo methods.
The OMASs should decide on the values of these fields - see #6700.

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/GlossaryCategoryHandler.java
```diff
@@ -149,22 +149,22 @@ public String createGlossaryCategory(String              userId,
...
+                                               OpenMetadataAPIMapper.GLOSSARY_TYPE_NAME,
+                                               glossaryCategoryGUID,
+                                               glossaryCategoryGUIDParameterName,
+                                               OpenMetadataAPIMapper.GLOSSARY_CATEGORY_TYPE_NAME,
+                                               false,
+                                               false,
+                                               supportedZones,
+                                               OpenMetadataAPIMapper.CATEGORY_ANCHOR_TYPE_GUID,
+                                               OpenMetadataAPIMapper.CATEGORY_ANCHOR_TYPE_NAME,
+                                               null,
```
mandy-chessell, line 166: effectiveTime should be passed as a parameter and not defaulted to null.

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIGenericHandler.java
```diff
@@ -13053,6 +13053,147 @@ public String multiLinkElementToElement(String             userId,
...
+        Date effectiveFrom = null;
+        Date effectiveTo   = null;
+
+        if (relationshipProperties != null)
+        {
+            effectiveFrom = relationshipProperties.getEffectiveFromTime();
+            effectiveTo   = relationshipProperties.getEffectiveToTime();
+        }
+
+        Date effectiveTime = getEffectiveTime(effectiveFrom, effectiveTo);
```
mandy-chessell, line 13130: This approach is not valid - effectiveTime should be passed explicitly

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/PortHandler.java
```diff
@@ -156,22 +156,22 @@ public String createPort(String              userId,
...
+            this.uncheckedLinkElementToElement(userId,
+                                               null,
+                                               null,
+                                               processGUID,
+                                               processGUIDParameterName,
+                                               OpenMetadataAPIMapper.PROCESS_TYPE_NAME,
+                                               portGUID,
+                                               portGUIDParameterName,
+                                               OpenMetadataAPIMapper.PORT_TYPE_NAME,
+                                               false,
```
mandy-chessell, line 168: forLineage and forDuplicateProcessing should be parameters not passed as fixed values

mandy-chessell:
Please see my comments - I am also in progress with #6700 and we need to coordinate so that we do not cause a lot ot rework for each other.

popa-raluca:
@mandy-chessell  I did the changes as we discussed, but I'm not sure that the parameters effectiveTo & effectiveFor are actually needed for the uncheckedLinkElementToElement method. In linkElementToElement they are used to retrieve only the relationships that are effective between those dates, but I'm not sure what should be their purpose in uncheckedLinkElementToElement. I was thinking maybe they need to be added in the relationshipProperties? This is not done at the moment, so they are unused

mandy-chessell:
@popa-raluca I agree that we should remove the effectiveFrom and effectiveTo parameters on the uncheckedLinkElementToElement method because those values would be difficult to edit if the effectivity dates changed.

mandy-chessell (approved):

PR #6707, opened 2022-07-06
DataEngineProxyChangePoller to include running with reports

alexandra-bucur:

Include running with reports in Data Engine Proxy

## Description

Have both the ability to run with previous implementation and also run with reports.

## Testing

Manually, using IGC and Data Stage Adapter.


alexandra-bucur:
The changes got into #6705 and this PR will be closed.

PR #6714, opened 2022-07-11
Resolves Sonar Scan "INSECURE CONFIGURATION" vulnerabilities  

wbittles:
## Description
The PR resolves the 27 insecure configuration vulnerabilities that are currently list against Egeria.

Sonar Scan had identified a number of incorrect boolean checks throughout the code base.
This PR just removes the unnecessary  boolean checks.

## Testing
This has passed a full mvn clean install

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
@@ -242,12 +242,12 @@ public enum ConformanceSuiteErrorCode
      */
     public String getFormattedErrorMessage(String... params)
     {
-        log.debug(String.format("<== ConformanceSuiteErrorCode.getMessage({})", Arrays.toString(params)));
+        log.debug(String.format("<== ConformanceSuiteErrorCode.getMessage(%s)", Arrays.toString(params)));
 
         MessageFormat mf = new MessageFormat(errorMessage);
         String result = mf.format(params);
         
-        log.debug(String.format("==> ConformanceSuiteErrorCode.getMessage({}): {}", Arrays.toString(params), result));
+        log.debug(String.format("==> ConformanceSuiteErrorCode.getMessage(%s): %s", Arrays.toString(params), result));
 
         return result;
     }


open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
```diff
@@ -234,6 +234,18 @@ public String getUnformattedErrorMessage()
         return errorMessage;
     }
 
+    /**
+     * Returns the string but with any special storage characters changes to underscores
+     *
+     * @return String
+     */
+    private String sanatiseLogString( String str ) {
```
planetf1, line 242: Suggest 'sanitize' or 'sanitise' for spelling, though only cosmetic

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
```diff
@@ -234,6 +234,18 @@ public String getUnformattedErrorMessage()
...
+        /*
+        quietly change any characters involved in storage hacks to a harmless underscore.
+         */
+        return str.replaceAll("[\n\r\t]", "_");
```
planetf1, line 246: I wonder if we should consider using the owasp logging plugin together with a logging configuration-> https://cheatsheetseries.owasp.org/cheatsheets/Injection_Prevention_in_Java_Cheat_Sheet.html

This could give us the best, quickest coverage - though wouldn't directly avoid the scanners reporting an issue in the code - and then we would need to deal with that!

Given that, I think the replacement of chars as you've done is a sound solution - and doesn't prevent us ALSO doing the owasp approach. it's something though they we would likely need across many components, which may need refactoring. But it's a step forward


May be useful to also check str isn't null? (or is this a no given the scan?)

planetf1 (changes requested):
An interesting PR -- many of the changes made me feel uncomfortable especially with null removal checks, but it seems we have additional insight here from the scanning tools verifying this is not the case.

I'm still not entirely convinced given code could change in future

I'll review the sonarcloud.io results (currently I'm not able to get to the scan results)

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedEntityProxyLifecycle.java
```diff
@@ -1371,7 +1371,7 @@ public void cleanup() throws Exception
          *   createdRelationshipsTUT
          */
 
-        if (createdRelationshipsTUT != null && !createdRelationshipsTUT.isEmpty()) {
+        if (!createdRelationshipsTUT.isEmpty()) {
```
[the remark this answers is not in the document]
planetf1: This can be seen in the sonar report at https://sonarcloud.io/project/issues?fileUuids=AW-QewLl1UqzkPNMUrDy&resolved=false&id=odpi_egeria&open=AXUwjvDz8Z59dRxV-JY9

As I noticed myself, this does have to be null , and so sonar is reporting this as a 'code smell'

This is explained at https://sonarcloud.io/organizations/odpi/rules?open=java%3AS2589&rule_key=java%3AS2589

On that basis I agree with the analysis, and therefore with the technical aspects of the change. In some ways it clarifies behaviour, and simplifies the reading of the code.

I think this may be worthy of wider discussion, perhaps posting an example on slack. I wonder also how important the 'code smell' is. We should certainly review, but do we need to action? If we decided not to, we could create a document listing rules we've decided to not adopt (but therein we may get issues with other scanners). Do our other  scans show issues with this logic, or is it unique to sonar?

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
wbittles: @planetf1 The scan results will really help and show most cases as obvious, this is the report I'm using and it's still available, I've just checked.

`https://sonarcloud.io/project/issues?resolved=false&sonarsourceSecurity=log-injection&id=odpi_egeria&open=AXnX8Vhw9ycZkvF8HLgx` 

most of these cases compared references to collection that had been "new" above.
Some where nested inside the same null check.
One was a duplicate check in the same 'if' 

if it helps with the confidence I've noticed that I get 'possible null pointer' alerts from one of the other scanners and this PR is coming in clean. 

While code will change continually I would expect Egeria use of 'null' to reduce rather than increase, the Java community seem to advocate empty lists and default values.

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedEntityProxyLifecycle.java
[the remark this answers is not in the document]
wbittles: @planetf1 In this case it doesn't feel that complicated, review the scans with the changes (or even review a few that are visible in the intellij plugin. In all but 1 case, each alert was obvious and just plainly wrong. 

These are being reported collectively as as a  classification  of security vulnerability  referred to as 'Insecure Configuration"

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSearch.java
```diff
@@ -207,11 +207,8 @@ public TestSupportedRelationshipSearch(RepositoryConformanceWorkPad workPad,
          */
         OMRSRepositoryConnector cohortRepositoryConnector = null;
         OMRSRepositoryHelper repositoryHelper = null;
-        if (workPad != null)
```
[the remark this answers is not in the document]
wbittles: on line 182 
.
`this.metadataCollectionId = workPad.getTutMetadataCollectionId();`

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
```diff
@@ -248,7 +247,8 @@
         MessageFormat mf = new MessageFormat(errorMessage);
         String result = mf.format(params);
 
-        log.debug(String.format("==> ConformanceSuiteErrorCode.getMessage(%s): %s", Arrays.toString(params), result));
+        /* The below code doesn't generate a SONAR vulnerability alert as the param has been formatted */
+        log.debug(String.format("==> ConformanceSuiteErrorCode.getMessage(%s): %s", Arrays.toString(params)), result);
```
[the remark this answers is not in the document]
davidradl: am I misunderstanding the code `String.format("==> ConformanceSuiteErrorCode.getMessage(%s): %s", Arrays.toString(params))` seems to indicate there are 2 expected string but there is only one value for them ,`Arrays.toString(params)`.  Would it not be better to have :` log.debug(String.format("==> ConformanceSuiteErrorCode.getMessage(%s): %s", Arrays.toString(params), result));`

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminServices.java
```diff
@@ -733,25 +731,22 @@ public VoidResponse deleteEventBus(String userId, String serverName)
...
-                {
-                    configAuditTrail = new ArrayList<>();
-                }
+            if (configAuditTrail == null)
+            {
+                configAuditTrail = new ArrayList<>();
+            }
 
-                configAuditTrail.add(new Date().toString() + " " + userId + " deleted configuration for default event bus.");
+            configAuditTrail.add(new Date().toString() + " " + userId + " deleted configuration for default event bus.");
```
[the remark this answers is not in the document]
planetf1: An interesting observation - one annoyance here too is that the timezone is not specified nor printed - it uses the 'default' which could be ambiguous.

That being said, it's something we can come back to another time (If so, suggest closing this feedback). May be worth considering an issue to check how we use dates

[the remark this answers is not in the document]
planetf1: as with previous date comment

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
planetf1: Yes, I'm ok with your proposed changes. It was a good challenge from the tools.

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedEntityProxyLifecycle.java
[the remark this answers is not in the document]
planetf1: I agree your proposal is correct. I'd probably myself have writing code following this anti pattern, so it was a good challenge. I'm ok with the change, especially as it will help us get to a much smaller list of issues, then we'll also be able to see much more easily any identified code flows where a change might mean a check is then needed

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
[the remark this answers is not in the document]
planetf1: I wonder if we could slf4j format replacement instead using {}. That's optional - especially as this is in test code (CTS). However there is a bug in the above - there is only one string replacement

open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedRelationshipSearch.java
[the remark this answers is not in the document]
planetf1: agreed.

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
planetf1: Ok - sounds good. Any improvement is fine, I was just wondering if - as you say - it was worth considering the bigger issue (async, diff pr)!. Will close as you are handling it

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminServices.java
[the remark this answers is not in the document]
wbittles: agreed

open-metadata-test/open-metadata-fvt/access-services-fvt/subject-area-fvt/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/fvt/CategoryFVT.java
```diff
@@ -89,7 +89,7 @@ public static void runIt(String url, String serverName, String userId) throws Su
         System.out.println("CategoryFVT runIt stopped");
         }
         catch (Exception error) {
-            error.printStackTrace();
+            log.error("The FVT Encountered an Exception", error);
```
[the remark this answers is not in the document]
wbittles: I missed one of these , so I'll open a new PR with  a fix for the missed one and I'll run that test with that fix.
This is no longer part of the PR as the commit has been merged.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminServices.java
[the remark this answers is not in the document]
mandy-chessell: The timezone is not included in the message so that machines running in different timezones produce consisent messages - we can not assume the administrator is local to the server.

open-metadata-conformance-suite/open-metadata-conformance-suite-api/src/main/java/org/odpi/openmetadata/conformance/ffdc/ConformanceSuiteErrorCode.java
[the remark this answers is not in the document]
wbittles: I've just tried using {}  mainly to see if the tool will accept it.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Graph.java
```diff
@@ -120,9 +120,14 @@ public String toString(StringBuilder sb) {
...
-        sb.append("NodeGUID=" + rootNodeGuid + "\n");
-        sb.append(("NodeTypeFilter=" + nodeFilter == null || nodeFilter == "") ? "none" : nodeFilter + "\n");
-        sb.append(("relationshipTypeFilter=" + relationshipFilter == null || relationshipFilter == "") ? "none" : relationshipFilter + "\n");
+        /*
+        The below code is triggering alerts in a number of scanners
+        It's not clear why only the first string + raised an alert and the two remaining calls didn't ?
+        using Objects.equals was recommended as being null-safe
+         */
+        sb.append("NodeGUID=").append(rootNodeGuid).append("\n");
+        sb.append((Objects.equals(nodeFilter, "")) ? "none" : nodeFilter + "\n");
```
davidradl, line 129: I assume you made the line 128 change to add in the append rather than using the + operator. If we are doing that we should change this line and the next to use append as well in the ternary else.

davidradl:
@wbittles I have revewed the open-types fvt, view , subject area changes -= which are the piecess I own. Apart from the one comment I raised - the changes look good.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminServices.java
```diff
@@ -678,10 +678,8 @@ public EventBusConfigResponse getEventBus(String userId, String serverName)
              */
             OMAGServerConfig serverConfig = configStore.getServerConfig(userId, serverName, methodName);
 
-            if (configStore != null)
-            {
-                response.setConfig(serverConfig.getEventBusConfig());
-            }
+            response.setConfig(serverConfig.getEventBusConfig());
+
```
mandy-chessell, line 682: I am not happy that this check for null (or any similar null check) is being removed.  The config store connector is pluggable and the logic that wraps it may change.  This check does no harm and ensures that it is resilient to such changes.

I just do not understand why this check is a security vulnerability.  If it is being flagged as such by the scanner, the issue should be losed as a false positive.

mandy-chessell (changes requested):
I am not happpy that checks for null are being removed.  Why are they are being flagged as security vulnerabilities?  Although they are not strictly necessary now, they make the code more resilient to change in the config store logic.  This project is too complex to assume a person changing a module is intimately acquainted with all of the modules it intertacts with.

open-metadata-implementation/access-services/subject-area/subject-area-api/src/main/java/org/odpi/openmetadata/accessservices/subjectarea/properties/objects/graph/Graph.java
wbittles: @davidradl  this fix wasn't planned , it was thrown up by the QL scan after the PR was opened.
In an attempt to keep the review as simple as possible we've adopted a policy of only changing exactly what the scan is alerting us against, we've tried the "cleaning up while we are there" approach and it quickly became complicated. 
So that's why I didn't make the changes you suggest , but if your prepared to approve it , I'll happily make the changes you suggest.

open-metadata-implementation/admin-services/admin-services-server/src/main/java/org/odpi/openmetadata/adminservices/OMAGServerAdminServices.java
wbittles: [Industrial Cybersecurity by Pascal Ackerman](https://learning.oreilly.com/library/view/industrial-cybersecurity/9781788395151/)

Insecure configuration vulnerabilities
Configurations play a key role in the security of an application. Often, systems and applications will run with a default configuration, pulled from the vendor's manual or from the Internet. This makes guessing passwords, bypassing login pages, and finding well-known setup vulnerabilities a breeze. Another form of insecure configuration management is where a configuration is just plain wrong, either from the start or after changes were made that compromise the security of the application or system. This faulty configuration can then end up getting used everywhere in the company.

Common attacks associated with configuration management vulnerabilities are as follows:

Server software flaws or misconfigurations ...

it's from O'Reilly.com and the snippet ends there , but you get the idea.

wbittles: @mandy-chessell it's perhaps worth calling out that the scanners are not claiming that we have these vulnerabilities, it's just that the code "smells" like it could have those type of vulnerabilities. It's a bit like what's considered "best practice"

wbittles: @mandy-chessell Apparently it can be difficult to review the actual SONAR alerts, which would make reviewing PR's like this extremely difficult. Are you able to review the original alerts 

https://sonarcloud.io/project/issues?resolved=false&sonarsourceSecurity=insecure-conf&id=odpi_egeria

if not would it help if I added a break down about why each conditional was being removed ?

wbittles: @mandy-chessell @planetf1 @davidradl 
There are internal requirements that state that,
While you are free to reject these fixes I will however need a commitment that security vulnerabilities will be resolved in a timely manner  ( depending on the severity ) and setting the warning to be a false positive isn't acceptable.
Can I please have confirmation on what the egeria communities policy is with regards to security issues in the egeria code base.

wbittles: Hi Mandy and thanks for your timely response.
There seems to be multiple points raised there , let me start with the easy one.

In every change made under this PR the following is true.
The check for null is performed after the object reference has already been assigned a value in the same module. 
The code that is testing for this branch condition is processing Egeria configuration data.
As the branch condition being tested is impossible the configuration of Egeria is deemed to be at best suspect. 
So to summarise what the tool is reporting is that the Egeria configuration processing is just wrong. 

It's not obvious how any of the changes you mention could possible change what the tool is reporting  and a discussion around a design that allows the injection of null pointer exceptions is out of scope of this PR.

The security security threat is a classification of known type of vulnerability named a "configuration vulnerability" , I first encountered the process as part of "hardening" but I am now seeing it discussed as a topic in it's own right. 
For what it worth , I strongly advise you to adhere to the expected behaviours rather that the the individual instances of a vulnerability , the numbers differences are  logarithmic in nature.

It's not a "code checking" tool , it a "standards adherence " reporting tool , it's the team that's being scanned and your already reporting the results. I pulled this piece of work from the Egeria technical debt backlog, so apologises if the discussion feels a bit like your arguing with a satnav.

mandy-chessell: Indeed I do believe it is a retrograde step to remove checking from the code - but if you want to put it in you can - I have registered my protest and we can fix the resulting NPE if a change is made to the underlying software.  I just dispair that we are making the code less resilient to change in the name of "security".

Most of these tools are designed for a single application - they do not understand about components and frameworks.   They do not understand that there are people with different levels of skill and experience working on the code.  However I would not be without them because they do find genuine problems in the code.  This type of change and reports (such like the one that requests removal of parentheses that the compiler does not need - but help humans understand the logic) are regretable.

PR #6721, opened 2022-07-20
Kafka debug improvements

planetf1:
## Description

Improved logging to aid in debugging a customer problem .. _work in progress_

* Adds additional debug to understand send retries
* Corrects log calls to use {} rather than string concatenation for performance
* Adds metrics for counts of messages sent, Kafka publish calls etc (log only for now)
* Reformats - file had mix of windows/dos formats with many CR, and some mixed tab/spaces - hard to edit
* Fixes some minor code warnings
* cleans up imports
* Sets thread id to topicName/threadid (Class already in log. Field truncates by default to final part of string, threadid preserves uniqueness, use log configuration to truncate less). There was some code to do this already, but it did not work. (this only applies where we have created a new thread)

## Related Issue(s)

tbd 

## Testing

tbd

tbd

## Additional notes

Still to do
 - Similar improvements for the consumer
 
At this point this is log only and primarily to support developer debugging. If it proves useful we can identify specific information to add to audit log messages, or perhaps another mechanism such as for metrics capture.

As an example, to configure useful debugging of the producer and consumer one could use:
```
export LOGGING_LEVEL_ORG_ODPI_OPENMETADATA_ADAPTERS_EVENTBUS_TOPIC_KAFKA=debug
export LOGGING_FILE_PATH="/Users/jonesn/logs/egeria"
export LOGGING_LOGBACK_ROLLINGPOLICY_MAX_FILE_SIZE=5GB
```

This will
* log debug messages for the Kafka topic connector
* set each log file to a max of 5GB
* Create a log file directory (in addition to console) for the log messages - this file does not have the threadid truncated
so instead of
```
2022-07-20 13:35:15.589 DEBUG 34318 --- [ic.registration] o.a.e.t.k.KafkaOpenMetadataEventConsumer : Found records: 0
```
we would get
```
2022-07-20 13:35:15.589 DEBUG 34318 --- [pool-8-thread-2/egeria.omag.openmetadata.repositoryservices.cohort.devCohort.OMRSTopic.registration] o.a.e.t.k.KafkaOpenMetadataEventConsumer : Found records: 0
```

### DIFF
--- a/open-metadata-implementation/governance-servers/data-engine-proxy-services/data-engine-proxy-services-server/src/main/java/org/odpi/openmetadata/governanceservers/dataengineproxy/rest/DataEngineProxyRestService.java
+++ b/open-metadata-implementation/governance-servers/data-engine-proxy-services/data-engine-proxy-services-server/src/main/java/org/odpi/openmetadata/governanceservers/dataengineproxy/rest/DataEngineProxyRestService.java
@@ -9,15 +9,24 @@ import org.odpi.openmetadata.governanceservers.dataengineproxy.auditlog.DataEngi
 import org.odpi.openmetadata.governanceservers.dataengineproxy.model.ProcessLoadResponse;
 import org.odpi.openmetadata.governanceservers.dataengineproxy.processor.DataEngineProxyService;
 
+import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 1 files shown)


planetf1:
The additional debugging is useful to understand behaviour & debug connectivity issues. However no change has been made to the actual function.

PR #6793, opened 2022-08-03
Propagate http headers to security connector context (related to #5700)

### LINKED ISSUE #5700: [Enhancement] User request HTTP header access from SecurityConnector

### Current Behavior

Currently the security connector only allows to manage access by checking userId of incoming request. For some cases this is okay. But now i need to add a little bit more restriction condition for incoming user request - the external group membership information, which is obtained from JWT token in my external application. I think that the best way to pass additional data in HTTP request is a header, so i put group membership inside it. And i faced the issue: there is no way to access any other data except userId from security connector. 

### Expected Behavior

I think a kind of interface to implement or overriding of existence access restriction methods could help to fix this issue. For example, if we will have a request`s header access, we will be able to process a JWT token, which is usually kept inside one 

### Alternatives

_No response_

### Any Further Information?

_No response_

- [X] I can work on this

MihaiIliescu:
Signed-off-by: Iliescu Cristian-Mihai <cristianmihaiiliescu@gmail.com>

## Description

Added a servlet filter that puts all headers configured in application properties into a thread local

## Related Issue(s)

Fixes #5700

## Testing

Manual multithreaded testing

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Link Checker
-        uses: lycheeverse/lychee-action@v1.5.0
+        uses: lycheeverse/lychee-action@v1.5.1
         with:
           # Can switch to true once we run clean
           fail: false
--- a/.github/workflows/merge-gradle.yml
+++ b/.github/workflows/merge-gradle.yml
@@ -39,7 +39,7 @@ jobs:
         run: './gradlew aggregateJavadoc'
       - name: publish
         if: ${{ github.repository == 'odpi/egeria' && github.ref == 'refs/heads/master'}}
-        uses: JamesIves/github-pages-deploy-action@v4.3.4
+        uses: JamesIves/github-pages-deploy-action@v4.4.0
         with:
           branch: gh-pages
           folder: build/docs/javadoc
--- a/.github/workflows/pr-gradle.yml
+++ b/.github/workflows/pr-gradle.yml
@@ -23,7 +23,7 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
       - name: Cache Gradle packages
-        uses: actions/cache@v3.0.4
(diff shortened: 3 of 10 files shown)


open-metadata-implementation/adapters/authentication-plugins/http-helper/src/main/java/org/odpi/openmetadata/http/HttpRequestHeadersFilter.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * HttpRequestHeadersFilter is a Java Servlet Filter that receives a list of http header names which are then
+ * saved from the request in HttpHeadersThreadLocal
+ */
+public class HttpRequestHeadersFilter implements Filter {
+
+    private Logger LOG = LoggerFactory.getLogger(HttpRequestHeadersFilter.class);
```
bogdan-sava, line 26: this suppose to be private static final

open-metadata-implementation/adapters/authentication-plugins/http-helper/src/main/java/org/odpi/openmetadata/http/HttpRequestHeadersFilter.java
```diff
@@ -0,0 +1,60 @@
...
+        HttpServletRequest req = (HttpServletRequest) servletRequest;
+        LOG.info(
+                "Reading request headers from : {}.",
+                req.getRequestURI());
+
+        Map<String, String> threadLocalHeaders = new HashMap<>();
+
+        for (String headerName : headerNames) {
+            String headerValue = req.getHeader(headerName);
+            LOG.info("Got header {} with value {}.", headerName, headerValue);
```
bogdan-sava, line 53: use debug instead of info

bogdan-sava (changes requested):

bogdan-sava (approved):

PR #6811, opened 2022-08-17
AC, AL - Change effectivity dates to now

alexandra-bucur:

## Description

Add now as date value for effective time in Asset Lineage and Asset Catalog handlers.

## Testing

JUnits


popa-raluca (approved):

bogdan-sava (changes requested):
LocalDateTime instead of LocalDate ?

alexandra-bucur:
> LocalDateTime instead of LocalDate ?

After we discussed, we decided that what was missing was the time and I changed the ClockService to include it too.

bogdan-sava (approved):

PR #6846, opened 2022-08-23
Governance Services & DE OMAS fixes

popa-raluca:
## Description

This PR comes as a result of testing the deduplication framework after the latest changes implemented in https://github.com/odpi/egeria/pull/6726 and covers the following fixes:
 * Fix status ordinal for Governance Action In Progress status
 * Retrieve SupportedGovernanceService relationships starting from the governance service end
 * Set statusPropertyName to null when retrieving ConsolidatedDuplcateLink relationships
 *  DE OMAS - set the qualifiedName on the Collection entity by using the correct super builder
 
## Testing

Tested locally with the sample Deduplication connector


open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIMapper.java
```diff
@@ -1635,7 +1635,7 @@
     public static final int    REQUESTED_GA_STATUS_ORDINAL               = 0;
     public static final int    APPROVED_GA_STATUS_ORDINAL                = 1;
     public static final int    WAITING_GA_STATUS_ORDINAL                 = 2;
-    public static final int    IN_PROGRESS_GA_STATUS_ORDINAL             = 3;
+    public static final int    IN_PROGRESS_GA_STATUS_ORDINAL             = 4;
```
davidradl, line 1638: I cant see the reason for this change

popa-raluca: The ordinal for GovernanceActionStatus.IN_PROGRESS is set to 4, not 3. The constant IN_PROGRESS_GA_STATUS_ORDINAL is used for logging, so without this change the audit log would say that the status is changed to UNKNOWN instead of IN_PROGRESS

alexandra-bucur (approved):

PR #6847, opened 2022-08-23
Add sample deduplication connector

popa-raluca:

Add sample deduplication connector. Provides basic implementation for handling duplicate entities. The matching of the duplicates is made based on the qualified names of the entities. 
## Testing

Tested locally with GAF and DE OMAS


planetf1:
Is this the first step?

I didn't see build/packaging related content? 
Should we place sample connectors in base egeria or elsewhere? We've got a general strategy of only keeping the core code in base egeria, though if it's a non-scenario based, basic connector (like in memory repo) I think there's a case for having in base.

popa-raluca:
> Is this the first step?
> 
> I didn't see build/packaging related content? Should we place sample connectors in base egeria or elsewhere? We've got a general strategy of only keeping the core code in base egeria, though if it's a non-scenario based, basic connector (like in memory repo) I think there's a case for having in base.

It's more a basic connector implementation for deduplication that we used in our tests so far. I don't think we're ready to have it scenario based yet, just wanted to have it somewhere in github. It is placed in governce-action-connectors along with the other basic connector implementations for governance. Maybe we can discuss about moving them to a more proper place at a later time?

planetf1:
@lcpopa Thanks for clarifying - makes sense, I agree with you. It does bring up a question of how we want to handle meantime, but that's a longer term question

planetf1 (approved):

lpalashevski (approved):
Approved with minor remark about the name, since this is remediation action perhaps name describing exactly the action (?) i.e. "QualifiedNamePeerDuplicateGovernanceActionConnector"

I see that you explain the purpose in the javadoc, I think this is more explicit.

PR #6900, opened 2022-09-22
Add ability to override AL OMAS topic listener connection properties

popa-raluca:

## Description
 This PR change the asset lineage topic connection auto-configuration processing to allow configuration properties overrides.

The connection object will be still retrieved from the paired OMAS (as defined in the "accessServiceConfig")
once the connection object is retrieved we check if "assetLineageTopicConnectionOverride" object is present in local server configuration we will override the configuraionProperties received with the local values (map)
if the "assetLineageTopicConnectionOverride" object is not present in the local configuration then everything stays as it is (connection is created based on the parameters received from the remote access server)


lpalashevski:
In addition further improvement: I do not like that initializeOLS call is blocking until the configuration is retrieved from the paired omas. This enforces certain deploy time rules and it is not optimal for containerized deployments. 
I think the server should try, if it fails then continues with startup and run background task to retrieve the connection and start the topic connection independently. We can open issue for this enhancement separately.

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/admin/OpenLineageServerOperationalServices.java
```diff
@@ -136,9 +136,8 @@ private void initializeOLS(OpenLineageServerConfig openLineageServerConfig) thro
         final String methodName = "initializeOLS";
         final String actionDescription = "Initialize Open lineage Services";
         Connection lineageGraphConnection = openLineageServerConfig.getLineageGraphConnection();
-        OLSSimplifiedAccessServiceConfig accessServiceConfig = openLineageServerConfig.getAccessServiceConfig();
 
-        Connection inTopicConnection = getAssetLineageOutTopicConnection(methodName, accessServiceConfig);
+        Connection inTopicConnection = getAssetLineageOutTopicConnection(openLineageServerConfig, methodName);
```
lpalashevski, line 140: At this point before the connector is create we need to merge the properties if they are defined in "assetLineageTopicConnectionOverride" coming form the server configuration document. 
Based on the current code, it looks like we need to replace configurationProperties map for the first object in the embeddedConnections we retrieved (the object inTopicConnection above).

=)

lpalashevski (changes requested):

PR #6905, opened 2022-09-27
OLS - update lineage job to handle partial lineage

pmarius22:
## Description

Modified the lineage job logic to handle processes that do not have either the input or the output link.


lpalashevski:
The current state:

- The implementation is complete
  - The background process creation is kept as is;
  - The sub-process node creation is done in two-way now; Input and Output are split into two logical steps so we query partially in or out and not break the end2end.
- Limited testing, we still need to properly test the functional change;

PR #6977, opened 2022-10-03
Propagate HTTP headers - Coco Pharma example

MihaiIliescu:
## Description

This change represents an example on how HTTP headers can be used for authorisation within a connector. This is a follow up to https://github.com/odpi/egeria/pull/6793 and https://github.com/odpi/egeria/pull/6851.

## Testing

 Functional testing

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-client/src/main/java/org/odpi/openmetadata/conformance/client/OpenMetadataConformanceTestReport.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-client/src/main/java/org/odpi/openmetadata/conformance/client/OpenMetadataConformanceTestReport.java
@@ -3,6 +3,7 @@
 package org.odpi.openmetadata.conformance.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import org.odpi.openmetadata.adapters.connectors.restclients.ffdc.exceptions.RESTConfigurationException;
 import org.odpi.openmetadata.adapters.connectors.restclients.ffdc.exceptions.RESTServerException;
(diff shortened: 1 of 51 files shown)


planetf1:
The open metadata security samples project includes 2 connectors currently, both based on coco pharma
 - platform
 - server
 
 This PR proposes a http security connector, and in that it follows the existing sample.
 
 As such I don't have any issue with merging this in, but it does bring up two questions
 
 a) As the three connectors are seperately identfied/configured, do we want them in seperately deployable jars - this is how we usually do things - which would mean 3 artifacts. However we don't have to do this, ie if closely related
 b) more of our samples are moving to repositories outside core egeria, since they are optional. Should we be thinking of that here?

open-metadata-resources/open-metadata-samples/open-metadata-security-samples/src/main/java/org/odpi/openmetadata/metadatasecurity/samples/CocoPharmaPlatformSecurityConnectorTokenBased.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * CocoPharmaPlatformSecurityConnector overrides the default behavior for the security connector
+ * to allow requests the Coco Pharmaceutical's server administrator APIs.  In this example,
+ * only Gary Geeke is allowed to issue these requests.
+ */
```
lpalashevski, line 23: Add example plain json token in the class javadoc so the user is able to understand clearly what this sample is using as input.

Example plan json payload data passed as part of the JWT in the Authorization header of the http request: 
```
{
  "sub": "garygeeke",
  "name": "Gary Geeke",
  "actions":["platform-administrator","platform-operator","platform-investigator"]
} 
```

lpalashevski (approved):
Approved. See minor comment about adding the example payload in the class javadoc.

PR #7047, opened 2022-10-21
Update demo users for the Egeria UI

mandy-chessell:
## Description

Removed James Bond as the admin user and made Erin Overview and Peter Profile admin users.


open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/resources/demo-users.yml
```diff
@@ -13,7 +13,7 @@ egeria:
   demoUsers:
     admin:
       username: admin
-      name: James Bond
```
sarbull, line 16: 👍

PR #7119, opened 2022-11-21
Update release template

lpalashevski:
Updated steps, added more links


planetf1 (approved):
thanks for the updates. Great to incrementally improve.

PR #7151, opened 2022-11-30
git7147 add classifications to update schema attribute

davidradl:

The update for schema attributes does not update the type embedded classification 

## Description

I set any classifications that are present into the entity. During testing this was only the embedded Type classification

## Testing

I ran the lineage sample to prove this worked

## Additional notes

I considered amending the update schema processing , so this change would benefit Egeria more widely and consistently. This looked like a big change, as it involved changing update Schema processing in the generic handlers and account for all callers. this fix I feel is more targeted and safer in the short term.


davidradl:
We will discuss this approach vs a more general approach in the community meeting tomorrow

mandy-chessell (changes requested):
This fix enables updateSchemaAttribute() to update any classification.  We are only seeking to update the TypeEmbeddedSchema classification, which is logically its schema type.  Other classifications should be update through their own methods (due to security checks etc).   The fix should look like the code in SchemaAttributeHandler lines 1817-1825.  As I said in previous comment, it is ok to create a new overloaded method that takes the schema attribute builder (with nested schema attribute builder) so there is no chance to regress the other methods.

It would be good to add a note to the javadoc of the updateSchemaAttribute() method that passes the InstanceProperties to say it only updates the entity.

Issue #7227, opened 2022-12-06
[BUG] Error in GlossarycontextHandler


davidradl:
### Current Behavior

Exception when running he sample lineage integration connector. It creates a glossaryTerm that is then checked to see if it has lineage relationships, unfortunately GlossaryContextHandler line 224 has this code
```
        return CollectionUtils.isNotEmpty(getTermCategorizations(userId, entityDetailGUID, entityDetail.getType().getTypeDefGUID()));
    }
```
The last parameter is incorrect - as the called code wants this to be the typename. The guid to checked to see if it a valid typename and an error occurs. 

Tue Dec 06 14:22:40 GMT 2022 cocoMDS1 Event OMAS-IT-INFRASTRUCTURE-0013 The IT Infrastructure Open Metadata Access Service (OMAS) has sent event of type: Element Reclassified
Tue Dec 06 14:22:40 GMT 2022 cocoMDS1 Event OMAS-ASSET-CONSUMER-0012 The Asset Consumer Open Metadata Access Service (OMAS) has sent event of type: Element Reclassified
Tue Dec 06 14:22:40 GMT 2022 cocoMDS1 Event OMAS-ASSET-OWNER-0012 The Asset Owner Open Metadata Access Service (OMAS) has sent event of type: Element Reclassified
Tue Dec 06 14:22:40 GMT 2022 cocoMDS1 Exception OMAG-REPOSITORY-HANDLER-0003 An unexpected error org.odpi.openmetadata.frameworks.connectors.ffdc.InvalidParameterException was returned to validateEntityGUID by the metadata server during getRelationshipsByType request for open metadata access service Asset Lineage OMAS on server cocoMDS1; message was OMAG-REPOSITORY-HANDLER-404-001 The validateEntityGUID method has retrieved a object for unique identifier (guid) a96f12f4-76f6-458a-9287-de34f6e217ea which is of type GlossaryTerm rather than type 0db3e6ec-f5ef-4d75-ae38-b7ee6fd6ec0a on behalf of method getRelationshipsByType
[...]

PR #7228, opened 2022-12-06
git7227 correct glossarycontext logic

davidradl:

correct the parameter to be a type name rather than a type guid


davidradl:
#7227

PR #7239, opened 2022-12-09
Change end2end endpoint in UI Chassis to end-to-end in order to have a uniform naming style

alexandra-bucur:


popa-raluca (approved):

open-metadata-implementation/user-interfaces/ui-chassis/ui-chassis-spring/src/main/java/org/odpi/openmetadata/userinterface/uichassis/springboot/api/lineage/OpenLineageController.java
```diff
@@ -64,7 +64,7 @@ public Graph ultimateSourceGraph(@PathVariable("guid") String guid , @RequestPar
      * @throws PropertyServerException from the underlying service
      * @throws OpenLineageException from the underlying service
      */
-    @GetMapping( value = "/entities/{guid}/end2end")
+    @GetMapping( value = "/entities/{guid}/end-to-end")
```
davidradl, line 67: The log entries in the openlineageservice to use` end2end` in the text - I suggest changing the text to `end-to-end` in line with this change

alexandra-bucur: Fixed in #7240

PR #7241, opened 2022-12-12
Asset Catalog exact-match filter fix

lcpopa:
Signed-off-by: Liviu Constantin Popa <liviu-constantin.popa@ing.com>

Search from asset-catalog omas was producing erroneous results when ticking exactMatch option

## Description

The reason for the erroneous result was the search string being regExed twice: once in the asset catalog code and once in the generic handler, when the exactMatch option is set to true. When it reached the graph repository it was no longer recognizable to allow entity extraction.
The solution that seem to fit best is to remove the search string decoration (regEx) in asset catalog and improve the one in generic handler to include the caseInsesitive flag, since their behaviour is intertwined. This meant the method signature needed the extra caseInsensitive param. Also, all calls to the method in generic handler are updated with the caseInsensitive flag defaulted to the value from the original behaviour, so there are no changes in expected outcome when called from other places

## Related Issue(s)

None found

## Testing

Manual

None

## Additional notes

None

### DIFF
--- a/open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIGenericHandler.java
+++ b/open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIGenericHandler.java
@@ -17,6 +17,8 @@ import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollec
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.search.SearchProperties;
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.PrimitiveDefCategory;
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.TypeDef;
+import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.TypeDefAttribute;
+import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.typedefs.TypeDefLink;
 import org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.repositoryconnector.OMRSRepositoryHelper;
 import org.odpi.openmetadata.repositoryservices.ffdc.exception.ClassificationErrorException;
 import org.odpi.openmetadata.repositoryservices.ffdc.exception.TypeErrorException;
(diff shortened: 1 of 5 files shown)


open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIGenericHandler.java
```diff
@@ -10835,6 +10845,7 @@ public List<EntityDetail> getEntitiesByValue(String       userId,
      * @param specificMatchPropertyNames list of property value to look in - if null or empty list then all string properties are checked.
      * @param exactValueMatch indicates whether the value must match the whole property value in a matching result, or whether it is a
      *                        RegEx partial match
+     * @param caseInsensitive set to true to have a case-insensitive exact match regular expression
```
davidradl, line 10848: From this javadoc It seems there are 3 permutations 
- exact match case insensitive
- exact match not case sensitive
- not exact match and the case insensitive flag is ignored.
 
I assume all 4 permutations are supported, if so we should remove the reference to exact match in this parameter description

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/AssetCatalogHandler.java
```diff
@@ -643,10 +643,7 @@ private List<EntityDetail> collectSearchedEntitiesByType(String userId,
             throws InvalidParameterException, PropertyServerException, UserNotAuthorizedException {
         List<EntityDetail> result = new ArrayList<>();
 
-        if(searchParameters.getExactMatch()) {
-            searchCriteria = repositoryHelper.getExactMatchRegex(searchCriteria, searchParameters.getCaseInsensitive());
-        }
-        else {
+        if(!searchParameters.getExactMatch()) {
             searchCriteria = repositoryHelper.getContainsRegex(searchCriteria, searchParameters.getCaseInsensitive());
```
davidradl, line 647: do we need to account for the case insensitive for the exact match case as well?

open-metadata-implementation/common-services/generic-handlers/src/main/java/org/odpi/openmetadata/commonservices/generichandlers/OpenMetadataAPIGenericHandler.java
lcpopa: The purpose of the changes in generic handler is to acomodate the caseInsensitive param, as part of a bug fix (minimal intrusion, not changing the actual behaviour). The generic handler was ignoring it and the method in repositoryHelper was defaulting it to false (case sensitive).  By exposing caseInsensitive param, it allows callers to pass it. The asset catalog will pass its value coming from the user, the other are sending what was previously used: false.

open-metadata-implementation/access-services/asset-catalog/asset-catalog-server/src/main/java/org/odpi/openmetadata/accessservices/assetcatalog/handlers/AssetCatalogHandler.java
lcpopa: - yes, hence the changes generic handler and not having the search string quoted twice (or regExed as I mentioned in the description). The value of the search string is modified according to ticked preferences
- only those that are not treated in the generic handler.

lpalashevski (approved):

Issue #7148, opened 2022-11-29
[BUG] CTS failures 


planetf1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Seeing CTS failures in both in-mem & graph (but this is on the experimental v4 build)- and I understand @dwolfson is also seeing in XTDB on v3.14

**This was seen when testing v4 (java 17/gradle)**

In my case I did a test run on Nov 22nd with in-memory using 'main'. 
That is currently failing 16 tests of 6105 with 6089 passing

### Expected Behavior

CTS to pass

### Steps To Reproduce

* Run CTS with current main - using inmemory for example
* retrieve results with
```
http --timeout=600 --verify=no GET https://localhost:9443/servers/cts/open-metadata/conformance-suite/users/garygeeke/report > out.json
```
* Print out the failing assertions with
```
cat out.json | jq -r '.testLabResults.testResultsFromWorkbenches[].failedTestCases[].unsuccessfulAssertions[]'
```

The result is
```
repository-relationship-reference-copy-lifecycle-03: DiscoveredNestedDataField reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: DiscoveredNestedDataField refreshed reference relationship matches original.
repository-relationship-reference-copy-lifecycle-03: RelatedKeyword reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: RelatedKeyword refreshed reference relationship matches original.
repository-relationship-reference-copy-lifecycle-03: LastAttachmentLink reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: LastAttachmentLink refreshed reference relationship matches original.
repository-relationship-reference-copy-lifecycle-03: AttachedRating reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: AttachedRating refreshed reference relationship matches original.
repository-relationship-reference-copy-lifecycle-03: DataFieldAnalysis reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: DataFieldAnalysis refreshed reference relationship matches original.
repository-relationship-reference-copy-lifecycle-03: AnnotationExtension reference relationship matches the relationship that was saved.
repository-relationship-reference-copy-lifecycle-11: AnnotationExtension refreshed reference relationship matches original.
[...]

PR #7247, opened 2022-12-16
Fix errors in headers and entity proxies detected by the CTS (#7148)

mandy-chessell:
## Description

There are three commits in this PR:

* A fix to EntityProxy.java to only setup qualifiedName in the uniqueProperties if it is not null.  This fixes the CTS failures.
* A fix to Subject Area OMAS's relationship mappers that is needed to prevent an NPE in the FVTs once the above fix is in place.
* A fix to set up the headerVersions in InstanceType, InstanceProperties and InstancePropertyValues.

The first two are good candidates to backport to 3.14.  The third is optional.

## Related Issue(s)

#7148

## Testing

Run CTS with graph repository.

Just bug fixes

### DIFF
--- a/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedEntityProxyLifecycle.java
+++ b/open-metadata-conformance-suite/open-metadata-conformance-suite-server/src/main/java/org/odpi/openmetadata/conformance/tests/repository/instances/TestSupportedEntityProxyLifecycle.java
@@ -143,7 +143,7 @@ public class TestSupportedEntityProxyLifecycle extends RepositoryConformanceTest
      * results in a 1000ms (1s) timeout.
      *
      */
-    private Integer           pollCount   = 100;
+    private Integer           pollCount   = 200;
     private Integer           pollPeriod  = 100;   // milliseconds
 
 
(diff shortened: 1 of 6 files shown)


planetf1:
The final commit to headers seems to break CTS. See referenced issue for more discussion

open-metadata-implementation/access-services/governance-engine/governance-engine-server/src/main/java/org/odpi/openmetadata/accessservices/governanceengine/handlers/MetadataElementHandler.java
```diff
@@ -698,6 +698,7 @@ else if (propertyValue instanceof StructTypePropertyValue)
         org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.instances.ArrayPropertyValue
                 omrsPropertyValue = new org.odpi.openmetadata.repositoryservices.connectors.stores.metadatacollectionstore.properties.instances.ArrayPropertyValue();
 
+        omrsPropertyValue.setHeaderVersion(InstancePropertyValue.CURRENT_INSTANCE_PROPERTY_VALUE_HEADER_VERSION);
```
planetf1, line 701: the header change appears to break CTS. See comments for details.

planetf1 (changes requested):

PR #7248, opened 2022-12-16
OpenLineageServices config - use 'inTopicConnection' to update the topic connection for AssetLineage

popa-raluca:
## Description
This PR updates the implementation for updating the AL OMAS connection object in the configuration. It adds a new property 'inTopicConnection' in the config that defines he connection object for AL. If present, it will override the default connection of the paired OMAS.


open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/admin/OpenLineageServerOperationalServices.java
```diff
@@ -189,21 +187,9 @@ private Connection getAssetLineageOutTopicConnection(OpenLineageServerConfig ope
...
-                for (EmbeddedConnection embeddedConnection : embeddedConnections) {
-                    Connection connection = embeddedConnection.getEmbeddedConnection();
-                    if (connection != null && KAFKA_OPEN_METADATA_TOPIC_PROVIDER.equalsIgnoreCase(connection.getConnectorType().getConnectorProviderClassName())) {
-                        connection.setConfigurationProperties(configurationProperties);
-                        embeddedConnection.setEmbeddedConnection(connection);
-                    }
-                }
-            }
-            assetLineageConnection.setConfigurationProperties(configurationProperties);
+        Connection inTopicConnection = openLineageServerConfig.getInTopicConnection();
```
lpalashevski, line 190: This check should be performed before the lookup with the omas client takes place. If locally configured then this method returns and does not perform remote call thus server should create the connection  and start independently.

lpalashevski (changes requested):

open-metadata-implementation/governance-servers/open-lineage-services/open-lineage-services-server/src/main/java/org/odpi/openmetadata/governanceservers/openlineage/admin/OpenLineageServerOperationalServices.java
lpalashevski: Referring to the check at line 191.

lpalashevski (approved):

Issue #7246, opened 2022-12-13
Remove 'split packages'


planetf1:
### Existing/related issue?

#2377

### Please describe the new behavior that that will improve Egeria

With 'Java Modules' (in Java 9 and above), 'split packages' are not allowed.

This is where the same package name is found in two different modules (=jars).
See https://www.logicbig.com/tutorials/core-java-tutorial/modules/split-packages.html

Generally in Egeria we adhere to this principle, however a few anomalies were found when working on the referenced issue https://github.com/odpi/egeria/issues/2377

In the 'demonstration' PR https://github.com/odpi/egeria/pull/7245 the modified directories needed were:
```
open-metadata-implementation/adapters/open-connectors/rest-client-connectors/rest-client-factory/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients >>>> open-metadata-implementation/adapters/open-connectors/rest-client-connectors/rest-client-factory/src/main/java/org/odpi/openmetadata/adapters/connectors/restclients/factory
open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices >>>> open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/api
open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/properties >>>> open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/api/configuration/properties
open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/configuration/registration >>>> open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/api/configuration/registration
open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/ffdc >>>> open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/api/ffdc
open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/ffdc/exception >>>> open-metadata-implementation/admin-services/admin-services-api/src/main/java/org/odpi/openmetadata/adminservices/api/ffdc/exception
[...]

PR #7307, opened 2023-01-04
Rename split packages (#7246)

mandy-chessell:
Signed-off-by: Mandy Chessell <mandy.e.chessell@gmail.com>

## Description

As described by @planetf1 in issue #7246, it is not good practice to have java classes from different modules contribute to the same Java package.  This PR renames classes and packages in order to ensure java packages sit within one module.

The changes to the java classes also required updates in consuming code - hence the large number of changed files.

## Related Issue(s)

#7246 

## Testing

Build, UT, FVT

Purely internal

## Additional notes

This fix is to enable the use of Java modules in release 4.0.

### DIFF
--- a/.github/workflows/pr-gradle.yml
+++ b/.github/workflows/pr-gradle.yml
@@ -23,7 +23,7 @@ jobs:
           distribution: 'temurin'
           java-version: '11'
       - name: Cache Gradle packages
-        uses: actions/cache@v3.0.11
+        uses: actions/cache@v3.2.2
         with:
           path: ~/.gradle/caches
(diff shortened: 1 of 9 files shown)


planetf1:
Thanks for doing the change. I've not validated this captures all the conflicts - hopefully, based on the original issue description.

The current module 'poc' is up to date with main as of today (including framework change) but I think this is harder to merge as there'll be a lot of conflicts from both my renames as well as the two large commits here. 

I think the next step is probably to discuss whether we are going ahead with module support at our f2f. I can show the poc, and we can make a decision (the module-info files can easily be removed). I'll then figure the best way to merge/rebuild the module info files.

Issue #7229, opened 2022-12-06
[BUG]  audit log entries are written twice to the console for sample lineage connector


davidradl:
### Current Behavior

Using the following configuration to define the Sample Lineage connector, log entries are written twice. 

I tried removing the repositoryServicesConfig section, but this is invalid. I assume there is a default log and the one that is specified in the config. I do not know how t configure so that audit log entries are not duplicated.

{
        "class": "OMAGServerConfig",
        "versionId": "V2.0",
        "localServerId": "02defb10-528c-4707-8049-ae7d82b88310",
        "localServerName": "lineagesample1",
        "localServerType": "Integration Daemon",
        "localServerURL": "https://localhost:9443",
        "localServerUserId": "OMAGServer",
        "maxPageSize": 1000,
        "integrationServicesConfig": [
            {
                "class": "IntegrationServiceConfig",
                "integrationServiceId": 606,
                "integrationServiceDevelopmentStatus": "IN_DEVELOPMENT",
                "integrationServiceContextManagerClass": "org.odpi.openmetadata.integrationservices.lineage.contextmanager.LineageIntegratorContextManager",
                "integrationServiceName": "Lineage Integrator",
                "integrationServiceFullName": "Lineage Integrator OMIS",
                "integrationServiceURLMarker": "lineage-integrator",
                "integrationServiceDescription": "Manage capture of lineage from a third party tool.",
                "integrationServiceWiki": "https://egeria-project.org/services/omis/lineage-integrator/overview/",
                "integrationServicePartnerOMAS": "Asset Manager OMAS",
                "defaultPermittedSynchronization": "FROM_THIRD_PARTY",
                "integrationServiceOperationalStatus": "ENABLED",
                "integrationConnectorConfigs": [
                    {
                        "class": "IntegrationConnectorConfig",
                        "connectorId": "ba6dc870-2303-48fc-8611-d50b49706f48",
                        "connectorName": "LineageIntegrator",
                        "metadataSourceQualifiedName": "TestMetadataSourceQualifiedName",
                        "connection": {
                            "class": "VirtualConnection",
                            "headerVersion": 0,
                            "qualifiedName": "Egeria:IntegrationConnector:Lineage:OpenLineageEventReceiverConnection",
                            "connectorType": {
                                "class": "ConnectorType",
                                "headerVersion": 0,
[...]

PR #7312, opened 2023-01-10
#7229 do not create a 2nd unneeded default audit log destination, when adding an audit log destination

davidradl:
#7229 the admin call to add an auditlog will add 2 audit log destinations if there is no RepositoryServicesConfig defined. 

## Description

I have added an overloaded method to create the default omrs configuration, the flag allows RepositoryServicesConfig to be created without a default audit log, so that in the add Audit log calling code can add the requested audit log and not end up with 2 destinations 

After this change, we don't add the default audit log destination during construction when we are about to add an audit destination

## Related Issue(s)

#7229

## Testing

Manual testing


mandy-chessell:
@davidradl This fix needs to be reversed as it is not the correct behaviour

Issue #7324, opened 2023-01-26
[BUG] Egeria Intellij build failure


MihaiIliescu:
### Current Behavior

I pulled the up to date egeria repo from main and after the maven build I tried to run the build from the intellij interface and I got an error: 

~/egeria/open-metadata-test/open-metadata-fvt/open-types-fvt/open-types-test/src/test/java/org/odpi/openmetadata/fvt/opentypes/server/TestBeansToAccessOMRS.java:126766:36
java: variable primitivePropertyValueForMap is already defined in method testCreateAgreementItem()

The class that fails is generated code and declares multiple times the same variable.

The issue is not only on my setup, I checked with a colleague and we got the error.

### Expected Behavior

The build finishes successfully.

### Environment

```markdown
- Egeria: 3.15-SNAPSHOT
- OS: MacOS Monterey v12.6.2 
- Java: 11
- Browser (for UI issues): -
- Additional connectors and integration: -
- Intellij Version my machine: IntelliJ IDEA 2022.3.1 (Ultimate Edition)
- Intellij Version colleague machine: IntelliJ IDEA 2021.1.3 (Ultimate Edition)
```

davidradl:
Locally I can build core Egeria and get a valid generated file  TestBeansToAccessOMRS.java.
I can confirm that the IJ build creates an invalid version of TestBeansToAccessOMRS.java with duplicate declarations.

planetf1:
I
 - cloned the egeria repo (with up to date main) - so completely clean
 - did a 'mvn clean install' at CLI
 - Opened the directory with intellij ie as a new project
 - Answered 'maven' when prompted for maven+gradle
 - Let IntelliJ index etc
 - Clicked 'build project'
 - Clicked ' enable annotation processing' when prompted
 
 The resultant build was clean
 
 I inspected the file, and at line 126766 of TestBeansToAccessOMRS.java I see
 ```
 MapPropertyValue testMapPropertyValue = null;
 ```
 which no errors - but also different content to the error you are reporting.
 
 I had left my JVM at default. On checking, I am using 'temurin 17' - and the language level defaulted to 17 (this is something I usually check/change after importing a project
 I also checked in settings under Maven, and I am running the Bundled maven 3 (3.8.1)
 
 @davidradl @MihaiIliescu is the sequence I followed exactly the same as you?

planetf1:
[...]

PR #7325, opened 2023-01-26
git7324 fix generator around maps

davidradl:

## Description

The generator was only working if there was one map attribute in an entity. I changed to code so that it did not define new variables for each map.

## Related Issue(s)

#7324 

## Testing

ran generator locally and saw correct output


planetf1:
Good to fix the generator, but note that additionally it appears the tests are not executed in either maven or gradle - and it's not clear when they ever were. So that would also need fixing

planetf1 (changes requested):
I think we should close this PR, and merge together in https://github.com/odpi/egeria/pull/7327 - if only to avoid merge issues.



That PR includes and credits your commits.

planetf1:
Ths change is included in https://github.com/odpi/egeria/pull/7327 which will be merged once ready so closing

Issue #6992, opened 2022-10-10
Add SPDX headers & enforcement to all egeria* repositories


planetf1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Please describe the new behavior that that will improve Egeria

Egeria currently enforces SPDX headers similar to:
```
# SPDX-License-Identifier: Apache-2.0
# Copyright Contributors to the ODPi Egeria project.
```
where the comments are file type dependent.

However this has not been applied across all repositories - typically files were initially created with these headers, but as there is no build-time enforcement, the headers can get dropped, or missed from new files.

In addition the current enforcement in the main egeria repository relies on maven, but we are moving to gradle.

This issue is opened to track verifying and enforcing across all repos, through a combination of
 - guidance in our documentation
 - seeing if we can create common build scripts to more easily maintain
 - adding in each repo

### Would you be prepared to be assigned this issue to work on?

- [ ] I can work on this

PR #7404, opened 2023-02-06
#6992 Add SPDX checking for copyright/license (java)

planetf1:

## Description

One of the missing aspects to the gradle build is SPDX header checking. There is a shortage of gradle plugins, or github actions.

My approximation so far is to use ‘checkstyle’  which allows specification of REGEXes which must match -for example one for copyright, one for license.

Due to the way the gradle build works, for now these only check java files.

In the process, I’ve noticed some omissions
* Some files incorrectly use `/* SPDX-License-Identifier: Apache 2.0 */` whilst it should be `/* SPDX-License-Identifier: Apache-2.0 */` .  Subtle, but it breaches the SPDX spec which specifically requires certain identifiers. There are over 2500 source files breaching this.
* We do have a number, even of java, source files with no copyright, or incorrect ie typos. The most entertaining was in glossary author where we have /* Copyright Contributors to the ODPi Egeria category. */ where clearly a search/replace on project->category was done

Both are easy to fix, especially the former, (the latter is slightly more tedious) and I’m happy to do so, but it will result in thousands of file changes in a commit.

For now I can submit this additional check, ERROR will be reported in the output log, but will not fail the build.
Interesting that our previous maven based checks did not find these errors.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,7 @@ plugins {
     id 'dev.jacomet.logging-capabilities' version "0.11.0"
     // This plugin helps resolve jakarta/javax dev.jacomet.logging-capabilities
     id 'org.gradlex.java-ecosystem-capabilities' version "1.1"
+    id 'checkstyle'
 }
 
 
@@ -29,6 +30,7 @@ allprojects {
     version = '4.0-SNAPSHOT'
 
     // Mostly java, so default to this for now
+    apply plugin: 'checkstyle'
     apply plugin: 'java'
     apply plugin: 'jacoco'
     apply plugin: 'org.gradlex.java-ecosystem-capabilities'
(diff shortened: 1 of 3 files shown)


planetf1:
Here's an example of the output in context:

Screenshot 2023-02-06 at 18 55 34

planetf1:
Note - It's possible to customize the message given for a rule violation

planetf1:
There is a 3rd party checkstyle plugin for intelliJ. This may make it easier to find issues in the UI.

Note that currently (with these rules) there are many errors ie

Checkstyle found 2,973 item(s) in 2,946 file(s)

planetf1:
This has been discussed in developer/tsc calls. No objects received so will now
 - merge this change using checkstyle to add WARNINGS into the build
 - raise new PRs to fix the warnings shown
 - raise PR to change from warning to error

PR #7406, opened 2023-02-10
Fix treating possible null members of collection

lcpopa:
## Description

Skipping processing of null members of portImplementations list, as they were sent by ds-proxy. Fix on that repo also in place, but keeping this one here for redundancy

## Related Issue(s)

None

## Testing

Manual

None

## Additional notes

None


popa-raluca (approved):

open-metadata-implementation/access-services/data-engine/data-engine-server/src/main/java/org/odpi/openmetadata/accessservices/dataengine/server/service/DataEngineRESTServices.java
```diff
@@ -1780,6 +1780,9 @@ private void upsertPortImplementations(String userId, String serverName, List<Po
         // first create port implementations sequentially
         try {
             for (PortImplementation portImplementation : portImplementations) {
+                if(portImplementation == null){
```
davidradl, line 1783: is it valid to have a null member? Should the fix be to not allow a null to be added rather than to ignore null members

lcpopa: This was fixed on the connector side, but decided to bring a null check in the omas too, even though it might seem redundant. In the general sense, I agree that a null in a list makes no sense; if such a case exists, I still have to come across it.

PR #7416, opened 2023-02-21
Upgrade to spring-boot 3 / java 17

bogdan-sava:
Upgrade dependencies to use spring-boot 3.
Change the code to use spring-boot 3.

### DIFF
--- a/.github/workflows/codeql-v4.yml
+++ b/.github/workflows/codeql-v4.yml
@@ -15,7 +15,8 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [main, egeria-release-4*]
-
+env:
+  CODEQL_ACTION_EXTRA_OPTIONS: '{"database":{"interpret-results":["--max-paths",0]}}'
 jobs:
   analyze:
     if: ${{ github.repository == 'odpi/egeria'}}


planetf1:
@bogdan-sava Thanks for proposing the Spring 3 update, which isn;t trivial! I'd like to review when appropriate. I'll plan to wait until the PR comes out of draft as you work through it.

build.gradle
```diff
@@ -520,7 +529,13 @@ dependencyAnalysis {
                 severity('fail')
             }
             onUnusedDependencies {
-                exclude("junit:junit", "org.junit.jupiter:junit-jupiter-api", "org.junit.jupiter:junit-jupiter-engine", "com.fasterxml.jackson.core:jackson-annotations")
+                exclude("junit:junit", "org.junit.jupiter:junit-jupiter-api",
+                        "org.junit.jupiter:junit-jupiter-engine", "com.fasterxml.jackson.core:jackson-annotations",
+                        "org.springframework.boot:spring-boot-starter-oauth2-resource-server",
+                        "org.springdoc:springdoc-openapi-starter-webmvc-ui",
```
planetf1, line 535: I think the spring exclusions can be removed. with just one change later they are not needed - so dependencies are correct. That one change relates to making the springdoc dependency 'runtimeOnly'.

This is a reminder there are some other cleanups that should be done with gradle/ junit, but I will look at that once your spring change is merged

I am also ok if you want to merge as-is - I don't notice anything that would cause problems/regression

I will look at the junit/jupiter/jackson in a separate issue after merging - that isn't specific to your spring change

planetf1 (commented):
Looks great. only a minor comment on dependency scope. It's pointed out some things I missed outside the spring scope also which I can look at in a future PR.

Looking at codeql issue next

planetf1:
Looking at the codeQL issue - https://github.com/github/codeql/issues/8712 is related. It seems as if the sarif file (which contains info about components and scan results) is > 512MB. The referenced issue talks though how to identify the cause. Most likely it's due to a lot of scanning findings added for springboot3/spring6

Either we can try and debug this now -- there may be genuine issues we want to fix, or, pragmatically we set codeQL to no longer be an enforced check & merge anyway. Since we are in the middle of a longer release cycle I think that would be ok. The risk is that there is a genuine issue found, but equally it could be very minor, or a scanner bug!

Interestingly sonatype was fairly happy....

bogdan-sava:
I did a change in the codeQL workflow, it is fail but at least now shows the report.
The problem is that it shows a lot of issues that aren't related to the PR.

planetf1:
The commit history here has gone somewhat awry as there are a lot of commits already in master.

I recommend the PR is rebuilt based off current master via cherry-pick or rebase, as otherwise 
 a) the list of change is confusing to review
 b) there's a risk of regression if any changes are accidentally undone

planetf1:
the changes looked ok until https://github.com/odpi/egeria/pull/7416/commits/ced86f3930dee1784d58546da68a363fff48060e

Another option may be to revert back to that version, and then pull from master.
Or potentially rebase (this worked well when I was working on changes for v4)

planetf1:
* I notice codeQL is no longer failing - it's just reporting alerts. I suggest leaving the change you made to the codeQL build in place, in case it fails again -- wan then followup with the dev team
* it's hard to see where all the error reports are - or why we suddenly got more, but I think it may be best to skip over the check for now. We know we have >2600 observations to action and need a seperate work thread to figure out how to deal with them
* As above, this PR seems to have an odd/divergent history and should be rebuilt as it could cause odd merge behaviour, I think it's important we are clean. This might mean either cherry-picking, unwinding to before the last merge, rebasing (happy to take a look if it helps)

planetf1:
@bogdan-sava let me know if you need help with any untangling/cherry-pick/rebase -- hopefully the fixes themselves are good, and there's just some git sorting to do.

planetf1 (changes requested):
Set review status - PR needs a rebuild.

planetf1:
@lpalashevski @mandy-chessell @bogdan-sava I offer #7506 as a cleaned-up PR which addresses the commit history/merge observations. It has 15 commits vs 143 here

See PR for more info, but I don't profess to offer anything new, I've just cherry-picked from the original work (which retains commit credit).

planetf1:
replaced by #7506

Issue #5955, opened 2021-11-26
Add docker build tag for 'latest release'


planetf1:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Docker images are tagged
 - with a version ie '3.3' or '3.3-SNAPSHOT' every time
 - with 'latest' when run from master
 
 The end result is that someone can follow a fixed release, or always get from master easily.
 
 However getting the 'latest officially released' image requires explicitly updating the version

### Expected Behavior

Add a tag to the release pipeline 'stable'
This will always pull the latest actual released code

### Alternatives

The alternative would be to
 * Add a new tag 'experimental', or even 'master' or similar to tag images built from master & not use latest
 * Use 'latest' to mean the last good release (a change of behaviour)

### Would you be prepared to be assigned this issue to work on?

- [X] I can work on this

Trivial to do technically, just need to decide what is best from a usability perspective

planetf1:
this was fixed by PR #7495

PR #7495, opened 2023-03-06
Use stable docker tag for releases

ilovechai:
## Description

- [x] Change latest docker tag to stable for releases

## Related Issue(s)

Fixes (https://github.com/odpi/egeria/issues/5955)

### DIFF
--- a/.github/workflows/release-v4.yml
+++ b/.github/workflows/release-v4.yml
@@ -74,7 +74,7 @@ jobs:
         uses: docker/build-push-action@v4
         with:
           push: true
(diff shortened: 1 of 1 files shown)


planetf1 (approved):
Thanks - a useful change!

Consider if any doc updates would help, though I'm aware that is a bigger area, especially as we are looking at improvements for cloud native generally

Also useful to rollout to other repos. However a better fix there is to develop some reusable actions/fragments/repo for common actions.

So, good to merge this now

PR #7500, opened 2023-03-07
(Test) #7499 Fix code QL scan

### LINKED ISSUE #7499: codeQL checks are failing
We have made multiple issues with codeQL scans failing in unknown ways
 - https://github.com/odpi/egeria/pull/7416
 - https://github.com/odpi/egeria/actions/workflows/codeql-v4.yml

* In some cases the analysis of rules just ends
* In some cases the job seems to be terminated
* In other cases the sarif file exceeds 0.5Gb and fails in json processing

Our codebase is large..

CodeQL works by
 - running a normal build
 - collecting analysis data into a sarif file
 - running rules against that file
 -
Options include
 - enable debug / try and identify cause
 - use a filter to slim the generated sarif file before rule execution
 - try and build less components (already tried with removing open types & other tests)
 - abandoning codeQL
 
 It's worth noting that sonatype also has issues with our code base - many tools are currently failing due to gradle hitting an out of memory error when analysing dependencies of our 300+ components (IntelliJ can also struggle with its tooling). This may indicate our repo is too large and complex and we need to consider refactoring
 
 For now the codeQL check is not mandatory and will likely fail

planetf1:
## Description

Fixes #7499

* reduces RAM usage limit for codeQL, as recommended by github support

### DIFF
--- a/.github/workflows/codeql-v4.yml
+++ b/.github/workflows/codeql-v4.yml
@@ -17,9 +17,6 @@ on:
     branches: [main, egeria-release-4*]
   workflow_dispatch:
 
-env:
-  CODEQL_ACTION_EXTRA_OPTIONS: '{"database":{"interpret-results":["--max-paths",0]}}'
-
 jobs:
   analyze:
     if: ${{ github.repository == 'odpi/egeria'}}


planetf1:
Ticket opened with github support -> https://support.github.com/ticket/personal/0/2045907

Issue #7505, opened 2023-03-09
[BUG] ConnectedAssetUniverse not gathering all asset properties


LeeVi3w:
### Current Behavior
Inside the context of a Java application, I want to use the `ConnectedAssetUniverse` class from the **ocf-metadata-client** to retrieve the properties of an asset using its GUID.

After creating a `ConnectedAssetUniverse` object using an asset's GUID, the resulting object contains null values for fields that  are otherwise present in the AssetResponse retrieved from the OMAS.

Some of the missing attributes which are found inside the `AssetResponse` but not inside the `ConnectedAssetUniverse` object after processing are:
- Classifications
- qualifiedName
- GUID
- AssetType

The `ConnectedAssetUniverse` is supposed to gather the `SchemaAttributes` of a `SchemaType` (if present), which are also missing, along with the base attributes of the Asset class (and some attributes higher in the hierarchy).

### Expected Behavior

The object should contain the same attributes that are returned when calling the assetUniverse API through a tool such as Postman, as well as some extra attributes specific to the functionality of the client (i.e. SchemaAttributes).

### Creating the asset
The asset was created following these steps:

1. _Create simple CSVFile asset using the Asset Manager OMAS_ 
(/access-services/asset-manager/users/{{user-id}}/data-assets)
```
{
    "class": "DataAssetRequestBody",
    "elementProperties": {
        "class": "DataAssetProperties",
        "typeName": "CSVFile",
        "qualifiedName": "Synthetic Financial Datasets For Fraud Detection",
        "guid": "paysim-csv"
    },
    "metadataCorrelationProperties": {
        "assetManagerGUID" : "4b202442-77ca-473e-89da-40d3806fb4ad"
    }
}
```
2. _Create simple TabularSchemaType using the Asset Manager OMAS_ 
(/access-services/asset-manager/users/{{user-id}}/schema-types)
```
{
    "class": "SchemaTypeRequestBody",
    "elementProperties": {
        "class": "SchemaTypeProperties",
        "typeName": "TabularSchemaType",
        "qualifiedName": "paysim-csv-schema-type"
    },
    "metadataCorrelationProperties": {
        "assetManagerGUID" : "1458cce2-8c6f-4126-8eb2-95a90922158f"
    }
}
```
3. _Link SchemaType to Asset using the AssetManager OMAS_ 
(/access-services/asset-manager/users/{{user-id}}/parents/{{asset-id}}/CSVFile/schema-types/{{schema-id}})
```
{
[...]

PR #7534, opened 2023-03-16
Redesigned ConnectedAssetUniverse instantiation

LeeVi3w:
## Description

Changed the instantiation method of `ConnectedAssetUniverse` from public constructors to static factory methods which return new objects.

This change was made to allow the use of the `AssetUniverse` super constructor with parameters as to set the missing properties reported in **issue** https://github.com/odpi/egeria/issues/7505. This decision was taken due to the fact that the super constructor needed data which was retrieved after its call inside the `ConnectedAssetUniverse` constructor.

The static factory methods introduced (one for each existing constructor) have the advantage of being able to process the needed data beforehand and then passing it to the `ConnectedAssetUniverse`, which in turn can call the super constructor that populates the missing properties with the given data. The newly created `ConnectedAssetUniverse` object is then returned.

These static factory methods are called create, and they make use of only one (private) `ConnectedAssetUniverse` constructor which takes an `AssetResponse` object and extracts the `Asset` bean containing the needed properties.

I have chosen this approach to avoid having to separate the data gathering logic from the class itself and to avoid having to set the missing fields in a dedicated function when there already exists a super constructor that deals with this.

## Related Issue(s)

https://github.com/odpi/egeria/issues/7505 [BUG] ConnectedAssetUniverse not gathering all asset properties

## Testing

The testing has been manually done using an instance of ConnectedAssetClient inside a Java application.


open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/SchemaAttributes.java
```diff
@@ -30,6 +31,74 @@ public SchemaAttributes(int totalElementCount,
...
+
+                        cachedElementPointer = 0;
+                    }
+                    catch (PropertyServerException error)
+                    {
+                        /*
+                         * Problem retrieving next cache.  The exception includes a detailed error message,
+                         */
+                        throw new OCFRuntimeException(OCFErrorCode.PROPERTIES_NOT_AVAILABLE.getMessageDefinition(error.getReportedErrorMessage(),
+                                this.toString()),
```
mandy-chessell, line 92: Align parameters for getMessageDefinition() under error.getReportedErrorMessage()

open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/SchemaAttributes.java
```diff
@@ -30,6 +31,74 @@ public SchemaAttributes(int totalElementCount,
...
+                        log.debug("==> maxCacheSize:" + maxCacheSize);
+
+                        return retrievedElement;
+                    }
+                    else
+                    {
+                        /*
+                         * Throw runtime exception to show the caller they are not using the list correctly.
+                         */
+                        throw new OCFRuntimeException(OCFErrorCode.NO_MORE_ELEMENTS.getMessageDefinition(this.getClass().getSimpleName()),
```
mandy-chessell, line 60: Align parameters for getMessageDefinition under this.getClass().getSimpleName()).

Use this.getClass.getName().

open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/SchemaAttributes.java
```diff
@@ -30,6 +31,74 @@ public SchemaAttributes(int totalElementCount,
...
+                                this.toString()),
+                                this.getClass().getName(),
+                                "next",
+                                error);
+                    }
+                }
+
+                return true;
+            }
+        };
```
mandy-chessell, line 101: is semi-colon needed?

open-metadata-implementation/framework-services/ocf-metadata-management/ocf-metadata-client/src/main/java/org/odpi/openmetadata/frameworkservices/ocf/metadatamanagement/client/ConnectedAssetUniverse.java
```diff
@@ -129,33 +166,29 @@ public ConnectedAssetUniverse(String   serviceName,
...
      */
-    public ConnectedAssetUniverse(String        serviceName,
-                                  String        remoteServerName,
-                                  String        omasServerURL,
-                                  String        userId,
-                                  String        assetGUID,
-                                  OCFRESTClient restClient) throws InvalidParameterException,
-                                                                   PropertyServerException,
-                                                                   UserNotAuthorizedException
+    public static ConnectedAssetUniverse create(String serviceName,
```
mandy-chessell, line 169: Align the parameter names

mandy-chessell (approved):
Looks good - just a few layout issues

open-metadata-implementation/frameworks/open-connector-framework/src/main/java/org/odpi/openmetadata/frameworks/connectors/properties/SchemaAttributes.java
LeeVi3w: The semi-colon is needed because the bracket is part of an anonymous class instantiation, otherwise it produces a compilation error.

mandy-chessell (approved):

PR #7652, opened 2023-05-04
Make displayName deprecated for Process type, adjust Asset Catalog to check for active status

alexandra-bucur:
## Description

I made displayName property deprecated for the Process type, as it was active. This is used in Asset Catalog for search.

## Testing

Manual


open-metadata-resources/open-metadata-archives/open-metadata-types/src/main/java/org/odpi/openmetadata/opentypes/OpenMetadataTypesArchive.java
```diff
@@ -773,6 +775,42 @@ private void update0710DigitalServices()
         this.archiveBuilder.addTypeDefPatch(updateDigitalProductClassification());
     }
 
+    private void update0010BaseModel() {
+        this.archiveBuilder.addTypeDefPatch(updateProcess());
```
mandy-chessell, line 779: Move first brace under "p" of private to match layout of the rest of the file

mandy-chessell (dismissed):

lcpopa (approved):

PR #7691, opened 2023-05-24
#7690 Add plain server chassis - for custom chassis composition

### LINKED ISSUE #7690: [Enhancement] Add 'plain' server chassis jar into published artifact
### Existing/related issue?

_No response_

### Please describe the new behavior that that will improve Egeria

The org.ordp.egeria:server-chass-spring artifact currently contains
 - application jar (the one built by springboot)
 - sources.jar (classifier:sources)
 - javadoc.jar (classifier:javadoc)
 
 With this setup, the artifact can be used to run egeria.
 
 However when building a custom server chassis, we may wish to selectively choose which modules to include. In particular this may be done externally rather than modifying the core egeria build
 
 In this case the 'plain' artifact is useful, since it includes a simple Spring Application, and I propose to add it with the 'plain' classifier
 

### Alternatives

_No response_

### Any Further Information?

I am making use of this as part of the egeria cloud-native work to prototype the build of a custom server chassis - in this case adding healthchecks.

I don't see any downsides

yes

planetf1:
## Description

Adds the 'plain' server chassis into the server-chass-spring artifact as per issue #7690

This can be consumed via
```
    implementation "org.odpi.egeria:server-chassis-spring:${egeriaversion}:plain"
```

## Related Issue(s)

Fixes #7690

## Testing

1. Apply PR #7690 into core egeria & rebuild with `./gradlew build publishToMavenLocal`
2. Build this project with `./gradlew clean build`
3. Go to healthcheck/build/libs
4. `cp ../../src/main/resources/*p12 .`
5. `jar ./healthcheck.jar`

This will launch the chassis. Then the new API can be verified with

```
 egeria-cloudnative git:(main) ✗ http --verify=no --pretty=format GET https://localhost:9443/open-metadata/healthcheck/platform/health
HTTP/1.1 200
Connection: keep-alive
Content-Length: 19
Content-Type: text/plain;charset=UTF-8
Date: Wed, 24 May 2023 12:41:41 GMT
Keep-Alive: timeout=60

Platform is healthy
```

### DIFF
--- a/open-metadata-implementation/server-chassis/server-chassis-spring/build.gradle
+++ b/open-metadata-implementation/server-chassis/server-chassis-spring/build.gradle
@@ -112,12 +112,9 @@ test {
 }
 
 // See https://stackoverflow.com/questions/61197984/bootjar-mavenjar-artifact-wasnt-produced-by-this-build
-// Remove the regular (plain) jar & replace with the SpringBoot version
+// Ensure the SpringBoot jar is included -- we leave the 'plain' jar in-situ too, for composing alternate chassis
 configurations {
     [apiElements, runtimeElements].each {
-        it.outgoing.artifacts.removeIf {
-            it.buildDependencies.getDependencies(null).contains(jar)
-        }
         it.outgoing.artifact(bootJar)
     }
 }


mandy-chessell (changes requested):
"plain" is not very descriptive . It would be easier to use if the name matched the build option - eg server-chassis-only

planetf1:
Thanks @mandy-chessell will be useful for prototyping.

The 'plain' designation is build technical, and spring-boot default. It refers to only the java code contained within this module, not any dependencies. It isn't even runnable in any way.
* A similar pattern may be useful in other maven artifacts where we currently build an uber jar, as it facilitates external composition
* The same approach of using classifiers can be used for any purpose - so we could define useful combinations of dependencies that provide different variations such as 'minimal-chassis' or 'integration-chassis' if we wanted, labelling each with a different classifier. Though they would all go into the same artifact which would get bigger.  For the HMS connector we have a need to build variants for different levels of HMS
* An alternative to classifiers is to publish to different artifacts. Which to use may depend on how similar or different the jar/file/archive is.
*

PR #7723, opened 2023-06-08
Add missing Javadoc to Asset Catalog and move it to Tech Preview

alexandra-bucur:
## Description
Added missing Javadoc for packages, classes, methods, enums, etc. Changed Asset Catalog OMAS status from In Development to Tech Preview.


lpalashevski (approved):
Looks good, I approve with one remark related to the comment I have for documenting all enums used for error/audit messages. I think we should add the value of the code in the javadoc text so it will show up in the html rendered docs. This makes it easier to read/troubleshoot using the docs - this is also done in other access services.

Lets make this change for consistency.

PR #7724, opened 2023-06-09
Improvements to deployment

mandy-chessell:
## Description

Principally this PR aims to align the contents of the distribution to our documentation:
* Adds a new folder called "platform" to the distribution which represents the platform.  In this folder are:
     * The connector jar files (same as jars in server/lib)
     * A jar file called omag-server-platform-{release}.jar.  This is a copy/rename of server/server-chassis-spring-{release}.jar
* The server folder in the distribution is unchanged to allow time for people to migrate to the new jar file name.
* The FVTs have been changed to run called omag-server-platform-{release}.jar

In addition
* I fixed a bug in the getServerStatus REST API in platform services to return the status of inactive servers.
* Many of the modules no longer have Serializable beans.  This serialization is unnecessary.  The only module that needs Serializable beans is the repository services.
* I added a sample logback.xml for the platform that includes rolling logs.

## Related Issue(s)

None

## Testing

* The FVTs work with either the server/server-chassis-spring-{release}.jar or omag-server-platform-{release}.jar
* the labs work with omag-server-platform-{release}.jar

There is a matching change needed to the docs to change the name of the Jar to use when starting up the platform.  Also to remove the use of server-chassis except in the module description.

## Additional notes

The impetus behind this change is that hte cloud native workgroup want to add a new runtime to support a single server.  Therefore it is important for understandability to make sure our code artifacts distinguish between code that supports the platform and code that supports a server.  The server-chassis naming is from a time before we added mulit-tenant support.

There is a follow-on PR to rename the server chassis modules to platform chassis

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,7 @@ plugins {
  */
 allprojects {
     group = 'org.odpi.egeria'
-    version = '4.1-SNAPSHOT'
+    version = '4.2-SNAPSHOT'
 
     // Signing/publish used in parent & child projects
     apply plugin: 'maven-publish'
--- a/developer-resources/License-Example-Files/License_for_POM_Files.xml
+++ b/developer-resources/License-Example-Files/License_for_POM_Files.xml
@@ -10,7 +10,7 @@
     <parent>
         <artifactId>egeria</artifactId>
         <groupId>org.odpi.egeria</groupId>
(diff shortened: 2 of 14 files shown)


planetf1:
At an appropriate time, for the current platform model, the egeria Dockerfile should be updated to use the new name/structure. The charts using the platform may not need changes, but should be checked.

PR #7999, opened 2024-01-10
Survey Action Framework (SAF)

mandy-chessell:
## Description

This is the initial implementation of the Survey Action Framework (SAF) with an implementation of three survey action services.  The aim is to share the new programming model with examples.

## Related Issue(s)

None

## Testing

Just regression tests with FVT.   These services can not run until the Survey Action OMES and the SAF client in Stewardship Action OMES is written.

## Additional notes

The new survey action services have been included in the OpenConnectorArchive.omarchive builder.  However they are not able to run so the archive file has not been updated in content-packs.


mandy-chessell:
There is also a fix to the XTDB connector to build and use a Fat Jar in the assemblies.

PR #8448, opened 2024-10-15
Revert "Bump org.hamcrest:hamcrest from 2.2 to 3.0"

dwolfson:
Reverts odpi/egeria#8361


dwolfson:
Reverting to older version.

PR #8587, opened 2025-01-20
Add Information Supply Chain content to CocoComboArchive

mandy-chessell:
## Description

This PR extends the metadata found in the CocoComboArchive Open Metadata Archive to include metadata about the clinical trial treatment validation information supply chain.

A skeleton Solution Architect OMVS module is also added to the platform and configured in the view server.

## Related Issue(s)

One of the Coco governance services was not processing a null response from getMetadataRelatonships correctly - now fixed.

## Testing

Manual testing of the new archive

A matching PR for egeria-docs is in progress

## Additional notes

The next step is to build out the queries in Solution Architect OMVS to visualize the information supply chain.


mandy-chessell:
This is the diagram of the new information supply chain - and the solution components that are supporting it.

image

PR #8778, opened 2025-08-04
Radical simplification of platform

mandy-chessell:
## Description

As the title suggests, this is a big change to Egeria's runtime.  Basically the following changes have been implemented:

* Removal of all Open Metadata Access Services (OMASs)
* Removal of all Open Metadata Integration Services (OMIS)
* Removal of the XTDB and JanusGraph repositories
* Move of framework services so that they are now access services.
* Building of a new set of handlers for both the view services and the connectors
* Building new contexts for integration connectors (integration context), governance action services (governance context), survey action services (survey context) based on a new base context from the Open Metadata Framework.  This base context uses the new handlers.
* Removal of ConnectedAssetProperties and the wrapper beans for Connection, ConnectorType and Endpoint from the open connector framework.  This removes considerable duplication between OMF and OCF.
* Restructuring of the core open metadata types' attributes so that more of the properties are consolidated in the base types rather than scattered throughout the subtypes.  This has allowed greater simplificaiton in the basns and metadata elements serving the APIs.
* Restructuring of the request bodies around a hierarchy of options that are also used on the OMF interfaces, reducing maintenance on the parameter lists as new functions are added in the future.  It has also meant that new features are automatically implemented consistently across all APIs because they use the same structures.

The purpose of these changes is to reduce the repetition in the platform and create a more consistent REST API and connector interfaces.  The connectors are also given access to the mermaid graphs and richer elements on retrieval.

## Related Issue(s)

There are still a lot of files to remove from the runtime.  There is also outstanding consolidation work around the metadata elements and mermaid graph classes.

## Testing

Simple testing at this point.  The aim is to make the new APIs available to the rest of the community to experiment with and update clients such as PyEGERIA.

The FVTs are disabled at this point since they were based on the original OMASs.  They need a complete rethink.

This has a big impact on the documentation which will be updated over the next few weeks.  There is an initial update covering the changes to the type system in progress.

## Additional notes

This is the preliminary change to the runtime for version 6.0 - There will be no version 5.4.


mandy-chessell:
## Usage Notes

Since the XTDB repository is no longer available, the repositories in simple-metadata-store and active-metadata-store are now both the in-memory repository implementation.  

These servers have new configuration to used the new access services.

## Workspaces
The repository in qs-metadata-store remains as postgres.  However it is necessary to delete and reconfigure this server because the access services have changed.  The changes to the type system means that you should delete your postgreSQL repository data since there are many changes to the type system that are not backward compatible.

PR #9194, opened 2026-08-07
Add additionalConnectionProperties config property to JDBCResourceConnector

### LINKED ISSUE #9193: JDBCResourceConnector has no way to pass extra JDBC connection properties (breaks Oracle REMARKS/comments)
### Description

`JDBCResourceConnector` (and the JDBC integration connector that uses it) opens connections via `DriverManager.getConnection(url)` or `DriverManager.getConnection(url, username, password)` only — there's no configuration property to pass additional, driver-specific JDBC connection properties through `DriverManager.getConnection(url, Properties)`.

This is a real problem for Oracle specifically: Oracle's JDBC thin driver silently omits `REMARKS` (table and column comments/descriptions) from all `DatabaseMetaData` calls (`getTables()`, `getColumns()`) unless the connection property `remarksReporting=true` is explicitly set. Since this connector never sets it and provides no way for a caller to set it either, every table and column description harvested from an Oracle database into Egeria's metadata catalog is silently empty — no error, no warning, just missing data.

This isn't Oracle-only in principle — other drivers have their own connection-property quirks (e.g. timezone handling, fetch size defaults) — so a generic `additionalConnectionProperties` configuration property (a `Map<String,String>` passed straight through to the driver) would fix this for Oracle and give an escape hatch for other drivers too, without hardcoding vendor-specific logic into what's meant to be a generic connector.

### Environment
Module: `open-metadata-implementation/adapters/open-connectors/data-store-connectors/jdbc-resource-connector`

### Do you plan to make this contribution yourself?
Yes - will follow up with a PR.

karthikchundi-commits:
## Description

There was no way to pass extra JDBC connection properties through to `DriverManager.getConnection()` - only the URL and, optionally, username/password. That breaks Oracle specifically: its JDBC driver silently drops table/column `REMARKS` (comments) from `DatabaseMetaData` calls unless `remarksReporting=true` is set on the connection, and there was no way to set it, so every table/column description harvested from an Oracle source was silently empty.

Adds an `additionalConnectionProperties` configuration property (a `Map<String,String>`) that gets merged into the `Properties` object passed to `DriverManager.getConnection(url, properties)` on every connection. This also lets the three separate `getConnection()` call sites (url-only, url+configured user/password, explicit username/password override) consolidate onto that one `Properties`-based `DriverManager` overload instead of branching between three different overloads - passing an empty `Properties` is equivalent to passing none.

The merge logic (`additionalConnectionProperties` -> `Properties`) lives in a new small `JDBCConnectionPropertiesUtil` class kept deliberately free of any Egeria framework dependency, so it can be unit tested directly without needing the rest of the connector framework/a live database on the classpath.

## Related Issue(s)

Closes #9193

## Testing

- `./gradlew :...:jdbc-resource-connector:test` - all tests pass (6 new tests covering the merge logic: valid map, non-map/null value, empty map, null keys/values skipped, non-string values converted, existing properties preserved)
- `./gradlew :...:jdbc-resource-connector:compileJava` - compiles clean
- `./gradlew :...:jdbc-resource-connector:checkstyleMain checkstyleTest` - passes clean
- No existing tests touch connection-establishment logic (there's no embedded/mock DB in this module currently), so I didn't add a live-connection integration test - the new unit tests exercise the actual logic that changes driver behavior (the properties merge) without needing one.

Worth a mention in release notes: a new `additionalConnectionProperties` config property is available on the JDBC resource/integration connectors, with `remarksReporting=true` as the documented example for Oracle users who want table/column comments captured.

## Additional notes

Disclosing per good practice even though there's no checkbox for it in this repo's template: I used AI (Claude) to help investigate this gap and implement/test the fix, under my direction and review.


mandy-chessell (approved):
Thank you for submitting this change.  I see you are working on Oracle.  I am in the process of creating new database integration connectors and surveyors - Just done MS SQL server.  Would you be interested in a set of similar connectors for Oracle databases?

karthikchundi-commits:
Thanks for the review! Yes, I'd be very interested in building a set of Oracle connectors/surveyors along the same lines as the MS SQL Server ones - Oracle is my main background (15+ years working with Oracle ERP Cloud, EBS, and database architecture), so this is a good fit. I'll start by reading through the Postgres and MS SQL Server surveyor/cataloguer code to understand the pattern, and I have access to an Oracle instance for testing. Will follow up here or open a new issue to track scope once I have a plan together.

mandy-chessell:
Here is a link to the docs:
* https://egeria-project.org/egeria-solutions/leveraging-postgres/overview/
* https://egeria-project.org/egeria-solutions/leveraging-mssql/overview/

mandy-chessell:
Here is the Oracle DB support ... https://egeria-project.org/egeria-solutions/leveraging-oracle/overview/

karthikchundi-commits:
Looks like you moved faster than I could get started - thanks for crediting the additionalConnectionProperties piece in #9196, glad it was useful for the remarksReporting requirement!

Is there anything specific still worth building here, on Oracle or elsewhere? Happy to take on another database/connector if there's one you haven't gotten to yet, or anything else that would be useful.

mandy-chessell:
I would look at what we have and look for ways it can be improved, or tools that you have in your environment that we don't cover.  It would be wonderful to know how you found us and what you are looking to do with Egeria.  We may be able to help, and all use cases are valuable.

karthikchundi-commits:
Happy to share the story - I originally reached out to LF AI & Data Foundation about getting involved as a contributor, and Nathan Southern pointed me toward the working groups. I ended up gravitating toward the Data Lineage side (OpenLineage) since it matched my Oracle/enterprise-data background, and while working there I found a few Oracle-specific gaps that generic contributors hadn't caught - the TNS connection format issue in OpenLineage's Oracle extractor, and then the additionalConnectionProperties/remarksReporting issue here. That's what brought me to Egeria.

On what I'm looking to do: I'm a Staff Software Engineer with 15+ years in Oracle ERP Cloud, EBS, and database architecture, and I'm looking to bring that depth to open metadata/governance tooling where Oracle tends to be a second-class citizen behind Postgres/MSSQL - which is exactly the pattern I keep finding.

On tools in my environment you don't cover yet: Oracle Integration Cloud (OIC) is a real gap - I searched the repo and found zero references to it. OIC is Oracle's integration/orchestration platform, commonly deployed alongside Oracle ERP Cloud and EBS in the same enterprise environments Egeria's Postgres/MSSQL/Oracle database connectors already target. It has its own metadata surface worth cataloguing - integration flows, connections, lookups, and the systems they connect - that's genuinely distinct from database-level surveying. Happy to scope what an OIC connector/surveyor would look like if that's useful, or take on database improvements first if that's a better place to start. Your call on what's most valuable.

karthikchundi-commits:
Following up in case this fell off your radar - no pressure either way, just wanted to check whether OIC is worth pursuing or if the DB-connector side is still the priority. If it's useful, I'm happy to just put together a short scoping writeup (metadata surface, connector/surveyor shape) for OIC so there's something concrete to react to, rather than waiting on more back-and-forth first. Let me know what's most helpful.

karthikchundi-commits:
## OIC (Oracle Integration Cloud) connector — scoping writeup

As promised — here's a concrete proposal to react to instead of more back-and-forth. I read through the existing `oracle-server-connectors` and `postgres-server-connectors` modules you built, plus `unity-catalog-connectors`, to ground this in the actual current codebase rather than guess at conventions.

### What OIC actually is (and why it's not "just another database")

Oracle Integration Cloud is Oracle's iPaaS — integration/orchestration flows between systems, commonly deployed right alongside Oracle ERP Cloud/EBS in the same shops Egeria's existing Oracle DB connector already targets. Its real metadata surface is:

- **Integrations** — the core design-time artifact: name, code, version, integration style (App-Driven, Scheduled, Orchestration), activation status, associated Connections
- **Connections** — configured adapter instances (REST, SOAP, FTP, ERP, Database, etc. adapters), each with its own security config
- **Lookups** — cross-reference tables used inside integrations
- **Packages** — deployable bundles of integrations + related artifacts, used to move content between environments
- **Connectivity Agents** — for on-prem connectivity via agent groups

None of this maps to a JDBC `ResultSet`. It's a genuinely different metadata shape from anything the database-connector family catalogs today — closer to "cataloging a fleet of ETL/orchestration jobs and their wiring" than "cataloging tables and columns."

### The bigger architectural point: this should NOT mirror the JDBC connector pattern

`oracle-server-connectors`, `postgres-server-connectors`, `mssql-server-connectors` etc. all share one shape because they're all fundamentally `DataSource`/`ResultSet`-based (`JDBCResourceConnector`, `OracleConfigurationProperty`'s `HOST_IDENTIFIER`/`PORT_NUMBER`/username-password model, `OracleServerSurveyActionService` querying system catalog views).

OIC has none of that. It's queried entirely over REST, authenticated via **OAuth2 client-credentials grant against Oracle IDCS or OCI IAM** (bearer token) — not username/password at all. That's actually the single biggest deviation from the whole existing DB-connector family's credential model, and I think it's worth flagging up front rather than discovering halfway through a PR.

Reading `unity-catalog-connectors`, I think **that's the right structural sibling to mirror instead** — it's the one existing connector in this family that's already REST-based rather than JDBC-based:

```
unity-catalog-connectors/
  controls/     - config properties, placeholders, template types   (mirror directly)
  ffdc/         - audit/error codes                                  (mirror directly)
  properties/   - typed domain model for what UC returns             (mirror directly)
  resource/     - the actual REST client (OSSUnityCatalogResourceConnector)
  resource/rest/ - typed request/response DTOs per endpoint (ListCatalogsResponse, etc.)
```

Proposed `oic-connectors` module, same shape:

```
oic-connectors/
  catalog/      - OICIntegrationConnector (DynamicIntegrationConnectorBase, same pattern as
                  OracleServerIntegrationConnector) + catalog target processor
  controls/     - OICConfigurationProperty / OICPlaceholderProperty (host, instance URL,
                  OAuth token endpoint, client ID/secret placeholders) + OICTemplateType
  ffdc/         - OICAuditCode / OICErrorCode
  properties/   - IntegrationInfo, ConnectionInfo, LookupInfo, PackageInfo domain classes
  resource/     - OICResourceConnector (OAuth2 token acquisition + REST calls against the
                  Integration Instance Management API)
  resource/rest/ - ListIntegrationsResponse, ListConnectionsResponse, etc.
```

### Proposed first increment (don't try to do all of it at once)

Matching how the DB connectors were built one backend at a time, I'd scope v1 to:

1. `OICResourceConnector` — OAuth2 client-credentials token acquisition + a thin REST client for exactly one endpoint: list Integrations for an OIC instance.
2. `OICIntegrationConnector` (catalog connector) — turns that into open-metadata elements: one asset per Integration, capturing name/code/version/style/status.
3. Config/placeholder/template/ffdc classes to support just that.
4. Skip Connections, Lookups, Packages, and the survey/stats side entirely for v1 — those are natural, self-contained follow-up PRs once the base shape is agreed, same incremental pattern as the existing connectors.

### Open questions I don't have a confident answer to

- **Auth model fit**: does Egeria's connector framework already have a clean pattern for OAuth2 client-credentials elsewhere I should mirror (rather than rolling something OIC-specific)? I didn't find one in the modules I checked, but I may have missed it.
- **Core vs. plugin**: same fork-in-the-road as was raised for the MLflow OCI proposal — does a new integration platform (as opposed to a new database backend) belong in core Egeria, or is this better suited as a separate plugin/connector repo?
- **Auth for the agent/on-prem case**: OIC deployments using Connectivity Agents for on-prem systems have a different topology than pure cloud-to-cloud — not sure if that's in scope for v1 or a later concern.

### Rough effort estimate

v1 as scoped above (list-integrations-only) feels like a similarly-sized unit of work to `additionalConnectionProperties` + the TNS fix I did on OpenLineage, i.e. a few focused days, not a multi-week undertaking — mainly because I'm deliberately not touching Connections/Lookups/Packages/survey yet.

Let me know if this direction looks right, or if you'd steer it differently before I start on anything.

mandy-chessell:
Nice write up ... this discussion may be better in an issue rather than a merged PR, then it is more visible.

So, the proposal looks fine.  I agree that the UC pattern is closer to what you are looking for since it is a metadata catalog.  One thing to think of is metadata flow direction.  The UC connectors support flow in 2 directions.  UC metadata can flow into Egeria, and Egeria can provision new entries in UC's catalog.  The connector needs to know where an element originates in to know which way it is going in.   It looks like lineage capture is going to be key to the design too.

The outbound authorization uses a client-side secret (https://egeria-project.org/concepts/client-side-secret/)

My main question would be one of testing.  We are an open source project with no budget to pay for software licenses.  This is why we tend to favour integrating with other open source software.  This does not mean we don't integrate with vendor software.  The fact that we have integrated with MS SQL and Oracle DB is evidence for that.  We are able to test the support because the vendors offer free licenses for testing.  Is that also true for OIC? 

I am planning of building some FVTs for the existing connectors.  The idea is that these tests are not integrated into the build, but can be run at need, when the required target system is available (see under open-metadata-test module) - this might be a route for OIC, particularly if there is no public, free copy.  We could document when and the versio nagainst which the last test was run so any consumer is aware that they should run the FVT  for their environment (they probably will anyway :)).

mandy-chessell:
Sorry - I meant to include this link - description of the UC connectors https://egeria-project.org/egeria-solutions/leveraging-unity-catalog/overview/

Also, the Egeria community calls begin again next week after our summer break.  Call in details are here: https://lists.lfaidata.foundation/g/egeria-technical-discuss/viewevent?eventid=3224675

karthikchundi-commits:
Thanks for the detailed feedback - a few responses:

**Testing access:** Being upfront - I don't have OIC access or a license for testing, and won't have going forward either. Rather than build against something I can't actually verify against, I'd like to propose pivoting the target: instead of Oracle Integration Cloud specifically, build this connector against an open-source iPaaS platform with the same architectural shape - I'm thinking **WSO2 Micro Integrator** (Apache-2.0, self-hostable indefinitely, free). It's REST-managed, has integrations/connectors/mediation flows as first-class concepts, and gives a real API surface to query "what integrations exist" - the same core shape as OIC, just something I (or anyone else) can actually stand up and test against without needing an enterprise license. Happy to go a different open-source direction if you have a preferred target in mind, but this keeps the design and the "why UC-pattern, not JDBC-pattern" reasoning from my last writeup fully intact.

**Metadata flow direction:** Good catch - I hadn't thought through the bidirectional case. For a v1 scoped to just listing Integrations, I think that's read-only/inbound-only (Egeria discovering what already exists), and provisioning-out would be a clearly separate, later capability rather than something v1 needs to handle. Let me know if that's the wrong instinct.

**Client-side secret:** Thanks for the pointer - that answers my auth-pattern question, I'll build against that instead of rolling something new.

**Moving to an issue:** Agreed, happy to open one and move this discussion there for visibility - I'll do that and link back here.

**Community call:** Thanks for the invite, I'll try to join once it resumes.

karthikchundi-commits:
Moved this to a dedicated issue for visibility: #9245

PR #9220, opened 2026-08-17
Externalize the JDBC Pool Size

mandy-chessell:
## Size the JDBC connection pool for the role the connector is playing

Follow-on to the HikariCP pooling change (`7359136cb1`). That commit introduced `jdbcMaximumPoolSize` with a default of 5. The default was reasoned about wrongly, and there was no way for a hosting connector to override it.

### The mistake in the original default

The default of 5 was justified by per-catalog-target arithmetic: a pool exists per connector instance, Egeria creates one instance per catalog target, so twenty targets at five connections is a hundred connections against the server - keep it small.

That reasoning does not apply to the case that matters most. When a `JDBCResourceConnector` is embedded in a repository or an audit log destination, there is exactly **one** instance and it is the single route to the database for every request the server handles. It should be sized for the server's concurrency, not shrunk to guard against a multiplication that never happens in that role. One default cannot serve both.

Measured on a running quickstart platform, sampling `pg_stat_activity` every two seconds through startup (192 samples):

```
max total = 5    max active = 4
```

The pool sat at its ceiling of 5 for most of startup with active peaking at 4. It coped, but with one connection of headroom on the busiest workload the platform runs.

### Changes

**Default raised from 5 to 10.** Double the observed peak, and still small enough to respect HikariCP's own guidance that pools should be sized around cores rather than around servlet thread count.

**`JDBCResourceConnector.setConfigurationProperty(name, value)`** - a hosting connector can pass a configuration property down to the connector it embeds, before `start()` builds the pool. Passing null changes nothing.

**`jdbcMaximumPoolSize` added to `PostgresConfigurationProperty`** under the same name, with `PostgresOMRSRepositoryConnector` passing it down. The same wiring is added to `PostgreSQLAuditLogDestinationConnector`, which already shares the JDBC enum. Both are single-instance, whole-server connectors, so both want to be tunable where they are configured.

### Precedence: the host wins

A value passed down by the host **replaces** whatever is on the embedded connection.

This is deliberate and worth understanding before reviewing. The embedded connection is normally generated from a template or content pack that a deployment cannot easily edit, so its value is effectively a default. The hosting connector's configuration properties are where a deployment is actually tuned. If the template's value is wrong for a particular environment, the operator needs to be able to override it without editing the template.

There is a subtlety that makes this more than a style preference. `ConnectorConfigurationFactory` sets the **same `storageProperties` map** on both the host and the embedded JDBC connection:

```java
jdbcResourceConnection.setConfigurationProperties(storageProperties);
```

Under the opposite rule - embedded connection wins - the pass-through would find the property already present and do nothing, making it inert in exactly the configuration Egeria generates for itself.

### Tests

Four unit tests in `JDBCResourceConnectorTest` cover the precedence rule. They construct the two sources independently, so they cannot be satisfied by the shared-map path above:

| Test | Asserts |
|---|---|
| `testHostValueOverridesTheConnectionValue` | connection says 3, host passes 17 → 17 |
| `testNullFromHostLeavesTheConnectionValueAlone` | connection says 3, host passes null → 3 |
| `testHostValueAppliesWhenConnectionHasNoProperties` | no properties at all → 17 |
| `testOtherConnectionPropertiesAreUntouched` | `databaseName` survives, map stays size 2 |

`JDBCResourceConnectorTest` passes at 10 tests, 0 failures. `query-fvt` passes at 22 tests, 0 failures against a real PostgreSQL server.

This required adding `open-connector-framework` and `open-metadata-framework` as `testImplementation` in the connector's `build.gradle` - they were `compileOnly`, so the test source set could not see `Connection` or `ConnectorBase`. Test scope only; the shadow jar is unaffected.

### Documentation

The README gains a *Choosing a pool size* section that splits the two roles - embedded-in-a-server versus created-per-catalog-target - with a JSON example of setting the value on the host. The previous blanket statement that `jdbcMaximumPoolSize` is "multiplied by the number of catalog targets" has been removed: it was only ever true for one of the two roles, and stating it unconditionally is what led to the repository pool being under-sized in the first place.

### Not caused by this, for the record

While investigating, a run of `HttpConnectTimeoutException` errors on the platform's own `host.docker.internal:9443` calls was checked as a possible pool bottleneck and **ruled out**. Sampling the pool across the exact windows when those connects were timing out showed 0-2 active connections with idle connections available in every sample, no HikariCP pool-wait timeouts anywhere in the log, and a TCP-connect failure mode rather than the response-timeout that blocked worker threads would produce. It is a Docker Desktop hairpin issue, unrelated to the database.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mandy-chessell:
## Fix buildHealth failure: expose open-metadata-framework transitively from open-connector-framework

### Problem

`buildHealth` was failing on `jdbc-resource-connector` with:

```
Advice for :open-metadata-implementation:adapters:open-connectors:data-store-connectors:jdbc-resource-connector
Unused dependencies which should be removed:
  testImplementation project(':open-metadata-implementation:frameworks:open-connector-framework')
  testImplementation project(':open-metadata-implementation:frameworks:open-metadata-framework')
```

The `open-metadata-framework` (OMF) declaration looked unused because nothing in the test source imports it directly — but removing it breaks compilation. `ConnectorCheckedException` (from `open-connector-framework`, which *is* directly imported in the test) extends `OMFCheckedExceptionBase`, a class defined in OMF. Resolving that exception's supertype chain requires OMF on the compile classpath, even though no file imports it by name.

### Root cause

`open-connector-framework`'s own `build.gradle` declared its dependency on `open-metadata-framework` (and `audit-log-framework`) as `implementation`, which deliberately hides a dependency from consumers' compile classpaths. Since OCF's public API (its exception hierarchy, and `AuditLog` in its connector base class constructors) genuinely exposes types from both, every direct consumer that touches that API had to work around the gap with a redundant explicit re-declaration — exactly the pattern the dependency-analysis tool was (correctly) flagging as unnecessary duplication, while still (incorrectly) recommending outright removal.

Grepping for the same `testImplementation`-on-both pattern found 4 modules relying on this transitive-but-undeclared relationship with no direct OMF import in their test source, all latent buildHealth failures waiting to happen: `jdbc-resource-connector`, `data-folder-connector`, `cohort-registry-file-store-connector`, `inmemory-repository-connector`.

### Fix

- **`open-connector-framework/build.gradle`**: added the `java-library` plugin (this repo applies the plain `java` plugin to all subprojects by default, which doesn't expose the `api` configuration) and changed `audit-log-framework` and `open-metadata-framework` from `implementation` to `api`, since both are genuinely part of OCF's public API surface. Also moved `jackson-databind` to `runtimeOnly` (verified `ObjectMapper` isn't referenced directly in OCF's own source — an existing, unrelated buildHealth finding that opting into `java-library` also surfaced).
- **`jdbc-resource-connector`, `data-folder-connector`, `cohort-registry-file-store-connector`, `inmemory-repository-connector`**: removed the now-redundant explicit `open-metadata-framework` declarations, since it's correctly available transitively via `open-connector-framework` now.

### Verification

- `./gradlew buildHealth` passes clean (empty report) across the full repo.
- All 5 modules build, checkstyle, and (where applicable) test successfully in isolation and together.

### Note for reviewers

While verifying this, found that `JDBCResourceConnectorTest` — the test class whose comment originally justified the `testImplementation` (“the connector lifecycle tests instantiate the connector”) — is never actually discovered/run by Gradle's TestNG integration (`--tests` filtering returns "no tests found for given includes"), regardless of dependency configuration. That's a pre-existing gap unrelated to this fix; flagging separately since it means this module's connector-lifecycle test coverage currently isn't exercised in CI at all.

PR #9244, opened 2026-08-26
Fix PostgresDatabaseStatsExtractor misreading negative pg_stats.n_distinct as a raw count

karthikchundi-commits:
## Description

`PostgresDatabaseStatsExtractor.getSchemaStatistics()` reads `pg_stats.n_distinct` directly into a `long` via `ResultSet.getLong("n_distinct")` and stores it as the survey's "Number of Distinct Values" column property.

Per PostgreSQL's own documentation for the `pg_stats` catalog view (https://www.postgresql.org/docs/current/view-pg-stats.html), `n_distinct` is **not always an absolute count**:
- A non-negative value is the estimated number of distinct values, used directly.
- A **negative** value is the negative of the ratio of distinct values to rows in the table - PostgreSQL uses this for columns (such as primary keys, UUIDs, and timestamps) where the number of distinct values scales with the size of the table, since a fixed absolute estimate would go stale as the table grows.

Reading the raw catalog value directly as a count produces a nonsensical negative result in the survey property - or, once truncated by `getLong()` on a fractional negative value (e.g. `-0.95`), silently becomes `0` - for exactly these common, typically near-unique columns.

## Fix

- Added `pg_stat_user_tables.n_live_tup` (the table's own live row-count estimate) to the existing column-stats query, joined the same way `pg_stat_user_tables` is already joined elsewhere in this file for the table-level query.
- Extracted the calculation into a small, independently-testable static method, `calculateNumberOfDistinctValues(double rawDistinctStat, long tableRowCount)`, that scales the ratio by the row count when a negative `n_distinct` is reported, matching PostgreSQL's documented semantics; passes the non-negative value through unchanged otherwise.

## Testing

Added `PostgresDatabaseStatsExtractorTest` (TestNG) covering: a non-negative `n_distinct` passed through directly, a negative ratio correctly scaled by row count (including the near-unique -1.0 case), a small fractional negative ratio that would previously have truncated to zero via `getLong()`, and a zero-row-count edge case.

While verifying this, found that `postgres-server-connectors/build.gradle` was missing the `test { useTestNG() } ` block present in sibling connector modules (e.g. `jdbc-resource-connector`) - without it, `gradle test` silently finds and runs nothing for this module. Added it. Confirmed the difference directly: `gradle test` for this module found 0 tests before this change and 6 passing (the 2 pre-existing `AuditCodeTest`/`ErrorCodeTest` + 4 new) after.

Ran locally (JDK 17, Gradle):
- `compileJava` - clean
- `test` - 6/6 passing
- `checkstyleMain` / `checkstyleTest` - clean

I don't have a live PostgreSQL instance in this environment to run the full `getSchemaStatistics()` method end-to-end against a real database, so verification of the SQL query change itself is by careful review against the existing, already-working join pattern in the same file (`pg_tablesSQLCommand` already joins `pg_stat_user_tables` the same way), rather than a live run - happy to have a maintainer or CI confirm the query against a real instance.

## Additional notes

Generated with the assistance of Claude Code, reviewed and directed by me.

### DIFF
--- a/open-metadata-implementation/adapters/open-connectors/data-manager-connectors/postgres-server-connectors/build.gradle
+++ b/open-metadata-implementation/adapters/open-connectors/data-manager-connectors/postgres-server-connectors/build.gradle
@@ -31,6 +31,12 @@ dependencies {
     testImplementation 'org.testng:testng'
 }
 
+test {
+    useTestNG()
+    // This will default to standard search pattern - see https://docs.gradle.org/current/userguide/java_testing.html#sec:test_detection
+    scanForTestClasses = false
+}
+
 shadowJar {
     mainClassName = ''
     archiveClassifier = 'jar-with-dependencies'
--- a/open-metadata-implementation/adapters/open-connectors/data-manager-connectors/postgres-server-connectors/src/main/java/org/odpi/openmetadata/adapters/connectors/postgres/survey/PostgresDatabaseStatsExtractor.java
+++ b/open-metadata-implementation/adapters/open-connectors/data-manager-connectors/postgres-server-connectors/src/main/java/org/odpi/openmetadata/adapters/connectors/postgres/survey/PostgresDatabaseStatsExtractor.java
@@ -39,6 +39,31 @@ public class PostgresDatabaseStatsExtractor
 
 
     /**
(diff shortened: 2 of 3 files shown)


mandy-chessell (approved):
Thank you for both finding and fixing this

PR #9264, opened 2026-08-30
Import the Spring Boot BOM instead of hand-pinning Spring versions 

mandy-chessell:
# Import the Spring Boot BOM instead of hand-pinning Spring versions (stage 4a)

## Summary

`bom/build.gradle` now imports `spring-boot-dependencies` as a platform, and the 26 constraints Boot
already manages are deleted. One file: 20 insertions, 34 deletions. Third-party constraints drop from 81
to 54.

**Spring Boot stays at 3.1.4.** Nothing here upgrades Spring. That is stage 4b, and keeping the two apart
is the point: this PR changes *how* versions are decided, so if something behaves differently afterwards
it is attributable to the restructure and not to a version bump.

## Why

Every Spring version was pinned by hand, and the pins had drifted out of step with each other: Boot 3.1.4
alongside Spring Framework 6.0.6 and Spring Data 3.0.3, which belong to different Boot releases. Nothing
kept them consistent, so each dependabot bump had to be reasoned about individually and the set could
silently disagree.

With the BOM imported, one `springbootVersion` decides the whole family and they cannot disagree again.
It is also what turns the 4b upgrade from eight coordinated version bumps into one.

## What changed

- Imported `api(platform("org.springframework.boot:spring-boot-dependencies:${springbootVersion}"))`.
- Restored `javaPlatform { allowDependencies() }`, which a platform import requires. It had been removed
  in the earlier BOM cleanup along with the unused Chronicle platform. The comment above it warns that it
  also disables the plugin's check that only constraints are declared here.
- Deleted **26 constraints**: 25 Spring artifacts (boot, security, framework, data) and
  `tomcat-embed-core`.
- Deleted 4 version variables left with nothing to name: `springdataVersion`, `springsecurityVersion`,
  `springwebVersion`, `tomcatVersion`.
- Deleted the `mockito` pin - see below.

Egeria's remaining constraints still win wherever they name a **higher** version than Boot does, because
Gradle resolves conflicts by taking the highest. The pins deliberately held ahead of Boot to pick up CVE
fixes are therefore unaffected.

## This does change versions, and that is the point

The plan described 4a as "restructure, no version change". That was wrong in a way worth correcting:
importing the BOM necessarily aligns versions, because removing drift *means* moving the outliers onto a
single version. What stays fixed is the Boot version itself.

The full resolved dependency graph was captured before and after - every configuration of every module,
around 44,000 entries each time. Every single difference collapses a duplicate:

| Artifact | Before | After |
|---|---|---|
| `spring-core`, `spring-beans`, `spring-jcl` | 6.0.6 **and** 6.0.12 | 6.0.12 |
| `spring-web` | 6.0.6, 6.0.11 **and** 6.0.12 | 6.0.12 |
| `spring-ldap-core` | 3.0.5 | 3.1.1 |
| `micrometer-commons` / `micrometer-observation` | four versions in play | two |
| `mockito-core` | 4.11.0 **and** 5.3.1 | 5.3.1 |
| `byte-buddy`, `httpcore`, `jboss-logging`, `slf4j` | mixed | Boot's |
| `jakarta.activation`, `hamcrest-core` 1.3 | on the classpath | gone |

Nothing moved to a different major version, and no Spring artifact moved outside the 6.0.x line that Boot
3.1.4 ships.

## The mockito pin

Boot's BOM manages mockito at 5.3.1, which is higher than Egeria's 4.11.0 pin - so Gradle was already
taking 5.3.1 in some configurations, and after this change takes it everywhere. The pin had stopped
deciding anything while still claiming 4.11.0, and its comment read `// TODO: Held as data engine breaks`.

That rationale is stale. Only two modules use mockito - `http-helper` and `inmemory-repository-connector`
- and no data-engine module does. Both ran their tests and passed on 5.3.1. The pin and its comment are
removed rather than left asserting something untrue.

## Verification

- `./gradlew clean build` green, including the BVT.
- **auth-fvt: 15 tests, 0 failures.** This is the suite added in stage 3b specifically so that a change
  touching Spring could be checked against the real security filter chain rather than a permit-all one.
- **query-fvt: 34 tests, 0 failures** against a real PostgreSQL-backed platform.
- `spring-boot:3.1.4 (by constraint)` confirmed on the platform chassis after the change.

## Notes for reviewers

- **`allowDependencies()` is back, so watch the dependencies block.** It disables the java-platform
  plugin's guard against declaring real dependencies there. Only constraints and the platform import
  belong in it.
- **Three pins are pre-existing lies, left alone deliberately.** `jackson-core` and
  `jackson-dataformat-yaml` pin 2.15.0 while `jackson-bom` 2.22.1 forces 2.22.1, and `slf4j-api` pins
  2.0.6 while 2.0.18 resolves. All three were already inert before this PR and none is Spring's business -
  a jackson/slf4j tangle worth its own look rather than something to fold in here.
- The BOM is published for downstream users. Consumers who relied on Egeria pinning a specific Spring
  patch version will now get whatever Boot 3.1.4 manages, which is the more predictable answer but is a
  change in what the published BOM says.


mandy-chessell:
# Upgrade Spring Boot to 3.5.16, stop the FVT suites binding fixed ports, and run 20 tests that never ran (stage 4b)

## Summary

Three related changes. The third only came to light because the first restored a build policy that had been
relaxed since stage 2.

1. **Spring Boot 3.1.4 → 3.5.16**, walked one minor at a time with the full test suite run at each hop.
   springdoc moves with it, 2.2.0 → 2.8.17, and `org.gradle.warning.mode` goes back to `fail`.
2. **The eight FVT suites no longer bind fixed ports.** Each allocates a free one at runtime.
3. **Eleven modules had test sources that never ran.** Enabling them exposed six real defects in FFDC
   message definitions that had never been validated.

Boot 3.1 has been out of support for a long time, so this is the stage of the 6.2 toolchain work with
actual security value. It is also the only stage that changes what the platform does at runtime rather
than how it is built.

---

# Part 1 - Spring Boot 3.1.4 to 3.5.16

## Hop by hop

Each hop moved `springbootVersion` in the BOM and the Boot Gradle plugin version in `settings.gradle`
together - they must stay in step - and was verified before the next one started.

| Hop | Boot | Build | auth-fvt | query-fvt |
|---|---|---|---|---|
| 1 | 3.2.12 | green | 15/15 | 34/34 |
| 2 | 3.3.13 | green | 15/15 | 34/34 |
| 3 | 3.4.13 | green | 15/15 | 34/34 |
| 4 | 3.5.16 | green | 15/15 | 34/34 |

**No source changes were needed at any hop.** Stage 4a is why: with the Boot BOM imported, one version
number decides the whole Spring family, so each hop is a single edit rather than eight coordinated ones.

## springdoc - this would have shipped broken

Every test platform in the repository sets `springdoc.api-docs.enabled=false`, while the **shipped**
`application.properties` sets it to `true`. So springdoc was invisible to the entire test suite, and
nothing would have failed if it stopped working.

It had stopped working. springdoc 2.2.0 under Boot 3.5:

```
NoSuchMethodError: 'void org.springframework.web.method.ControllerAdviceBean.<init>(java.lang.Object)'
```

A hard startup failure of the API documentation in every deployed platform, with a perfectly green build.

- springdoc bumped to **2.8.17**, which serves a valid OpenAPI document under Boot 3.5.
- **The auth-fvt platform now leaves the API docs switched on**, and `publicEndpointsStayOpen` asserts that
  `/v3/api-docs` returns 200 with an actual OpenAPI document. Those endpoints are in `SecurityConfig`'s
  permitAll list, so this belongs in that suite regardless - and springdoc can no longer fall out of step
  unnoticed.

The failure was reproduced deliberately, by pinning 2.2.0 back and re-running, so the bump is known to be
necessary rather than precautionary.

## Stage 2's debt is repaid

`org.gradle.warning.mode` is back to **`fail`**. Stage 2 had to relax it because the Boot 3.1.4 Gradle
plugin called the deprecated `FileTreeElement.getMode()`; Boot 3.5's does not. The `cors()` and `jwt()`
deprecation warnings in `user-authn`'s `SecurityConfig` are gone too.

---

# Part 2 - FVT ports allocated at runtime

The eight FVT suites each bound a fixed HTTP port, 9446 through 9453. Running query-fvt while another
checkout was already running it produced `PortInUseException` and failed every test in the suite at
`initializationError` - seventeen failures, none related to what the suite tests. This happened in
practice during this work, with a Gradle test worker from a second clone holding 9446.

**The extension** now allocates a genuinely free port with `new ServerSocket(0)` and passes it in as
`server.port`. Every extension already derived `platformURLRoot` from the port actually bound, so nothing
downstream needed changing.

**`application.properties`** no longer pins a port, and `egeriaEndpoint` becomes
`http://localhost:${server.port}`. That second half is why this is not simply `server.port=0`:
`egeriaEndpoint` is substituted into `BasicServerProperties.defaultLocalServerURL` and becomes the
server's own `localServerURL`, so with port 0 it would have resolved to a literal `0` and left the server
advertising an address nothing listens on.

The BVT and auth-fvt were never affected - they already started on an ephemeral port.

---

# Part 3 - 20 tests that never ran

Restoring `warning.mode=fail` made two Gradle deprecations fatal:

```
No test executed. There are test sources present but no test was executed.
The automatic loading of test framework implementation dependencies has been deprecated.
```

**These do not appear in a normal local build.** With `org.gradle.caching=true` the test tasks are served
from the build cache, and a cached task emits no output; CI runs `cache-read-only`, so they actually
execute. Anyone verifying this class of change locally needs `--no-build-cache`.

Rather than chase them one failing build at a time, a static scan found every module with test sources but
no test framework configured. **Eleven modules, 20 test files, none of which had ever run:** the db2luw,
duckdb, mssql, oracle, unity-catalog, jdbc-integration, audit-log-postgres and governance-action
connectors, and the governance-action, repository-governance, survey-action and watchdog-action APIs.

This is the same latent bug stage 2 found in `omf-metadata-api`, repeated across twelve more modules.

## Six defects those tests immediately found

| Defect | Where |
|---|---|
| Duplicate system actions | `GovernanceActionConnectorsAuditCode` 0018/0033 and 0034/0035 |
| Duplicate system action | `DuckDBAuditCode` 0006/0007 |
| Trailing space in the message id | `RepositoryGovernanceErrorCode` 400-008 and 400-022 |
| Duplicate user actions | `JDBCIntegrationConnectorAuditCode` 0007/0008/0010 |
| Wrong object named | `UCAuditCode` 0011/0012 told the reader to update "the table" in the catalog and schema messages |
| Wrong component named | `GovernanceActionConnectorsAuditCode` 0007/0008/0022 told the reader to restart "the connector in the integration daemon" - these are governance action services |

The last two were not merely duplicated, they were wrong. Each message now says what actually happened
and what to do about it, and a concatenation-aware re-scan of all twelve modules reports no remaining
duplicates or malformed ids.

The five `messages-and-codes/*.md` files in the diff are the generated message documentation, regenerated
by the build from those edits.

---

## Verification

At the final state - Boot 3.5.16, springdoc 2.8.17, `warning.mode=fail`, runtime ports:

- `./gradlew clean build --scan --no-build-cache` - **the exact CI command, with a cold cache** - green,
  zero Gradle deprecations
- **auth-fvt: 15 tests, 0 failures**
- **query-fvt: 34 tests, 0 failures**, including a run deliberately made **while a second checkout still
  held port 9446** - the exact condition that previously failed the whole suite
- **BVT: 13 tests, 0 failures**
- The 20 newly-enabled tests all pass

auth-fvt is the suite added in stage 3b precisely so a Spring upgrade could be checked against the real
security filter chain rather than the permit-all one every other suite installs. It is what caught the
springdoc failure.

## One unexplained run

During hop 1, auth-fvt failed 12 of 15 tests, everything returning a plain 404 including endpoints that
need no authentication at all. Six subsequent runs - same sequence, same versions - all passed, and it has
not recurred.

Rather than assume it was a fluke, `OMAGPlatformExtension` now probes its own platform's origin endpoint
before any test runs and fails with a clear message if something other than the suite's platform is
answering on that port. If it happens again the report will say so directly, instead of twelve misleading
assertion failures.

I could not reproduce or explain it, and would rather record that than tidy it away.

## Notes for reviewers

- **This is the runtime-behaviour stage.** Everything before it in the 6.2 work was a build-tool change.
  Boot 3.1 to 3.5 crosses four Spring Framework and Spring Security minors, and a filter chain regression
  compiles perfectly - which is what the FVT evidence above is for.
- **Not every suite was executed.** query-fvt, auth-fvt and the BVT were. cts-fvt, postgres-fvt,
  duplicate-fvt, subscription-fvt, client-fvt, type-fvt and templates-fvt compile but have not been run
  since the port change. The edit is identical and mechanical in each, but that is not the same as having
  run them.
- **Trailing-space message ids exist elsewhere too** - in the Kafka topic connector and the REST client
  connector. Those modules' tests were already running and do not check for it, so they are untouched here
  and worth a separate look.
- The ports 9446-9453 are no longer reserved, and a suite's platform URL now differs between runs, so a
  port in an old log is not a constant to compare against.
- `user-authn`'s `SecurityConfig` still uses `httpSecurity.cors()` and `OAuth2ResourceServerConfigurer::jwt`.
  Both still exist in Spring Security 6.5 and their warnings have gone, but both have documented
  replacements and are worth revisiting before any move to Boot 4.
- Boot 4 is deliberately **not** attempted here. It brings Spring Framework 7 and a much larger surface of
  change, and wants its own release window.

PR #9309, opened 2026-09-15
Fix DB2LUWDatabaseStatsExtractor misreading COLCARD = -1 (ungathered stats) as a raw distinct-value count

karthikchundi-commits:
## Description

`DB2LUWDatabaseStatsExtractor.getSchemaStatistics()` reads Db2's `SYSSTAT.COLUMNS.COLCARD` directly into the survey's "Number of Distinct Values" measurement via `resultSet.getLong("ndistinct")`. Per Db2's own documentation, `COLCARD` is `-1` when `RUNSTATS` has not been run on the column — a documented "statistics not gathered" sentinel, not a real negative count. A column without gathered statistics is a routine, common state (not a rare edge case), so this fired constantly, silently reporting `-1` distinct values as though it were a real measurement.

This is the same failure class already fixed for PostgreSQL's equivalent `pg_stats.n_distinct` sentinel in `PostgresDatabaseStatsExtractor` (#9244) — found via the same sibling-comparison method, since this file's own header comment already draws the explicit parallel to Postgres's `pg_stats`. Nobody closed the loop on the Db2 side when that fix landed.

**Fix:** extracted a testable static `calculateNumberOfDistinctValues(long)`, following the same pattern as `PostgresDatabaseStatsExtractor.calculateNumberOfDistinctValues(double, long)`. The fix here is simpler than Postgres's ratio-scaling case — `COLCARD`'s sentinel needs no scaling, just recognition: treat it as "no data" (`0`, the framework's own default for this field) rather than a literal negative count.

**Also checked, no gap:** the other `*-server-connectors` modules (mssql, oracle, duckdb, unity-catalog) for the same sentinel-flowing-through-unvalidated pattern. `mssql-server-connectors` computes its distinct estimate from histogram steps rather than a raw vendor cardinality column; `duckdb-connectors` has no equivalent stats-extraction code path; `unity-catalog-connectors` is REST-based with no raw SQL stats query — none of the three are exposed to this bug shape. All `*-server-connectors` modules (including this one) already have `test { useTestNG() }` in their `build.gradle`, so no repeat of the gap #9244 found in `postgres-server-connectors`.

## Related Issue(s)

None filed — a contained, self-explanatory fix mirroring an already-merged pattern (#9244), found via a source-reading pass rather than a separate bug report.

## Testing

Added `DB2LUWDatabaseStatsExtractorTest`, mirroring `PostgresDatabaseStatsExtractorTest`'s structure: a non-negative-passes-through case and a sentinel-does-not-produce-negative-count case. `./gradlew :...:db2luw-server-connectors:test` — 4/4 pass (2 new + 2 existing FFDC tests, no regressions). `checkstyleMain`/`checkstyleTest` clean. No live Db2 instance available in this environment to verify against a real, un-RUNSTATS'd table — disclosed honestly, same limitation as prior contributions in this connector family.

None needed — internal survey-measurement correctness fix, no API/behavior change beyond the corrected value.

## Additional notes

Found while auditing sibling `*-server-connectors` modules for the same class of gap as #9244, as a deliberate depth pass rather than a new area.


mandy-chessell (approved):
Thank you :)

